PR #4, opened 2012-02-07
T333

dseif:
1. Plugin tray items are now droppable inside targets on the page
2. The created trackevent defaults the target to whatever target is what dropped in
3. Defaults to 1 second track event and position is at the current time of the media


secretrobotron:
I'm assuming the "undefined" that appears in the targets is just because there's no data for popcorn yet. Right?

dseif:
Yeap

PR #6, opened 2012-02-07
Scrubber follows currentTime of popcorn'd object.

secretrobotron:
1. Scrubber follows currentTime of popcorn'd object.
2. Added paused/playing event mechanism to media object.


secretrobotron:
This is too old to matter now.

PR #8, opened 2012-02-09
T346 - Editor powerup

secretrobotron:
Editor works again, and timeline/preview modules cause less trouble.


dseif:
Alright if we are going to leave this in for now then this looks good to me

src/editor/editor.js
```diff
@@ -0,0 +1,246 @@
...
+            manifest: Popcorn.manifest[ trackEvent.type ],
+            popcornOptions: trackEvent.popcornOptions,
+            targets: targets
+          });
+        },
+        submit: function( e ) {
+          trackEvent.update( e.data );
+        }
+      });
+/*
```
jbuck: Remove the commented block below

PR #10, opened 2012-02-10
T350

secretrobotron:
1. Merge of t346
2. Popcorn underpinnings reworked, removed from preview module.


dseif:
Also still looks like the popcorn functionality is still in the previewer. We should be able to tear that out now, no?

src/core/popcorn-wrapper.js
```diff
@@ -0,0 +1,304 @@
...
+    function prepareMedia( url, target, onError ){
+      var mediaElement = document.getElementById( target );
+      if ( _mediaType === "object" ) {
+        if (  !mediaElement || [ 'AUDIO', 'VIDEO' ].indexOf( mediaElement.nodeName ) === -1 ) {
+          var video = document.createElement( "video" ),
+              src = document.createElement( "source" );
+
+          src.addEventListener( "error", onError );
+          src.src = url;
+          video.style.width = document.getElementById( target ).style.width;
```
jbuck, line 90: Will video.style.width/height always exist? isn't it video.width and video.height?

PR #17, opened 2012-02-16
T334 - Add helper to event draggables

cadecairos:


cadecairos:
I... don't know why all those other commits are there. all I did was rebase my commits onto develop 0.o

PR #24, opened 2012-02-17
T377

secretrobotron:
Fixed core api and tests.

### DIFF
--- a/src/butter-main.js
+++ b/src/butter-main.js
@@ -66,7 +66,13 @@ THE SOFTWARE.
 
     var __guid = 0;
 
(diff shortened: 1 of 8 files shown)


mjschranz:
The only issues I see are some styling problems in core.js

Lines:

406
303
254
244
152
130

dseif:
Butter isn't as strict as Popcorn with styling, so I would say everything other than the rogue space on 406 is fine. I commented on the line to make sure Bobby see's that.  In the future I would do that as it's easier for everyone to see and track.

PR #25, opened 2012-02-18
Round timecodes to 3 decimal places

mbuttu:


src/core/trackevent.js
```diff
@@ -1,5 +1,7 @@
 define( [ "./logger", "./eventmanager", "util/lang" ], function( Logger, EventManager, util ) {
 
+  const NUMBER_OF_DECIMAL_PLACES = 3;
```
jbuck: Change this to var; only Firefox and Chrome support const. See MDN's article on it: https://developer.mozilla.org/en/JavaScript/Reference/Statements/const

PR #29, opened 2012-02-23
T368b

dseif:


src/core/target.js
```diff
@@ -32,7 +32,7 @@ THE SOFTWARE.
 
       options = options || {};
       var name = options.name || "Target" + id + Date.now();
-      this.object = options.object;
+      this.object = document.getElementById( options.object );
```
jbuck: should this be options.id, if id is supposed to be the element id?

src/core/target.js
```diff
@@ -63,9 +60,9 @@ THE SOFTWARE.
         },
         set: function( importData ) {
           if ( importData.name ) {
-            name = importData.name
+            name = importData.name;
           }
-          this.object = importData.object
+          this.object = document.getElementById( importData.object );
```
jbuck: Same comment as above

test/core.js
```diff
@@ -47,9 +47,9 @@
 
     createButter( function( butter ){
 
-      var m1 = butter.addMedia( { name: "Media 1", target: "audio-test", url: "www.google.ca" } );
+      var m1 = butter.addMedia( { name: "Media 1", target: "audio-test", url: "http://videos-cdn.mozilla.net/serv/webmademovies/laylapop.ogv" } );
```
jbuck: Can you change this to a local video? We can use the Popcorn.js test videos

test/core.js
```diff
@@ -47,9 +47,9 @@
...
       ok( m1.name === "Media 1", "Name is correct" );
-      ok( m1.target === "audio-test" && m1.url === "www.google.ca", "Media storage is correct" );
+      ok( m1.target === "audio-test" && m1.url === "http://videos-cdn.mozilla.net/serv/webmademovies/laylapop.ogv", "Media storage is correct" );
```
jbuck: Same as above

test/core.js
```diff
@@ -75,7 +75,7 @@
         mediaState = [ 0, media.data ];
       });
 
-      var m1 = butter.addMedia( { name: "Media 1", target: "audio-test", url: "www.google.ca" } ),
+      var m1 = butter.addMedia( { name: "Media 1", target: "audio-test", url: "http://videos-cdn.mozilla.net/serv/webmademovies/laylapop.ogv" } ),
```
jbuck: Same as aboce

src/core/target.js
dseif: currently the id is passed in on options.object, but we could do what you suggested and make them set it on id instead of object. You're probably right that using object doesn't make sense so I will fix that up

PR #33, opened 2012-02-24
T254

secretrobotron:
Default editor changes.
- Automatic submission on data change
- Situatied in `src/editor` instead of `test/`

### DIFF
--- a/src/editor/module.js
+++ b/src/editor/module.js
@@ -35,24 +35,25 @@ THE SOFTWARE.
             EventManager, 
             TrackEvent,
             Editor
-          ) {
+          ){
 
-    var EventEditor = function( butter, options ) {
+    return function( butter, options ){
 
       options = options || {};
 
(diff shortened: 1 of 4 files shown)


src/editor/module.js
```diff
@@ -79,10 +80,8 @@ THE SOFTWARE.
        return oldSource;
       }; //remove
 
-      var defaultEditor = options.defaultEditor || DEFAULT_EDITOR;
```
jbuck, line 82: Delete the declaration of DEFAULT_EDITOR as well

jbuck, line 82: Actually, does that make sense? DEFAULT_EDITOR is still being used on https://github.com/secretrobotron/butter/blob/9046a38007e29a4f3ac11263128ee61da470b77c/src/editor/module.js#L46

Either way, once that's cleared up, SR+

PR #36, opened 2012-02-24
t186

secretrobotron:
Added trackeventupdated listener to editor.js and passthrough for default-editor.

### DIFF
--- a/src/editor/editor.js
+++ b/src/editor/editor.js
@@ -49,6 +49,11 @@ define( [ "core/eventmanager" ], function( EventManager ) {
     });
 
     this.open = function( trackEvent ) {
+
+      function onTrackEventUpdated( e ){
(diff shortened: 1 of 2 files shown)


dseif:
Code looks good abut there seems to be an issue when picking up a track event and pushing it all the way to the left on a track ( it gives you a start time of -0.01 )

secretrobotron:
Think it's the same problem Chris was having with selenium tests?
To be clear, you're not able to move the trackevent way off the track area, right?

dseif:
Nope you can't drag a trackevent way off, that part still works fine.  I don't think this is the issue Chris was having, as he was seeing the problem when dropping a trackevent from the plugin tray onto the timeline and it would be about 1 or 2 px off.  This just seems like we leave a little bit too much room on the track bars on the left hand side.  Even on page load you can see that there is a tiny bit of space to the left of the trackevent that is at 0 seconds.

PR #37, opened 2012-02-24
T400

mjschranz:
Stops audio/video from playing when video has started, mouse is down, scrubber has been moved and is currently still.


dseif:
We are also going to need to reset the flag if the video is ever paused, as if the user pauses then drags the scrubber the video will start playing again.

Also mind fixing this pull request so that it is just the last commit instead of the merges as well?

PR #41, opened 2012-02-27
T351

secretrobotron:
Mouse wheel scrolling for the timeline.

### DIFF
--- a/src/timeline/scrollbars.js
+++ b/src/timeline/scrollbars.js
@@ -24,6 +24,8 @@ THE SOFTWARE.
 
 define( [], function(){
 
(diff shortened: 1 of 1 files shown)


src/timeline/scrollbars.js
```diff
@@ -24,6 +24,8 @@ THE SOFTWARE.
 
 define( [], function(){
 
+  const MOUSE_WHEEL_SCROLL_DIST = 10;
```
jbuck, line 27: change const to var, not supported in IE, Safari, or Opera

jbuck:
This was rolled into pull request #45

PR #47, opened 2012-02-27
Validate a track event's start and end time

mbuttu:
Ensure that a track event's start time is less than the duration of the
video. Also, make sure that a track event's end time is less than or
equal to the video's duration. Display an error message to the user if
any of these conditions are not true.

### DIFF
--- a/editors/default-editor.html
+++ b/editors/default-editor.html
@@ -1,5 +1,10 @@
 <html>
   <head>
+    <style>
+      p#message {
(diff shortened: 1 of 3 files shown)


dseif:
This got staged yesterday in a different pull request, closing.
https://webmademovies.lighthouseapp.com/projects/65733-butter/tickets/390-setting-trackevent-out-time-video-duration-causes-issues

PR #51, opened 2012-02-27
Make sure times are rounded when track event moves

mbuttu:
Also, the _round function is now a static function on Butter, named
round. numberOfDecimalPlaces is a static field on Butter as well, and
controls how many decimal places Butter.round rounds to.


dseif:
This ticket got resolved yesterday, closing this.
https://webmademovies.lighthouseapp.com/projects/65733-butter/tickets/256-timecodes-should-round-to-1-decimal

PR #55, opened 2012-02-28
Cornfield integration

jbuck:
Too hard to find the comments otherwise...

### DIFF
--- a/src/io/cornfield.js
+++ b/src/io/cornfield.js
@@ -1,57 +1,36 @@
-/**********************************************************************************
+/* This Source Code Form is subject to the terms of the MIT license
+ * If a copy of the MIT license was not distributed with this file, you can
+ * obtain one at http://www.mozillapopcorn.org/butter-license.txt */
 
-Copyright (C) 2012 by Mozilla Foundation
+define(['util/xhr'], function(XHR) {
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
+  function audience() {
+    return location.protocol + "//" + location.hostname + ( location.port ? ":" + location.port : "" );
+  }
 
-The above copyright notice and this permission notice shall be included in
-all copies or substantial portions of the Software.
+  function parameterize(data) {
+    var s = [];
 
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-THE SOFTWARE.
-
-**********************************************************************************/
-
-(function() {
-
-  define(['util/xhr'], function(XHR) {
-
-    function audience() {
-      return location.protocol + "//" + location.hostname + ( location.port ? ":" + location.port : "" );
+    for (var key in data) {
+      s[s.length] = encodeURIComponent(key) + "=" + encodeURIComponent(data[key]);
     }
 
-    function parameterize(data) {
-      var s = [];
-      
-      for (var key in data) {
-        s[s.length] = encodeURIComponent(key) + "=" + encodeURIComponent(data[key]);
-      }
+    return s.join("&").replace("/%20/g", "+");
+  }
 
-      return s.join("&").replace("/%20/g", "+");
-    }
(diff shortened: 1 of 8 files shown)


jbuck:
What should I do with the html changes? Just remove em completely?

jbuck:
Idea for testing: Have a separate test page that only tests cornfield integration, as it requires user interaction with BrowserID. This will be the simplest method of getting some tests in.

src/io/cornfield.js
```diff
@@ -0,0 +1,112 @@
...
+            var response = JSON.parse(this.response);
+            callback(response);
+          }
+        });
+      }
+
+      this.list = function(callback) {
+        XHR.get(server + "/files", function() {
+          if (this.readyState === 4) {
+            var response = JSON.parse(this.response);
```
secretrobotron, line 61: This check for response was correct. If something fails, an error will occur inside here. Let's catch it somehow (either with the if again, or a try/catch) and use the callback to generate a safe error.

src/io/cornfield.js
```diff
@@ -0,0 +1,112 @@
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+
+**********************************************************************************/
+
+(function() {
```
secretrobotron: You don't really need this `function`. Just a relic from the good ol' days.

test/io/cornfield.js
```diff
@@ -0,0 +1,99 @@
...
+
+( function ( window, document, undefined ){
+
+  QUnit.config.reorder = false;
+
+  var b = new Butter({
+    config: "../../config/default.conf",
+      ready: function( butter ){
+
+        module( "Unauthenticated tests" );
```
secretrobotron, line 15: So, instead of enforcing `QUnit.config.reorder`, couldn't you use the `module`'s `setup` function to just authenticate before each test? I'd prefer not to use something edge-y like this flag if we can avoid it.

S'okay if it's a huge effort right now. But, really, exploiting the module setup function is good practice.

test/io/cornfield.js
```diff
@@ -0,0 +1,99 @@
...
+            deepEqual( JSON.parse(res), { error: 'unauthorized' }, 'Not allowed to get files' );
+          });
+
+          butter.cornfield.push("test1", "test1", function(res) {
+            deepEqual( res, { error: 'unauthorized' }, 'Not allowed to put files' );
+          });
+
+          setTimeout(function() {
+            start();
+          }, 500);
```
secretrobotron, line 41: Why the half-second wait? No callback can do this instead?

test/io/cornfield.js
```diff
@@ -0,0 +1,99 @@
...
+            ok(res, 'The login response has data');
+            equal(res.status, 'okay', 'Login status is "okay"');
+            ok(res.email, 'The login has an email: ' + res.email);
+            equal(res.email, butter.cornfield.user(), "Email is stored");
+            start();
+          });
+
+          var failSafe = setTimeout(function() {
+            start();
+          }, 10000);
```
secretrobotron, line 58: There _is_ a global QUnit async timeout. I think it's `QUnit.config.timeout` or something.

test/io/cornfield.js
```diff
@@ -0,0 +1,99 @@
...
+
+        module( "Authenticated tests" );
+
+        asyncTest("Async API", 7, function() {
+          // Create a random filename we'll use for testing
+          var filename = "test" + Date.now();
+
+          butter.cornfield.list(function(res){
+            ok(res, 'The file list response has data');
+            console.log(res);
```
secretrobotron: Probably don't need this, right?

test/template.html
```diff
@@ -3,6 +3,7 @@
 <head>
     <title>Basic template</title>
     <meta content="text/html; charset=UTF-8" http-equiv="Content-Type">
+    <script src="https://browserid.org/include.js" type="text/javascript"></script>
```
secretrobotron, line 6: Are we always going to have to put this in templates? Can `butter` just include it automagically somehow? Perhaps we can `require` it?

test/template.js
```diff
@@ -44,6 +43,7 @@ document.addEventListener( "DOMContentLoaded", function( e ){
 
           });
         });
+
```
secretrobotron, line 46: Super glad there's a space here now.

test/template.html
jbuck, line 6: Nope, I filed https://webmademovies.lighthouseapp.com/projects/65733/tickets/436-have-butteriocornfield-write-out-a-script-tag-to-load-browserid-if-its-not-loaded on that

test/template.js
jbuck, line 46: Me too!

test/io/cornfield.js
jbuck: Nope!

test/io/cornfield.js
jbuck, line 41: Boooo

test/io/cornfield.js
jbuck, line 58: Also boooo

test/io/cornfield.js
jbuck, line 15: Module setup isn't async

src/io/cornfield.js
jbuck, line 61: You bet

src/io/cornfield.js
jbuck: Done

PR #56, opened 2012-02-28
T420

secretrobotron:
Dialog module doesn't really exist anymore, and the constituent library is much simpler.


jbuck:
Some nits: 
- IFRAMEDialog looks weird, can you rename to something like InlineDialog? Looks better
- We're missing licenses on many of these files, but I think we should strip all of the licenses on individual files and just have a project-wide LICENSE file.

src/dialog/iframe-dialog.js
```diff
@@ -27,53 +34,69 @@ define( [ "./util", "./base-dialog", "core/comm", "./event-manager" ], function(
...
+      _comm.unlisten( "cancel", onCancel );
+      _comm.unlisten( "close", _this.close );
+      _comm.destroy();
+      _open = false;
+      for( var e in _listeners ){
+        _em.unlisten( e, _listeners[ e ] );
+      } //for
       _em.dispatch( "close" );
-      _currentComm = _currentTemplate = undefined;
     } //close
```
jbuck: Needs a semi-colon

src/dialog/modal.js
```diff
@@ -0,0 +1,30 @@
+define( [], function(){
+
+  var __container;
```
jbuck, line 3: Rename to _container

src/dialog/window-dialog.js
```diff
@@ -1,9 +1,15 @@
...
-  const DEFAULT_WINDOW_HEIGHT = 479;
+  var DEFAULT_WINDOW_WIDTH = 640,
+      DEFAULT_WINDOW_HEIGHT = 479,
+      WINDOW_CHECK_INTERVAL = 300;
 
-  var WindowDialog = function( context, dialogOptions ) {
+  return function( dialogOptions ) {
     dialogOptions = dialogOptions || {};
 
     if( !dialogOptions.url ){
```
jbuck, line 15: Bah, I can't get this on the correct line, but the error message below refers to the iframe dialog, not the window dialog.

src/dialog/window-dialog.js
```diff
@@ -33,63 +44,87 @@ define( [ "./util", "./base-dialog", "core/comm", "./event-manager" ], function(
...
+        _window.close();
+      } //if
+      clearInterval( _statusInterval );
+      _comm = _window = undefined;
+      _open = false;
+      for( var e in _listeners ){
+        _em.unlisten( e, _listeners[ e ] );
+      } //for
+      _em.dispatch( "close" );
     } //close
```
jbuck: Needs a ;

src/dialog/modal.js
secretrobotron, line 3: This was on purpose so that it's known that the scope is meant to be class-wide, not instance-wide.

jbuck, line 3: I need a cheatsheet

PR #64, opened 2012-02-29
[t211]

secretrobotron:
getHTML functionality (must use console for now)

### DIFF
--- a/src/butter.js
+++ b/src/butter.js
@@ -40,11 +40,11 @@
         path = path.join( '/' ) + '/';
 
         if ( !window.require ) {
-          document.write( '<script src="' + path + '../external/require/require.js"></' + 'script>' );
+          document.write( '<script butter-exclude="true" src="' + path + '../external/require/require.js"></' + 'script>' );
         } //if
 
         // Set up paths to find scripts.
-        document.write('<script>' + 
+        document.write('<script butter-exclude="true">' + 
           '(function(){' + 
           'var ctx = require.config({ ' + 
             'baseUrl: "' + path + '",' +
--- a/src/butter.js
+++ b/src/butter.js
@@ -40,11 +40,11 @@
         path = path.join( '/' ) + '/';
 
         if ( !window.require ) {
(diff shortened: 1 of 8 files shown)


jbuck:
For HTML validation, can we s/butter-exclude/data-butter-exclude/ ?

src/butter.js
```diff
@@ -40,11 +40,11 @@
         path = path.join( '/' ) + '/';
 
         if ( !window.require ) {
-          document.write( '<script src="' + path + '../external/require/require.js"></' + 'script>' );
+          document.write( '<script butter-exclude="true" src="' + path + '../external/require/require.js"></' + 'script>' );
```
jbuck: should it be butter-exclude=true or just butter-exclude ?

src/core/page.js
```diff
@@ -40,6 +40,48 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
...
+        var node = toExclude[ i ];
+        node.parentNode.removeChild( node );
+      } //for
+
+      toClean = body.querySelectorAll( "*[butter-clean=\"true\"]" );
+      for( var i=0, l=toClean.length; i<l; ++i ){
+        var node = toClean[ i ];
+        node.removeAttribute( "butter-clean" );
+        node.removeAttribute( "data-butter" );
+        node.className = node.className.replace( /ui-droppable/g, "" );
```
jbuck: Using a regexp to delete text? that's AWFUL! But classList (https://developer.mozilla.org/en/DOM/element.classList) doesn't exist everywhere, so bleh

src/butter.js
secretrobotron: I don't know if it matters. Like `controls` in a `video` tag. I never put `="true"`. Is that a HTML validation infringement too?

secretrobotron: The w3c page on <video> seems to not care. I prefer to leave it off because it's prettier and harder to screw up.

jbuck: That's cool, I just wanted consistency. and data-butter-exclude is shorter!

PR #65, opened 2012-03-01
T412

benrito:
Scrollbar height addressed. This also correctly aligns the play button, timeline, and mute.

It might break the transition effect (timeline to the bar when collapsed) let me look into that.


dseif:
Looks like #412 got set to a dupe and #438 got staged, we good to close this ben?

PR #69, opened 2012-03-01
[t268]

secretrobotron:
Fade-ins for dialogs.


src/dialog/iframe-dialog.js
```diff
@@ -84,6 +84,12 @@ define( [
         });
       }, false );
       _parent.appendChild( _iframe );
+
```
jbuck, line 87: Wat. Seriously?

PR #82, opened 2012-03-01
[t390]

secretrobotron:
Took Mo's work, and factored in proper rounding mechanisms.

### DIFF
--- a/editors/default-editor.html
+++ b/editors/default-editor.html
@@ -57,7 +57,7 @@
 
           _comm.listen( "trackeventupdatefailed", function( e ) {
             if( e.data.error === "trackeventupdate::invalidtime" ){
-              document.getElementById( "message" ).innerHTML = "You've entered an invalid start or end time. Please verify that they are both greater than 0 and less than the duration of the media.";
+              document.getElementById( "message" ).innerHTML = "You've entered an invalid start or end time. Please verify that they are both greater than 0, the end time is equal to or less than the media's duration, and that the start time is less than the end time.";
             } //if
           });
 
--- a/editors/default-editor.html
+++ b/editors/default-editor.html
@@ -28,11 +28,12 @@
 
           okPressed = function( e ) {
             sendData();
-          };
+            _comm.send( "close" );
+          }; //okPressed
 
           cancelPressed = function( e ) {
             _comm.send( "cancel" );
-          };
+          }; //cancelPressed
 
           document.addEventListener( "keydown", function( e ) {
             if( e.keyCode === 13 ) {
@@ -49,7 +50,6 @@
               var element = document.getElementById( item );
               element.value = e.data[ item ];
             } //for
-            _comm.send( "close" );
           });
 
           _comm.listen( "trackeventupdatefailed", function( e ) {
--- a/editors/default-editor.html
+++ b/editors/default-editor.html
@@ -17,18 +17,21 @@
               okPressed,
               cancelPressed;
 
-          function sendData(){
+          function sendData( alsoClose ){
+            alsoClose = !!alsoClose; 
             var popcornOptions = {};
             for( var item in _manifest ) {
               popcornOptions[ item ] = document.getElementById( item ).value;
             }
             document.getElementById( "message" ).innerHTML = "";
-            _comm.send( "submit", popcornOptions );
+            _comm.send( "submit", {
+              eventData: popcornOptions,
+              alsoClose: alsoClose
+            });
           } //sendData
 
           okPressed = function( e ) {
-            sendData();
-            _comm.send( "close" );
+            sendData( true );
           }; //okPressed
 
           cancelPressed = function( e ) {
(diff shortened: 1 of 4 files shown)


src/util/time.js
```diff
@@ -0,0 +1,27 @@
+define( [], function(){
+
+  var __timeAccuracy = 3;
+
+  function roundTime( time ){
```
mbuttu, line 5: This function rounds any number, so perhaps a more general name could be used, such as `round`, or `roundNumber`

src/util/time.js
```diff
@@ -0,0 +1,27 @@
...
+    return Math.round( time * ( Math.pow( 10, __timeAccuracy ) ) ) / Math.pow( 10, __timeAccuracy );
+  } //round
```
mbuttu: Depending on the note above, the `//round` comment should match the function name

src/util/time.js
```diff
@@ -0,0 +1,27 @@
...
+
+  var utils = {
+    roundTime: roundTime
+  };
+
+  Object.defineProperties( utils, {
+    timeAccuracy: {
```
mbuttu, line 14: Because the rounding is done to any number, perhaps a better suited name for `timeAccuracy` is `numberAccuracy`

editors/default-editor.html
```diff
@@ -46,6 +52,12 @@
             } //for
           });
 
+          _comm.listen( "trackeventupdatefailed", function( e ) {
+            if( e.data.error === "trackeventupdate::invalidtime" ){
+              document.getElementById( "message" ).innerHTML = "You've entered an invalid start or end time. Please verify that they are both greater than 0 and less than the duration of the media.";
```
mbuttu: The end time can be equal to the duration of the video

src/editor/editor.js
```diff
@@ -98,16 +102,31 @@ define( [ "core/eventmanager", "dialog/iframe-dialog", "dialog/window-dialog" ],
...
         },
         submit: function( e ) {
-          if( e.data.target !== _currentTarget ){
-            _currentTarget = e.data.target;
-            blinkTarget();
+          var duration = TimeUtil.roundTime( butter.currentMedia.duration );
+          if( e.data &&
+              ( e.data.start < 0 ||
+                e.data.end > duration ||
+                e.data.start > e.data.end ) ){
```
mbuttu: This should be `e.data.start >= e.data.end`, since a track event can't start when a video ends

mbuttu:
Also, do we want a number such as 12.33 to show up as 12.330?

editors/default-editor.html
```diff
@@ -17,17 +22,18 @@
...
               popcornOptions[ item ] = document.getElementById( item ).value;
             }
-            _comm.send( "submit", popcornOptions);
+            document.getElementById( "message" ).innerHTML = "";
+            _comm.send( "submit", popcornOptions );
           } //sendData
 
           okPressed = function( e ) {
             sendData();
             _comm.send( "close" );
```
mbuttu: The popup shouldn't close when okay is pressed in case there is an error. The closing should be done in the call back for `trackeventupdated`.

src/core/trackevent.js
```diff
@@ -1,4 +1,4 @@
-define( [ "./logger", "./eventmanager", "util/lang" ], function( Logger, EventManager, util ) {
+define( [ "./logger", "./eventmanager", "util/lang", "util/time" ], function( Logger, EventManager, LangUtil, TimeUtil ) {
 
   var NUMBER_OF_DECIMAL_PLACES = 3,
```
mbuttu: This is no longer needed

src/editor/editor.js
```diff
@@ -98,16 +102,31 @@ define( [ "core/eventmanager", "dialog/iframe-dialog", "dialog/window-dialog" ],
...
+            trackEvent.dispatch( "trackeventupdatefailed", {
```
mbuttu: The issue with dispatching the `trackeventupdatefailed` event here is that if a track event is edited somewhere else in the code, or if there will be another editor created later on, this error will not be thrown. The error checking for this should go in a place where no matter how a track event is updated, this check will be done.

src/util/time.js
secretrobotron, line 5: In the "time" util file? I'm going to leave this for now. If someone wants to use it, we can make it more general.

src/util/time.js
secretrobotron: Done.

src/util/time.js
secretrobotron, line 14: Not in "util/time". Keeping unless someone wants to generalize later.

editors/default-editor.html
secretrobotron: I agree.

src/editor/editor.js
secretrobotron: Created #458 to deal with this specifically: https://webmademovies.lighthouseapp.com/projects/65733-butter/tickets/458-protect-from-invalid-times-in-trackevents

PR #87, opened 2012-03-09
Update Popcorn.js to v1.2

jbuck:
Template works and tests pass with this new version


dseif:
PR+

rwaldron:
SR+

... I'm having way too much fun with this. Sorry, I actually didn't notice that this was in "Butter"

PR #88, opened 2012-03-12
Fix error message for invalid start and end times

mbuttu:
1. The error message now refers to start and end times as in and out
   times, since that is how it is displayed to the user in the editor.
2. The media's duration is displayed in the error message.
3. The error message is now displayed as a bulleted list for
   readability.


jbuck:
Closing old pull requests

PR #92, opened 2012-03-14
T408

secretrobotron:
Removed a lot of the timeline module tackliner silliness. It feels a lot more like a module now, and less of a constant hack-stack.


src/core/media.js
```diff
@@ -187,6 +197,21 @@
...
 
       Object.defineProperties( this, {
+        view: {
+          enumerable: true,
+          get: function(){
+            return _view;
+          },
+          set: function( newView ){
+            if( _view ){
+              //unlisten to view
```
jbuck: Dead code here, 2 lines below as well

PR #97, opened 2012-03-19
Validate invalid track event times

mbuttu:
Track events were only validated in the editor.
Now track events are validated no matter where they are updated from.


dseif:
This got landed https://webmademovies.lighthouseapp.com/projects/65733-butter/tickets/458-protect-from-invalid-times-in-trackevents

PR #102, opened 2012-03-22
Add shelljs magic to Butter

jbuck:
Lint code with jshint [#231]


package.json
```diff
@@ -0,0 +1,29 @@
...
+  "description": "The Popcorn authoring tool",
+  "contributors": [
+    {
+      "name": "Bobby Richter",
+      "email": "secretrobotron@gmail.com"
+    },
+    {
+      "name": "Jon Buckley",
+      "email": "jon@jbuckley.ca"
+    }
```
secretrobotron: maybe add others here :).

src/almond.js
```diff
@@ -0,0 +1,277 @@
+/**
```
secretrobotron: Does this have to be in src? It's not really ours, which is why it was in "tools" before, and it's not really a lib.

README.md
```diff
@@ -3,10 +3,16 @@ Butter
 
 An SDK for authoring Popcorn projects.
 
+Prerequisites
```
secretrobotron: Maybe this should be "Build Prerequisites" instead. Not really required to just use or develop.

tools/almond.js
```diff
@@ -1,66 +1,26 @@
 /**
- * almond 0.0.0 Copyright (c) 2011, The Dojo Foundation All Rights Reserved.
+ * almond 0.0.3 Copyright (c) 2011, The Dojo Foundation All Rights Reserved.
  * Available via the MIT or new BSD license.
```
secretrobotron, line 3: Hahahaha ...fun.

PR #107, opened 2012-03-23
T437 arrow key and del key functionality on selected track event

ScottDowne:


src/butter-main.js
```diff
@@ -45,7 +45,8 @@
             icons: {}
           },
           _defaultTarget,
-          _this = this;
+          _this = this
```
secretrobotron: missing comma!!! GLOBAL LEAK!!

src/butter-main.js
```diff
@@ -405,6 +406,15 @@
             } //if
           },
           enumerable: true
+        },
+        selectedEvents: {
```
secretrobotron, line 410: drop an "enumerable: true" in here so that we can see the property in firebug, etc.

src/butter-main.js
```diff
@@ -405,6 +406,15 @@
...
+          get: function() {
+            return _selectedEvents;
+          },
+          set: function(selectedEvents) {
+            _selectedEvents = selectedEvents;
+            return _selectedEvents;
```
secretrobotron: I'm ... faaaairly certain returning anything here has no effect. Right?

src/core/track.js
```diff
@@ -137,7 +137,6 @@ define( [
         _trackEvents.target = _target;
       } //if
       _trackEvents.push( trackEvent );
-      trackEvent.track = _this;
```
secretrobotron, line 140: Why did this get removed, actually?

src/butter-main.js
ScottDowne, line 410: Done.

src/butter-main.js
ScottDowne: This was a test. You passed!

src/butter-main.js
ScottDowne: I'm pretty sure I agree, but I was following the setter inside currentTime above it. I have seen returns in setter before, and I _think_ it is to allow code like this:

var item = obj.item = 4;

With obj.item being a setter, but also returning the value? I will test this and get back to you.

src/core/track.js
ScottDowne, line 140: Because I see the same line again 8 lines down. I actually removed that out of curiosity to see if both lines were needed, with intending to file it, but completely forgot.

I will ping you on irc and we can see if it is needed, and see if we should file it. Good catch though! I forgot I did that.

src/butter-main.js
ScottDowne: Nope, not needed.

It knows to call the getter and setter when used like the above. I will remove the return.

PR #113, opened 2012-03-26
[#t364] Fixed issue where single and double quotes inside plugin text we...

dseif:
...re possibly not working


dseif:
Much better

dseif:
Deleting in replace of an updated version https://github.com/mozilla/butter/pull/144

PR #114, opened 2012-03-26
[#492] Rewrite timebar to use a fixed-width <canvas>

jbuck:


src/timeline/timebar.js
```diff
@@ -14,45 +14,50 @@ define( [ "util/lang", "./scrubber" ], function( util, Scrubber ) {
...
 
       _canvasContainer.style.width = containerWidth + "px";
 
-      _canvas.style.width = width + "px";
-
       var context = _canvas.getContext( "2d" );
 
       _canvas.height = _canvas.offsetHeight;
-      _canvas.width = _canvas.offsetWidth;
+      _canvas.width = containerWidth;
```
secretrobotron: Can we move this out of here, actually? My fear is that silly things like this still happen: http://stackoverflow.com/questions/2142535/how-to-clear-the-canvas-for-redrawing

src/timeline/timebar.js
```diff
@@ -14,45 +14,50 @@ define( [ "util/lang", "./scrubber" ], function( util, Scrubber ) {
...
           padding = 20,
           lastPosition = 0,
-          lastTimeDisplayed = -( ( textWidth + padding ) / 2 );
+          lastTimeDisplayed = -( ( textWidth + padding ) / 2 ),
+          start = _tracksContainer.element.scrollLeft / inc,
+          end = ( _tracksContainer.element.scrollLeft + containerWidth ) / inc;
 
       context.clearRect ( 0, 0, _canvas.width, _canvas.height );
-
+      context.translate( -_tracksContainer.element.scrollLeft, 0 );
```
secretrobotron, line 51: So, if we're just drawing inside the width of the viewport now, why are translations by the value of `scrollLeft` necessary?

src/timeline/timebar.js
```diff
@@ -93,16 +98,24 @@ define( [ "util/lang", "./scrubber" ], function( util, Scrubber ) {
       // stroke color
       context.strokeStyle = "#999999";
       context.stroke();
-      context.closePath();
+      context.translate( _tracksContainer.element.scrollLeft, 0 );
 
       _scrubber.update( containerWidth, zoom );
```
secretrobotron, line 106: You can completely skip passing `zoom` into this function now if you just call `scrubber.update` in the event listener and `update` functions below.

src/timeline/timebar.js
jbuck: I'm setting the width to match the container width, which can be less than the width of the screen with a small video. I clear the canvas on line 47 below

secretrobotron: Right, and that's accurate. But, it looks like it is (or recently was) used as a way to clear the canvas as well. So, in effect, you'd be clearing it twice. Can we set the width less often than every scroll update?

src/timeline/timebar.js
jbuck, line 51: scott's drawing algorithm assumes that it's drawing the entire timebar in one pass (look above, the inc is for the scrollWidth, not the canvas' width anymore). I translate by -scrollLeft to bring the part that should be drawn into the drawable part of the canvas. Getting rid of the translate calls here would require rewriting the drawing algorithm below, which is why I stuck with just translating the canvas and... manual culling haha

src/timeline/timebar.js
jbuck, line 106: If I move the scrubber.update calls into the event listeners, then that means I need to duplicate the containerWidth calculations into both functions. That doesn't sound like a win to me.

I will remove the _lastZoom member though, as the scrubber already stores the previous zoom level

src/timeline/timebar.js
secretrobotron, line 51: Ok, I think I get it now. Filing a "profiling" ticket.

jbuck, line 51: Covered by https://webmademovies.lighthouseapp.com/projects/65733/tickets/523-decouple-timeline-from-dom-widthheight maybe?

src/timeline/timebar.js
secretrobotron, line 106: I didn't even see the containerWidth calc. Nvm. This is fine :).

PR #126, opened 2012-03-29
Fix lint errors [#421]

jbuck:


src/butter-main.js
```diff
@@ -170,8 +170,10 @@
        ****************************************************************/
       //importProject - Import project data
       this.importProject = function ( projectData ) {
+        var i;
```
secretrobotron: It didn't complain about l?

src/core/comm.js
```diff
@@ -10,7 +10,7 @@ define( [ "./eventmanager" ], function( EventManager ){
     var _this = this,
         _em = new EventManager( _this ),
         _readyInterval,
-        _context = __context++
+        _context = __context++,
```
secretrobotron, line 13: omg. seriously.

This might be what was hurting Scott. :/

src/core/eventmanager.js
```diff
@@ -27,7 +27,7 @@ define( [], function(){
         e = {
           type: eventName + "",
           target: _target,
-          data: eventData,
+          data: eventData
```
secretrobotron, line 30: sigh.

src/butter-main.js
jbuck: That.... seems wrong. Let me fix that!

src/core/popcorn-wrapper.js
```diff
@@ -157,12 +157,13 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
     } //findMediaType
 
     var generatePopcornString = this.generatePopcornString = function( popcornOptions, url, target, method ){
-      var popcornString = "",
-          popcornOptions = "";
+      var popcornString = "";
 
       if ( popcornOptions ) {
         popcornOptions = ", " + JSON.stringify( popcornOptions );
```
secretrobotron, line 163: Where is popcornOptions declared now?

src/core/trackevent.js
```diff
@@ -53,10 +53,10 @@ define( [
           _popcornOptions[ prop ] = updateOptions[ prop ];
         } //if
       } //for
-      if ( _popcornOptions.start != null ) {
+      if ( _popcornOptions.start !== null ) {
```
secretrobotron: VETO.

There was a purpose to this being a lazy compare. If `_popcornOptions` doesn't have the `start` attribute at all, this will catch that, and `null !== undefined`. Perhaps it should be `undefined` instead.

src/core/views/trackevent-view.js
```diff
@@ -34,10 +34,10 @@ define( [ "core/logger", "core/eventmanager", "util/dragndrop" ], function( Logg
     this.update = function( options ){
       options = options || {};
       _element.style.top = "0px";
-      if( options.start != null ){
+      if ( options.start !== null ) {
```
secretrobotron: Same as above.

src/core/popcorn-wrapper.js
jbuck, line 163: in the function definition... before we'd just overwrite it with "" haha

src/core/views/trackevent-view.js
```diff
@@ -54,14 +54,6 @@ define( [ "core/logger", "core/eventmanager", "util/dragndrop" ], function( Logg
         enumerable: true,
         get: function(){ return _element; }
       },
-      duration: {
```
secretrobotron, line 57: This was just unused, I'm assuming.

jbuck, line 57: it's actually a duplicate defintion of "duration", if you look lower in the file I just used that

src/core/popcorn-wrapper.js
secretrobotron, line 163: loooooooool

noted.

src/core/views/trackevent-view.js
secretrobotron, line 57: Oh! Very nice. I caught myself doing that a couple of times.

src/editor/module.js
```diff
@@ -23,7 +23,7 @@
 
       var _editors = {},
           _logger = new Logger( "EventEditor" ),
-          _defaultEditor = options.default || DEFAULT_EDITOR,
+          _defaultEditor = options[ "default" ] || DEFAULT_EDITOR,
```
secretrobotron, line 26: I'm curious why lint thinks . is better than array-like access. Just for readability, I'm assuming.

jbuck, line 26: default is a javascript keyword, which is why it prefers it in array-like access in this case

src/plugin/module.js
```diff
@@ -181,13 +181,17 @@
...
+              i = 0,
+              l = 0,
+              check = function() {
+                if ( ++counter === plugin.length && cb ) {
+                  cb();
+                }
+              };
+        
+          for( i = 0, l = plugin.length; i < l; i++ ) {
+            __this.add( plugin[ i ], check );
```
secretrobotron, line 194: Much less gross. Very nice.

To make it a little less confusing to read, think we can just declare

`function check(){` instead of making it a var?

src/timeline/module.js
```diff
@@ -49,12 +49,14 @@ define( [
...
-      var curleft = curtop = 0;
+      var curleft = 0,
+          curtop = 0;
+
       if( obj.offsetParent ) {
         do {
           curleft += obj.offsetLeft;
           curtop += obj.offsetTop;
-        } while ( obj = obj.offsetParent );
+        } while ( ( obj = obj.offsetParent ) );
```
secretrobotron, line 59: Seriously?

src/timeline/module.js
```diff
@@ -138,7 +144,7 @@ define( [
     });
 
     this.currentTimeInPixels = function( pixel ){
```
secretrobotron: I think this function isn't used anywhere ever for anything.

Can we destroy it? It's old and slow.

src/timeline/scrollbars.js
```diff
@@ -17,7 +17,6 @@ define( [ "core/eventmanager" ], function( EventManager ){
         _childHeight,
         _scrollHeight,
         _handleHeight,
-        _handleHeight,
```
secretrobotron, line 20: Wow.

src/plugin/module.js
jbuck, line 194: lint complains with:

src/plugin/module.js: line 188, col 11, Function declarations should not be placed in blocks. Use a function expression or move the statement to the top of the outer function.

If I do that :(

src/core/trackevent.js
jbuck: switching to isNaN() here

src/core/views/trackevent-view.js
jbuck: switching to isNaN() here

src/timeline/module.js
jbuck, line 59: Seriously.js

src/timeline/module.js
jbuck: nuke it from orbit

src/editor/module.js
secretrobotron, line 26: annnnnd ignore me, because i prefer that too. I misread the diff and thought it was in reverse. CARRY ON.

src/plugin/module.js
secretrobotron, line 194: Hmmmm. That ... is odd.

This is worth talking about. Not here though.

jbuck, line 194: I think it's because it'll declare that function everytime it goes into add(), so it's a valid reason to complain about it. We'd need to another another closure and function outside the add() function to get the lexical scoping we desire though :/

PR #132, opened 2012-03-30
T457

dseif:


src/core/popcorn-wrapper.js
```diff
@@ -97,54 +112,6 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
...
-          mediaElement.removeAttribute( "src" );
-          mediaElement.addEventListener( "error", onError );
-          mediaElement.src = url;
-          mediaElement.load();
-          _popcornTarget = mediaElement.id;
-          return mediaElement;
-        } //if
-      } //if _mediaType
-    } //prepareMedia
-
```
secretrobotron, line 147: YYEESSSSSSS!!

src/core/popcorn-wrapper.js
```diff
@@ -174,29 +141,43 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
...
+            }
+          }, 100 );
+        }
+        return "var popcorn = Popcorn.smart( '#" + target + "', '" + url + "' );\n";
       };
-      
+
       // call certain player function depending on the regexResult
-      popcornString += players[ _mediaType ]();
+      popcornString += players( _mediaType );
```
secretrobotron: Can you make this function not called `players` anymore? It needs a more descriptive name since it's actually... a function hahaha.

src/core/popcorn-wrapper.js
```diff
@@ -174,29 +141,43 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
...
         }
+
+        if( type === "object" ) {
+          _playerReady = true;
+        } else {
+          _playerReady = false;
+
+          // our regex only handles youtu ( incase the url looks something like youtu.be )
+          if ( type === "youtu" ) {
+            type = "youtube";
```
secretrobotron, line 162: We only care about youtube? Not soundcloud or vimeo?

src/core/popcorn-wrapper.js
```diff
@@ -257,15 +238,6 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
...
         if ( _popcorn.media.readyState >= 2 && _popcorn.duration() > 0 ) {
-          if ( _mediaType === "youtu" ) {
-            setTimeout( function() {
-              _popcorn.pause();
-            }, 1000 );
-          }
-          else if ( _mediaType === "vimeo" || _mediaType === "soundcloud" ) {
-          }
-          else {
-          } //if
```
secretrobotron, line 268: So glad this is leaving.

test/template.html
```diff
@@ -30,9 +30,8 @@
     </style>
 </head>
 <body>
-    <video id="main" data-butter="media" controls >
-      <source src="http://videos-cdn.mozilla.net/serv/webmademovies/laylapop.ogv" />
-    </video>
+    <div id="main" data-butter="media" controls data-butter-source="http://videos.mozilla.org/serv/webmademovies/laylapop.ogv">
+    </div>
```
secretrobotron: Does this only work with divs now? Or can things be <video> or <audio> still?

src/core/popcorn-wrapper.js
dseif, line 162: the getMediaType function returns that value from the regex, which searches for youtu, vimeo, and soundcloud. So youtube obviously needs to be "youtube" and not "youtu". Vimeo and soundcloud just work ( well soundcloud not at the moment, but it will once the player is rewritten ).

src/core/popcorn-wrapper.js
dseif: will do

test/template.html
dseif: Currently only divs, which sort of sucks. Feel free to fail it if you want me to make it handle the audio and video elements as well.

src/core/popcorn-wrapper.js
secretrobotron, line 162: Oh, I see! So, this is just to correct what is spewed out. This is fine then.

src/core/popcorn-wrapper.js
```diff
@@ -174,29 +141,53 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
...
+        } else {
+          _playerReady = false;
+
+          // our regex only handles youtu ( incase the url looks something like youtu.be )
+          if ( type === "youtu" ) {
+            type = "youtube";
+          }
+
+          if( [ "VIDEO", "AUDIO" ].indexOf( target.nodeName ) === -1 ) {
+            var targetElement = document.getElementById( target ),
```
secretrobotron: This is a bit confusing. It looks like only audio and video elements are the things that won't be strings here. I don't think that's right.

/me checks

Yeah, just a little bit above, there's an if block that makes sure target is a string. So, you probably want to grab the element to do that node check here.

src/core/popcorn-wrapper.js
```diff
@@ -174,29 +141,53 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
...
+              _playerReady = true;
+            }
+          }, 100 );
+        }
+        return "var popcorn = Popcorn.smart( '#" + target + "', '" + url + "' );\n";
+      }
+
       // call certain player function depending on the regexResult
-      popcornString += players[ _mediaType ]();
+      popcornString += constructPlayerString( _mediaType );
```
secretrobotron: How about just `constructPlayer`, since it's actually manipulating the dom here?

src/core/popcorn-wrapper.js
```diff
@@ -174,29 +141,53 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
...
+
+          if( [ "VIDEO", "AUDIO" ].indexOf( target.nodeName ) === -1 ) {
+            var targetElement = document.getElementById( target ),
+                parentNode = targetElement.parentNode,
+                newElement = document.createElement( "div" );
+
+            newElement.id = targetElement.id;
+            newElement.style.cssText = getComputedStyle( targetElement ).cssText;
+            parentNode.replaceChild( newElement, targetElement );
+          }
```
secretrobotron, line 172: Let's file a ticket to make sure that we make bi-directional change possible. Right now, if you go from <video> -> <div>, it doesn't let you go back, you always have a new div in your page.

No need to pile it into this ticket though. Just file a new one.

dseif, line 172: Does it matter if it goes into a div tho? Adding in that functionality would make it a good bit more complex than it needs to be and a lot of the work thats done in smart would then be done here as well.

secretrobotron, line 172: Yeah, that's what I feared. It's important for people that want to make sure their dom stays the same, and just want to change the url of things that are already there. Style might get messed up, etc.

No worries though. It _functions_ now, just file a new bug to take care of it in 0.4.

PR #135, opened 2012-03-30
[#320] Can now edit track data, super basic error checking ( only for ma...

dseif:
...lformed JSON )

### DIFF
--- a/dialogs/track-data.html
+++ b/dialogs/track-data.html
@@ -12,7 +12,9 @@
 
         document.addEventListener( "DOMContentLoaded", function( e ){
           var closeButton = document.getElementById( "close" ),
+              okButton = document.getElementById( "ok" ),
               trackName = document.getElementById( "track-name" ),
+              errorMessage = document.getElementById( "error" ),
               trackData = document.getElementById( "track-data" );
 
           function closePressed( e ){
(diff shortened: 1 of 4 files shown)


src/core/track.js
```diff
@@ -119,6 +119,8 @@ define( [
           return _trackEvents[ i ];
         } //if
       } //for
+
+      return undefined;
```
secretrobotron: You don't need to say `undefined`, actually. `return` will just make it `undefined` if you're not returning any data.

dialogs/track-data.html
```diff
@@ -68,13 +85,20 @@
...
     </style>
   </head>
   <body>
     <div id="container">
       <h1>Data for <span id="track-name"></span></h1>
-      <textarea readonly id="track-data">Please wait...</textarea>
+      <textarea id="track-data">Please wait...</textarea>
+      <div id="error"></div>
       <div id="buttons">
+        <button id="ok">Ok</button>
```
secretrobotron: Make sure the "disabled" attribute is on this button too, otherwise, you can press "ok" before any data arrives and bork things.

src/core/track.js
dseif: Oh cool, will remove.

dialogs/track-data.html
dseif: kk

PR #137, opened 2012-03-30
Account for width of line when determining if scrubber should be hidden ...

cadecairos:
...#529

### DIFF
--- a/src/timeline/scrubber.js
+++ b/src/timeline/scrubber.js
@@ -62,7 +62,7 @@ define( [], function(){
         var pos = currentTime / duration * _tracksContainerWidth,
             adjustedPos = pos - scrollLeft;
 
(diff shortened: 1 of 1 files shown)


src/timeline/scrubber.js
```diff
@@ -62,7 +62,7 @@ define( [], function(){
         var pos = currentTime / duration * _tracksContainerWidth,
             adjustedPos = pos - scrollLeft;
 
-        if( pos <  scrollLeft || Math.floor( pos ) > _width + scrollLeft ){
+        if( pos <  scrollLeft || Math.floor( pos ) - _line.clientWidth > _width + scrollLeft ){
```
secretrobotron, line 65: Make sure you cache this value somewhere in this file. This function needs to run as fast as possible. You can probably store this when `update` happens.

PR #143, opened 2012-03-31
T570

dseif:


dseif:
Bobby wrote a similar patch so this is no longer needed.

PR #144, opened 2012-03-31
[#346] Create a way to encode strings and wrote a simple test ( which ma...

dseif:
...y not suffice, can't think of a better way to test this

### DIFF
--- a/src/core/popcorn-wrapper.js
+++ b/src/core/popcorn-wrapper.js
@@ -24,6 +24,12 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
         _playerReady = false,
         _this = this;
 
+    function encodeString( string ) {
+      return String( string )
+             .replace(/"/g, "\\" + '\"' )
+             .replace(/'/g, "\\" + "\'" );
+    }
+
     this.unbind = function(){
       try{
         removePopcornHandlers();
@@ -208,7 +214,7 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
             popcornString += "popcorn." + trackEvents[ i ]._natives.type + "({";
             for ( var option in popcornOptions ) {
               if ( popcornOptions.hasOwnProperty( option ) ) {
-                popcornString += "\n" + option + ":'" + trackEvents[ i ][ option ] + "',";
+                popcornString += "\n" + option + ":'" + encodeString( trackEvents[ i ][ option ] ) + "',";
               } //if
             } //for
             if ( popcornString[ popcornString.length - 1 ] === "," ) {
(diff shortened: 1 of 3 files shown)


test/core.js
```diff
@@ -568,5 +568,30 @@
...
+        butter.currentMedia = m2;
+        t3 = m2.addTrack();
+        t4 = m2.addTrack();
+        te1 = t4.addTrackEvent( { popcornOptions: { start: 2, end: 6 }, type: "footnote", text: String( "this'" + 'should"' + '""""""b"e' + "f'i'ne") } );
+        butter.addTarget( { name: "beep" } );
+
+      try {
+        var html = document.createElement( "html" );
+        html.innerHTML = butter.getHTML();
+        console.log( html );
```
secretrobotron: Lasers.

test/core.js
```diff
@@ -568,5 +568,30 @@
...
+  test( "exported HTML is properly escaped", function() {
+    expect( 1 );
+    createButter( function( butter ){
+    var m1 = butter.addMedia( { url:"www.test-url-1.com", target:"test-target-1" } );
+        m2 = butter.addMedia( { url:"www.test-url-2.com", target:"test-target-2" } );
+        t1 = m1.addTrack();
+        t2 = m1.addTrack();
+        butter.currentMedia = m2;
+        t3 = m2.addTrack();
+        t4 = m2.addTrack();
```
secretrobotron: You probably don't need some much code here to run this test and have it finish decisively.

Since the popcorn code is created using `popcorn-wrapper::generatePopcornString` eventually, why not just test with that instead? I believe you can get at it just by using `media.popcornString`. Maybe just come up with some tricky text to put in a footnote plugin text field and make sure it gets escaped properly on export.

You might have to wait until #242 lands, since it skips constructing the player here and just writes out the string instead.

dseif: media.popcornString sadly won't do the trick here. It only returns what's neccessary to create the Popcorn instance, not all the trackevent data as well. But I agree that I probably don't need so much stuff to test this functionality, so i'll tone it down.

secretrobotron: No? What's this code do? https://github.com/mozilla/butter/blob/master/src/core/popcorn-wrapper.js#L204

dseif: ooops :S

I read the constructPlayer method instead haha, my bad.

secretrobotron: Hahahah no worries. I'm only assuming it works properly... :P

src/core/popcorn-wrapper.js
```diff
@@ -24,6 +24,12 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
         _playerReady = false,
         _this = this;
 
+    function encodeString( string ) {
+      return String( string )
+             .replace(/"/g, "\\" + '\"' )
```
cadecairos: managed to write a regex to do this in one call: replace(/['\"]/g, "\$&")

cadecairos: lost a backslash there.. lets see if I can get the right one in there: 

replace(/['\\"]/g, "\$&")

PR #158, opened 2012-04-04
T597

secretrobotron:


cornfield/external/popcorn-js
```diff
@@ -0,0 +1 @@
+../../external/popcorn-js
```
jbuck, line 1: Ughhhhh this is a hack and we should fix this properly in v0.4

secretrobotron, line 1: Agreed. FILE IT!

PR #161, opened 2012-04-04
T431

benrito:
I have never been more ashamed in my life.

But these dialogs are about 40 percent there. Bob says he'll make these happen per #431


dseif:
431 has been landed, closing.

PR #162, opened 2012-04-04
Added a way to turn logging on and off

dseif:


test/core.js
```diff
@@ -611,4 +612,27 @@
     });
   });
 
+  module( "Debug functionality" );
+  asyncTest( "Debug enables/disables logging", 4, function() {
+    createButter(function( butter ) {
+      var count = 0;
+      equals( butter.debug, false, "debugging is initially false, logging should be enabled" );
+      console.log = function() {
```
jbuck, line 622: You should store console.log in a var, then set console.log back to the normal function once this test is done

src/core/logger.js
```diff
@@ -4,20 +4,34 @@
...
+    } //Logger
 
-  }); //define
+    logger.debug = function( value ) {
+      if ( value !== undefined ) {
+        __debug = value;
+      } else {
+        return __debug;
+      }
+    }
```
jbuck: Add a semi-colon here

PR #182, opened 2012-04-09
[t637] - media element ui positioning

secretrobotron:
1. Created PositionTracker class in UI namespace
2. Moved test template's video for testing.
3. Added listener to media-view for pageElement movement.
4. Fixed "blink" class to be "data-blink" for HTML validation
5. Cleaned up pageElement position checking control flow a lot

### DIFF
--- a/src/ui/position-tracker.js
+++ b/src/ui/position-tracker.js
@@ -1,0 +1,19 @@
+define([], function(){
+
+  var POLL_INTERVAL = 20;
+  
+  return function( object, movedCallback ){
+    var _rect = {};
+    setInterval( function(){
+      var newPos = object.getBoundingClientRect();
+      if( newPos.left !== _rect.left ||
+          newPos.right !== _rect.right ||
+          newPos.top !== _rect.top ||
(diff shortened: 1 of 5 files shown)


src/ui/position-tracker.js
```diff
@@ -0,0 +1,19 @@
...
+    setInterval( function(){
+      var newPos = object.getBoundingClientRect();
+      if( newPos.left !== _rect.left ||
+          newPos.right !== _rect.right ||
+          newPos.top !== _rect.top ||
+          newPos.bottom !== _rect.bottom ){
+        _rect = newPos;
+        movedCallback( _rect );
+      }
+    }, POLL_INTERVAL );
```
dseif, line 16: can't we listen for page resize instead of polling?

secretrobotron, line 16: It's got less to do with page resize and more to do with the position/dimensions of the element changing. Gecko supports DOMAttrModified, but it's pretty weak and less effective here.

If there's something better, we should use it. I just made some effort to compartmentalize this functionality so that when we find a better solution, this is the only code we'll have to change.

PR #201, opened 2012-04-12
[t663] - html snapshotting for export

secretrobotron:
1. Added _page to butter to expose functions
2. Added data-butter-exclude to dialog modal container
3. Added snapshotHTML/eraseSnapshot functions in Page class
4. Added test for html snapshotting

### DIFF
--- a/src/butter-main.js
+++ b/src/butter-main.js
@@ -566,6 +566,11 @@
         //prepare the page next
         preparePage(function(){
           moduleCollection.ready(function(){
+
+            if( _config.snapshotHTMLOnReady ){
+              _page.snapshotHTML();
+            }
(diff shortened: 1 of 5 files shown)


src/butter-main.js
```diff
@@ -600,6 +605,8 @@
         readConfig();
       } //if
 
+      this.page = _page;
```
jbuck, line 608: butter.page isn't already a module?

secretrobotron, line 608: Nope! It never was. Just an object the core uses to scrape the page.

secretrobotron:
No idea why this PR is still open. All of these commits have been merged in.

PR #203, opened 2012-04-12
[t662] - targets added several times

secretrobotron:
1. Fixed check for redundant target and media elements.
2. Bail when older elements are found instead of ...adding them a bunch of times.

### DIFF
--- a/src/butter-main.js
+++ b/src/butter-main.js
@@ -486,21 +486,28 @@
             medias = scrapedObject.media;
 
         _page.preparePopcorn(function() {
-          var i, j, il, jl, url;
+          var i, j, il, jl, url, oldTarget, oldMedia, mediaPopcornOptions;
+
           for( i = 0, il = targets.length; i < il; ++i ) {
+            oldTarget = null;
             if( _targets.length > 0 ){
               for( j = 0, jl = _targets.length; j < jl; ++j ){
                 // don't add the same target twice
-                if( _targets[ j ].id !== targets[ i ].id ){
-                  _this.addTarget({ element: targets[ i ].id });
+                if( _targets[ j ].id === targets[ i ].id ){
+                  oldTarget = _targets[ j ];
+                  break;
                 } //if
               } //for j
             }
-            else{
+
+            if( !oldTarget ){
               _this.addTarget({ element: targets[ i ].id });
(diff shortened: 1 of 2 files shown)


jbuck:
This needs a rebase on master. Also, are the changes in templates/test.html needed?

src/butter-main.js
```diff
@@ -486,21 +486,28 @@
...
+
           for( i = 0, il = targets.length; i < il; ++i ) {
+            oldTarget = null;
             if( _targets.length > 0 ){
               for( j = 0, jl = _targets.length; j < jl; ++j ){
                 // don't add the same target twice
-                if( _targets[ j ].id !== targets[ i ].id ){
-                  _this.addTarget({ element: targets[ i ].id });
+                if( _targets[ j ].id === targets[ i ].id ){
+                  oldTarget = _targets[ j ];
```
jbuck, line 497: Why not just make this a bool?

src/butter-main.js
```diff
@@ -510,17 +517,20 @@
             if( _media.length > 0 ){
               for( j = 0, jl = _media.length; j < jl; ++j ){
                 if( _media[ j ].id !== medias[ i ].id && _media[ j ].url !== url ){
-                  _this.addMedia({ target: medias[ i ].id, url: url });
+                  oldMedia = _media[ j ];
```
jbuck, line 520: Could make this a bool as well (=true)

src/butter-main.js
secretrobotron, line 497: No real reason. I was just taught to avoid using flags in uni.

src/butter-main.js
secretrobotron, line 520: Same reason as above.

templates/test.html
```diff
@@ -37,7 +37,59 @@
     <video id="main" data-butter="media" controls src="../external/popcorn-js/test/trailer.ogv">
     </video>
     <div class="container-div">
-        <div class="content-div" data-butter="target" id="Area1"></div>
+        <div class="content-div" data-butter="target" id="Area1"><div></div></div>
```
jbuck: Are these changes needed? It looks like the code from Brian's quiz, but we don't need that, right?

PR #204, opened 2012-04-13
Add package target to make.js [ticket 645]

humphd:


make.js
```diff
@@ -12,6 +14,14 @@ target.all = function() {
   target.build();
 };
 
+target.clean = function() {
+  rm('-fr', 'dist');
```
jbuck: You can use DIST_DIR here too

jbuck: And hm, maybe have clean re-create the DIST_DIR as well?

make.js
```diff
@@ -51,3 +61,19 @@ target.server = function() {
...
+
+  cp('-R', 'resources', DIST_DIR);
+  cp('-R', 'dialogs', DIST_DIR);
+  cp('-R', 'config', DIST_DIR);
+  cp('-R', 'editors', DIST_DIR);
+  cp('-R', 'templates', DIST_DIR);
+
+  echo('### Creating butter.zip');
+  cd(DIST_DIR)
+  exec('zip -r ' + PACKAGE_NAME + '.zip ' + ls('.').join(' '));
```
jbuck, line 78: We should find a node compatible zip implemetation that we can use here, but I'm fine with using this for now

PR #206, opened 2012-04-13
Added tooltips for most UI elements [#272]

cadecairos:
Did s bit of work on this. Guessed for most of the tool-tips, but they can be changed if necessary. Let me know if there are any other elements that need tool-tips.

As for tool-tips in the event editors, that might need tool-tip info in the manifests or the tool-tips coded in the custom editors.


jbuck:
This needs rebasing on master

PR #208, opened 2012-04-13
Make text in the default editor slightly larger [#260]

cadecairos:
That extra line change was my editor automatically stripping EOL whitespace...

### DIFF
--- a/editors/default-editor.html
+++ b/editors/default-editor.html
@@ -6,6 +6,9 @@
 <html>
   <head>
     <style>
+      table#table tr  td:nth-child(2) * {
+        font-size: 0.9em;
+      }
       table {
         font-family: helvetica !important;
         color: #fff !important;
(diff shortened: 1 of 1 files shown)


editors/default-editor.html
```diff
@@ -122,6 +123,7 @@
                 newItem.id = item;
                 newItem.style.width = "100%";
                 newItem.value = popcornOptions[ item ];
+                newItem.style.fontSize = "0.9em";
```
secretrobotron: This too.

editors/default-editor.html
```diff
@@ -95,6 +95,7 @@
               if( item === "target" ) {
                 var select = document.createElement( "SELECT" );
                 select.id = "target";
+                select.style.fontSize = "0.9em";
```
secretrobotron: This should probably be in a CSS declaration, not code-wise inline.

PR #211, opened 2012-04-13
T670

secretrobotron:


dseif:
ARGGGHHHHHH

src/dependencies.js
```diff
@@ -0,0 +1,150 @@
...
+          document.head.appendChild( scriptElement );
+          scriptElement.onload = scriptElement.onreadystatechange = function(e){
+            callback();
+          };
+        }
+        else{
+          // keep event-loop behaviour consistent
+          setTimeout(function(){
+            callback();  
+          }, 0);
```
dseif: Why do we need to call this in a setTimeout?

secretrobotron: Just to keep the behaviour of the callback system consistent, so you can expect the event loop to roll over once regardless of whether or not your check function succeeds.

src/dependencies.js
```diff
@@ -0,0 +1,148 @@
+/* This Source Code Form is subject to the terms of the MIT license
+ * If a copy of the MIT license was not distributed with this file, you can
+ * obtain one at http://www.mozillapopcorn.org/butter-license.txt */
+
+define([], function(){
+
+  var DEFAULT_DIRS = {
+        "popcorn-js": "../external/popcorn-js",
+        "css": "css"
```
jbuck, line 9: This doesn't look right to me. The path above looks relative to /src, but this path is relative to / ?

jbuck, line 9: By path above, I mean the popcorn-js path

src/dependencies.js
```diff
@@ -0,0 +1,148 @@
...
+  return function( config ){
+
+    var _configDirs = config.dirs;
+
+    function fixUrl( url ){
+      var replaceVars = url.match( VAR_REGEX );
+      if( replaceVars ){
+        for( var i = 0; i < replaceVars.length; ++i ){
+          var rv = replaceVars[ i ],
+              varOnly = /\{([\w-\._]+)\}/g.exec( rv )[ 1 ],
```
jbuck: Why not use VAR_REGEX?

src/dependencies.js
jbuck: I agree with dseif, this feels weird. Does it matter to the callback whether the event loop rolls over or not?

jbuck:
Also, some lint errors:

### Linting JS files

src/core/page.js: line 52, col 27, Expected an assignment or function call and instead saw an expression.
src/dependencies.js: line 11, col 25, Unescaped '-'.
src/dependencies.js: line 30, col 31, Unescaped '-'.
src/dependencies.js: line 76, col 43, Missing semicolon.
src/dependencies.js: line 98, col 13, Function declarations should not be placed in blocks. Use a function expression or move the statement to the top of the outer function.
src/dependencies.js: line 113, col 13, Function declarations should not be placed in blocks. Use a function expression or move the statement to the top of the outer function.
src/main.js: line 582, col 12, Missing semicolon.
src/main.js: line 617, col 8, Missing semicolon.

src/dependencies.js
secretrobotron: I actually did try to do that, ... and it returned different results...

I'll try again though because you're right.

src/dependencies.js
secretrobotron, line 9: It's most likely relative to `templates/` right now, which will be the default setup for those who don't care.

src/dependencies.js
secretrobotron: No, I suppose not. Consider it removed.

src/dependencies.js
secretrobotron: Yeah, I have no idea what was happening. Problem doesn't seem to be there anymore. Herp derp computers.

PR #212, opened 2012-04-14
T658

brianchirls:
Lighthouse ticket:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/658-export-complex-data-types-in-popcorn-event-options-handle-crlf-in-value

If Popcorn options for a given event contains arrays, objects or strings with weird characters like CR or LF, this patch will save it properly as JSON.  Circular object errors are caught, and that event is excluded.


src/core/popcorn-wrapper.js
```diff
@@ -264,20 +260,33 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
...
             }
-            if ( popcornString[ popcornString.length - 1 ] === "," ) {
-              popcornString = popcornString.substring( 0, popcornString.length - 1 );
+
+            //stringify will throw an error on circular data structures
+            try {
+              //pretty print with 2 spaces per indent
+              optionString = JSON.stringify( saveOptions, null, 2 );
+            } catch ( jsonError ) {
+              optionString = false;
```
dseif, line 289: If we fail here we should probably log something, maybe do something like https://github.com/brianchirls/butter/blob/3e77abac543722b28ec14af96339b86181ab2c22/src/core/popcorn-wrapper.js#L143

src/core/popcorn-wrapper.js
```diff
@@ -264,20 +260,34 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
...
+              _logger.log( "WARNING: Unable to export event options: \n" + jsonError.message );
```
brianchirls, line 290: dseif: Good idea. Here is the logged error.

src/core/popcorn-wrapper.js
```diff
@@ -264,20 +260,34 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
       if ( _popcorn ) {
 
         // gather and serialize existing trackevents
-        var trackEvents = _popcorn.getTrackEvents();
+        var trackEvents = _popcorn.getTrackEvents(),
+          trackEvent, optionString, saveOptions;
```
secretrobotron: Extra spacing here to keep our style consistent please. It might be terrible, but at least it's (mostly) all the same.

brianchirls: Can you be more specific? Do you mean a new line for each declared variable?

This section has a number of declared variables further down. Is it worth taking this opportunity to move them all up to the top?

secretrobotron: Not really necessary to gather them all here. You can if you want to though. I'm just talking about the tab spacing. Just trying to keep them all in line. Let me find an example...

Got one: https://github.com/mozilla/butter/blob/master/src/main.js#L42

brianchirls: Ah, okay. That makes sense. Will move the variables up and space them as in that example. Stay tuned...

PR #216, opened 2012-04-17
Added done and cancel buttons to the default editor, cancel functionality

k88hudson:


editors/default-editor.html
```diff
@@ -132,7 +137,44 @@
...
+              }
+            })(document.querySelectorAll("button"));
+            
           });
+          
+          //Helper
+          function hasClass(element, aClass) {
+            var result = element.className.match(new RegExp('(\\s|^)'+ aClass +'(\\s|$)'));
+            return result;
+          }
```
dseif: Instead of doing this and using querySelectorAll for button above, we can probably use document.getElementsByClassName( "done-button" ); and the same for cancel button.  getElementsByClassName returns an array of elements which have the given class.

jbuck:
Obsoleted by something I'm sure

PR #219, opened 2012-04-17
[#592]All iFrames now have focus upon opening

mjschranz:
Fixes the issues in another ticket about iFrames and esc in general.

### DIFF
--- a/src/dialog/iframe-dialog.js
+++ b/src/dialog/iframe-dialog.js
@@ -87,6 +87,10 @@ define( [
             _this.send( popped.type, popped.data );
           } //while
           _open = true;
+          // Immediately set focus
+          setTimeout( function() {
+            _this.focus();
+          }, 0 );
           _em.dispatch( "open" );
         });
       }, false );
@@ -107,6 +111,10 @@ define( [
         _commQueue.push({ type: type, data: data });
       } //if
     }; //send
+
+    this.focus = function() {
+      _iframe.contentWindow.focus();
+    }; //focus
 
     Object.defineProperties( this, {
       iframe: {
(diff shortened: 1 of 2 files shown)


src/dialog/iframe-dialog.js
```diff
@@ -108,6 +112,10 @@ define( [
       } //if
     }; //send
 
+    this.focus = function() {
```
secretrobotron, line 115: Does anything use this function?

src/dialog/iframe-dialog.js
```diff
@@ -87,6 +87,10 @@ define( [
             _this.send( popped.type, popped.data );
           } //while
           _open = true;
+          // Immediately set focus
+          setTimeout( function() {
```
secretrobotron, line 91: Did we resolve that this needs to be in a setTimeout? I can't remember. I'd prefer it not to be, but... if we must, we must.

src/dialog/iframe-dialog.js
mjschranz, line 115: As far as I know, nothing does now since I removed the one spot that used it previously in editors/editor.js

src/dialog/iframe-dialog.js
mjschranz, line 91: Without it being in the setTimeout the editor dialog iframe was never becoming the active element when calling focus on _iframe.contentWindow.focus()

ghost:
#### From: Bot.io (butyr.org)

---

# Help

Available commands:

```
preview
test
```

ghost:
#### From: Bot.io (butyr.org)

---

# Help

Available commands:

```
preview
test
```

PR #221, opened 2012-04-17
Output the bindIP from config [#595]

jbuck:


cornfield/app.js
```diff
@@ -263,6 +263,6 @@ app.post('/project/:id?', function( req, res ) {
 
 app.listen(CONFIG.server.bindPort, CONFIG.server.bindIP, function() {
   var addy = app.address();
-  console.log('Server started on http://' + addy.address + ':' + addy.port);
+  console.log('Server started on http://' + CONFIG.server.bindIP + ':' + CONFIG.server.bindPort);
```
secretrobotron, line 266: Is this necessary/better? If the output is always going to be the same anyway, disregard this comment.

jbuck, line 266: By changing to `CONFIG.server.bindIP` it'll output `localhost` as opposed to `127.0.0.1`, which is what was tripping up dseif

secretrobotron, line 266: That's frustrating. I thought just changing the bindIP to localhost was enough.

PR #222, opened 2012-04-17
Work around RegExp inconsistencies across browser [#718]

jbuck:

PR #223, opened 2012-04-18
[t718] Moved regex to work around chrome bug

secretrobotron:


jbuck:
Closing in favour of pull request #222

PR #229, opened 2012-04-23
Added new track creation upon dropping a plugin outside an existing track [#540]

jbuck:


src/timeline/media.js
```diff
@@ -54,6 +56,24 @@ define( [
     _rootElement.id = "media-instance" + media.id;
     _container.className = "media-container";
 
+    DragNDrop.droppable( _container, {
+      hoverClass: "draggable-hover",
```
secretrobotron: It's a good idea to have a CSS definition to go along with this class to signify to the user that you can drop things on the empty space. Perhaps a simple white inset box-shadow.

PR #237, opened 2012-04-25
T721

secretrobotron:


src/cornfield/module.js
```diff
@@ -10,6 +10,10 @@ define(['util/xhr'], function(XHR) {
 
   var Cornfield = function( butter, config ) {
 
+    if( !config ){
+      return;
+    }
```
dseif, line 15: Should we warn the user about there not being a config?

secretrobotron, line 15: We could. Might just be extra noise though, since it's not really a requirement for butter itself to work as advertised.

PR #250, opened 2012-04-30
[t710] Media loader disappears

secretrobotron:
1. Added constructing event to popcorn-wrapper and media
2. View is updated as several places to reflect different states of media loading
3. PositionTracker can be destroyed properly now, and is.

### DIFF
--- a/src/ui/position-tracker.js
+++ b/src/ui/position-tracker.js
@@ -3,8 +3,10 @@ define([], function(){
   var POLL_INTERVAL = 20;
   
   return function( object, movedCallback ){
-    var _rect = {};
-    setInterval( function(){
+    var _rect = {},
+        interval;
+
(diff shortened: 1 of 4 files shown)


src/ui/position-tracker.js
```diff
@@ -3,8 +3,10 @@ define([], function(){
   var POLL_INTERVAL = 20;
   
   return function( object, movedCallback ){
-    var _rect = {};
-    setInterval( function(){
+    var _rect = {},
+        interval;
+
+    interval = setInterval( function(){
```
jbuck, line 9: A timer inside a constructor seems a little awful

secretrobotron, line 9: This isn't new though. It's just how this little helper guy works. It's been here for a while. I'm just saving the interval now so it can be stopped :P.

PR #251, opened 2012-04-30
Added complete config files and testing page

k88hudson:


config/complete.conf
```diff
@@ -0,0 +1,145 @@
...
+    "frameAnimation": true
+  },
+  "popcorn": {
+    "scripts": {
+      "init": "test-init.js",
+      "beforeEvents": "test-before.js",
+      "afterEvents": "test-after.js"
+    },
+    "callbacks": {
+      "init": "_testInitCallback",
```
secretrobotron: These will just cause silly export errors. Let's remove 'em.

templates/complete.js
```diff
@@ -0,0 +1,52 @@
...
+      }
+
+      media.onReady( start );
+      
+      window.butter = butter;
+    } 
+  }); //Butter
+}, false );
+
+_testInitCallback = function(){
```
secretrobotron: Same as above. These are useless here and will just give us errors.

secretrobotron: Can these be removed too?

k88hudson: Yep

templates/complete.js
```diff
@@ -0,0 +1,199 @@
...
+            gmltag: "29582",
+            target: "Area2"
+          }
+        });
+
+      }
+
+      media.onReady( start );
+      
+      window.butter = butter;
```
jbuck: We shouldn't pollute the global scope here

templates/complete.js
```diff
@@ -0,0 +1,199 @@
...
+          }
+        });
+
+        butter.tracks[ 0 ].addTrackEvent({ 
+          type: "webpage",
+          popcornOptions: {
+            start:5,
+            end: 6,
+            id:"webpages-a",
+            src:"http://www.mozilla.org/",
```
jbuck: Change this to http://popcornjs.org/ ; mozlla.org sends https://developer.mozilla.org/en/The_X-FRAME-OPTIONS_response_header which means that it can't be embedded

PR #261, opened 2012-05-02
T766

secretrobotron:


src/core/page.js
```diff
@@ -55,33 +55,46 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
...
+      }
+      node.parentNode.replaceChild( newNode, node );
 
       toClean = body.querySelectorAll( "*[butter-clean=\"true\"]" );
       for ( i = 0, l = toClean.length; i < l; ++i ) {
         node = toClean[ i ];
+
         node.removeAttribute( "butter-clean" );
-        node.removeAttribute( "data-butter" );
+        node.removeAttribute( "`er" );
```
jbuck: that doesn't look right

src/core/page.js
```diff
@@ -55,33 +55,46 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
...
     this.getHTML = function( popcornStrings ){
-      var html, head, body, i, toClean, toExclude, node, base;
+      var html, head, body, i, toClean, toExclude, node, newNode, base, mediaElements;
 
       //html tag to which body and head are appended below
       html = document.createElement( "html" );
 
       // if there is already a snapshot, clone it instead of cloning the current dom
       if( !_snapshot ){
-        head = document.getElementsByTagName( "head" )[ 0 ].cloneNode( true );
```
jbuck, line 65: Any reason you want to get rid of the head cloning here? What if the template is doing something like... loading jquery?

src/core/page.js
```diff
@@ -55,33 +55,46 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
...
         body = document.getElementsByTagName( "body" )[ 0 ].cloneNode( true );
       }
-      else{
-        head = _snapshot.head.cloneNode( true );
+      else {
         body = _snapshot.body.cloneNode( true );
       }
+      
+      head = document.getElementsByTagName( "head" )[ 0 ].cloneNode( true );
```
jbuck, line 71: Oh, you clone it down here. Whoops, sorry!

src/core/page.js
secretrobotron: Hahahahh I don't even know... what happened.

PR #271, opened 2012-05-04
[t1071] Dispatch "getHTML" event on Pages to allow modification of exported document

banksJeremy:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/1071-allow-the-snapshotted-dom-to-be-manipulated

A single event with the html element that's about to be converted to HTML is probably enough.

### DIFF
--- a/src/core/page.js
+++ b/src/core/page.js
@@ -130,6 +130,8 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
         } //for
       } //if
 
+      _eventManager.dispatch( "getHTML", html );
+
       return "<html>" + html.innerHTML + "</html>";
     }; //getHTML
 
--- a/test/core.js
+++ b/test/core.js
@@ -668,7 +668,7 @@
       });
 
       butter.listen( "mediaready", function( e ) {
-        equals( /test text at end of body\s*<\/body>/.test( butter.getHTML() ), true, "Text appended to body is missing from exported HTML." );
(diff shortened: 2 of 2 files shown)


src/core/page.js
```diff
@@ -130,6 +130,8 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
         } //for
       } //if
 
+      _eventManager.dispatch( "getHTML", html );
```
secretrobotron, line 133: I'd argue that this isn't enough actually. Instead, why not put this in `main.js` where it can be emitted from the Butter object and will have the correct `<html>` wrapper already.

banksJeremy, line 133: Hmm... after adding the `<html>` wrapper we're only left with a string. Giving users a DOM tree to manipulate seems much more friendly than forcing them to deal with HTML source.

secretrobotron, line 133: I thought you would have preferred just doing string replacements. But, if it's DOM you want, this is good.

ghost:
#### From: Bot.io (butyr.org)

---

# Received

Command **cmd_check** from @jbuck received. Current queue size: 0

Live output at: http://butyr.org:8000/fbdd7398f8e3ff6/output.txt

ghost:
#### From: Bot.io (butyr.org)

---

# Success

Full output at http://butyr.org:8000/fbdd7398f8e3ff6/output.txt

Total script time: 0.33 mins

Checking 48fe7717de4a96aa126c67bad7351b2478065779 of pull request #271

PR #273, opened 2012-05-07
T241 - documentation automation

secretrobotron:
Dox.py for automating documentation generation. Examples in `core/logger`, `core/comm`, and `core/trackevent`.


src/core/trackevent.js
```diff
@@ -148,6 +194,13 @@ define( [
           _this.update( _popcornOptions );
         }
       },
+
+      /**
+       * Property: view
+       *
+       * A reference to the view object generated for this TrackEvent.
+       * @malleable: No.
```
jbuck, line 202: What a weird keyword...

PR #286, opened 2012-05-09
[t628] initial dom fragment work

secretrobotron:
1. Changed media-view.js to use DOMFragment
2. DOMFragment module
3. text require.js plugin
4. layouts folder, with media-view.xml


make.js
```diff
@@ -36,6 +36,9 @@ target['check-lint'] = function() {
   echo('### Linting JS files');
 
   var files = find('src').filter( function( file ) {
+    if(file === "src/text.js"){
```
jbuck: Move the require.js text plugin out of src/text.js so that we don't need to include this filter in the makefile

src/core/views/media-view.js
```diff
@@ -1,50 +1,25 @@
-define( [ "ui/page-element", "ui/logo-spinner" ], function( PageElement, LogoSpinner ){
-  
+define( [ "ui/page-element", "ui/logo-spinner", "util/dom-fragment", "text!layouts/media-view.xml" ],
```
jbuck: Change the extension to .html instead of .xml

src/util/dom-fragment.js
```diff
@@ -0,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the MIT license
+ * If a copy of the MIT license was not distributed with this file, you can
+ * obtain one at http://www.mozillapopcorn.org/butter-license.txt */
+
+define( [], function() {
+
+  return function( inputString ) {
+    var renderContainer = document.createElement( "div" );
+    renderContainer.innerHTML = inputString;
```
jbuck: Use createContextualFragment instead; then you don't need to worry about removing this wrapper div

PR #293, opened 2012-05-15
[#619] Cleaned up exported JSON

dseif:


dialogs/export.html
```diff
@@ -30,13 +31,60 @@
           jsonExport.style.display = "none";
         }, false );
 
+        function removeComma( string ) {
```
secretrobotron: At first, it looks like `removeComma` is following a functional approach: not changing the input. But, it turns out that the `string` variable is probably not even required here if you're just manipulating `tempJSON` directly.

So, let's choose one or the other. Either return the morphed string from this function or remove the `string` variable and just manipulate `tempJSON` directly.

dseif: Ah good call, I'll remove the string parameter

dialogs/export.html
```diff
@@ -30,13 +31,60 @@
...
+          tempJSON = string.substring( 0, string.length - 2 ) + "\n";
+        }
+
+        function jsonBeautifier( json, padding, lineStart ) {
```
secretrobotron: Instead of this guy, can we just use the built-in function? http://stackoverflow.com/questions/2614862/how-can-i-beautify-json-programmatically

dseif: That's overpowered, thank you for showing me this.

dialogs/export.html
```diff
@@ -30,13 +31,60 @@
...
           }
-          catch( e ){
-            jsonExport.value = "There was an error trying to parse the JSON blob for this project. Please file a bug at https://webmademovies.lighthouseapp.com/projects/65733-butter/ and let us know.";
+          catch( e ) {
+            tempJSON = "There was an error trying to parse the JSON blob for this project. Please file a bug at https://webmademovies.lighthouseapp.com/projects/65733-butter/ and let us know.";
+            ok = false;
           }
+
+          if ( ok ) {
+            tempJSON = "";
```
secretrobotron: Is there a reason the constituents of this branch block can't just be placed into the try/catch block above?

PR #298, opened 2012-05-16
Add build step for stamping Butter with git commit sha version info.

humphd:


make.js
```diff
@@ -75,6 +75,13 @@ target.build = function() {
   exec(RJS + ' -o tools/build.js');
   exec(RJS + ' -o tools/build.optimized.js');
 
+  // Stamp Butter.version with the git commit sha we are using
+  var version = exec('git show -s --pretty=format:%h',
+                     {silent:true}).output.replace(/\r?\n/m, "");
+  console.log("'" + version + "'");
```
jbuck: I don't think we need this here, do we?

make.js
```diff
@@ -75,6 +75,13 @@ target.build = function() {
   exec(RJS + ' -o tools/build.js');
   exec(RJS + ' -o tools/build.optimized.js');
 
+  // Stamp Butter.version with the git commit sha we are using
+  var version = exec('git show -s --pretty=format:%h',
```
jbuck: Instead of using `git show`, you can use `git describe` which output looks like this: `v0.4-81-g24d7d1b`

It gives you 3 pieces of information: the tag that the commit is based off, how many commits have been made since that tag, and the commit sha.

jbuck:
Do you think we should have a VERSION override usable from the command line, or do you think the output of `git describe` is good enough for our purposes, including a tagged release?

PR #303, opened 2012-05-18
BUG: Default loaded trackEvents were not selectable by clicking.

pankajp:
The trackeventmouse{down,out,over} events are dispatched by the
trackEvent.view and not trackEvent.


jbuck:
@secretrobotron Can you take a look at this, and see if it's still a bug?

jbuck:
Closing due to age

PR #329, opened 2012-05-25
[t1212] - Auto load saved data

secretrobotron:
1. Removed majority of templates/test.js
2. Added XHR for saved data and default config option to specify where it is
3. Added saved-data.json

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -512,59 +512,62 @@
             medias = scrapedObject.media;
 
         _page.prepare(function() {
-          var i, j, il, jl, url, oldTarget, oldMedia, mediaPopcornOptions, mediaObj;
-          for( i = 0, il = targets.length; i < il; ++i ) {
-            oldTarget = null;
-            if( _targets.length > 0 ){
-              for( j = 0, jl = _targets.length; j < jl; ++j ){
-                // don't add the same target twice
-                if( _targets[ j ].id === targets[ i ].id ){
-                  oldTarget = _targets[ j ];
-                  break;
-                } //if
-              } //for j
-            }
-
-            if( !oldTarget ){
-              _this.addTarget({ element: targets[ i ].id });
-            }
-          }
-
-          for( i = 0, il = medias.length; i < il; i++ ) {
-            oldMedia = null;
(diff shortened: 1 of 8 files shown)


src/main.js
```diff
@@ -661,6 +661,38 @@
         }
       }
 
+      function attemptDataLoad( finishedCallback ){
+        var xhr = new XMLHttpRequest(),
```
jbuck: Why not use my XHR module?

secretrobotron: Because it wasn't as robust as the code that Dave added for loading configs, which was incredibly simple to copy & paste from right below this code. Perhaps we can merge that code in for a better XHR wrapper.

test/core.js
```diff
@@ -857,4 +859,19 @@
     });
   });
 
+  asyncTest( "Auto-load saved data", 4, function(){
+
+    Butter({
+      config: "test-config-auto-load.json",
+      ready: function( butter ){
+        start();
```
mjschranz: this call should be after all the tests right?

secretrobotron: Actually, I'm not sure this matters. I checked some documentation, and nowhere does it say `start()` must be in a specific place.

Also (from start function in qunit.js):

``` javascript
        // A slight delay, to avoid any current callbacks
        if ( defined.setTimeout ) {
            window.setTimeout(function() {
                if ( config.timeout ) {
                    clearTimeout(config.timeout);
                }

                config.blocking = false;
                process();
            }, 13);
        } else {
            config.blocking = false;
            process();
        }
```

secretrobotron: To make it easy on the eyes though, I'll certainly change it :).

ghost:
#### From: Bot.io (butyr.org)

---

# Received

Command **cmd_check** from @dseif received. Current queue size: 0

Live output at: http://botio.butyr.org:8000/ead87e387260871/output.txt

ghost:
#### From: Bot.io (butyr.org)

---

# Success

Full output at http://botio.butyr.org:8000/ead87e387260871/output.txt

Total script time: 0.58 mins

PR #332, opened 2012-05-28
[#1267] Cornfield tests fail gracefully without manual input. Fixed Styl...

mjschranz:
...e issues

https://webmademovies.lighthouseapp.com/projects/65733-popcorn-maker/tickets/1267-allow-cornfield-tests-to-fail-gracefully-if-no-manual-input-is-received

### DIFF
--- a/test/io/cornfield.js
+++ b/test/io/cornfield.js
@@ -8,10 +8,10 @@
 
   QUnit.config.reorder = false;
 
-  document.addEventListener( "DOMContentLoaded", function( e ){
+  document.addEventListener( "DOMContentLoaded", function( e ) {
     var b = new Butter({
       config: "cornfield-test-config.json",
-        ready: function( butter ){
+        ready: function( butter ) {
           var filename = "test" + Date.now(),
               data = {
                 name: filename,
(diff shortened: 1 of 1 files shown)


test/io/cornfield.js
```diff
@@ -19,25 +19,25 @@
...
                 template: "pop"
               },
-              stringedData = JSON.stringify(data);
+              stringedData = JSON.stringify( data );
 
           module( "Unauthenticated tests" );
 
           test( "Sync API", 1, function() {
-            ok( !butter.cornfield.user(), 'Username is ""' );
+            ok( !butter.cornfield.user(), "Username is \"\"" );
```
jbuck: This looks ugly. And is wrong too; change it to Username is undefined

test/io/cornfield.js
```diff
@@ -92,22 +91,27 @@
...
+                  butter.cornfield.load( filename, function( res ) {
+                    deepEqual( JSON.parse( res.project ), data.data, "The project is the same" );
 
                     start();
                   });
                 });
               });
             });
+
+            var failSafe = setTimeout( function() {
```
jbuck, line 111: There's no manual input required here, so why do we need the timeout?

PR #339, opened 2012-05-28
[#1209] Plugin scripts no longer load again if it alraedy exists

dseif:


dseif:
Accidentally created another pull request

PR #354, opened 2012-05-29
Reading the token from window.CSRF_TOKEN instead of parsing

alfredo:
Bugfix for: https://webmademovies.lighthouseapp.com/projects/65733/tickets/1293-xhr-library-is-not-sending-the-csrf-token#ticket-1293-1


alfredo:
Closing since there is a better fix for this.

PR #363, opened 2012-05-30
[t1230] iframe dialogs removed

secretrobotron:
1. Wrote new dialog.js to replace IFrameDialog and WindowDialog
2. Removed IFrameDialog and WindowDialog
3. Morphed IFrameDialog slightly specifically for editor module
4. Moved dialogs into src/
5. Ported dialog iframe code to new system
6. Fixed up html/css
7. Fixed try/catch in cornfield which caused debug hilarity and confusion


ghost:
#### From: Bot.io (butyr.org)

---

# Received

Command **cmd_check** from @dseif received. Current queue size: 0

Live output at: http://botio.butyr.org:8000/ae59d0ee3ebbd9d/output.txt

ghost:
#### From: Bot.io (butyr.org)

---

# Failed

Full output at http://botio.butyr.org:8000/ae59d0ee3ebbd9d/output.txt

Total script time: 1.33 mins

dseif:
Don't know why the robot is whining, tested locally and it works fine

src/dialog/dialog.js
```diff
@@ -0,0 +1,186 @@
...
+          _rootElement = LangUtils.domFragment( layoutSrc ).querySelector( ".butter-dialog" ),
+          _enterKeyActivity,
+          _escapeKeyActivity,
+          _modal;
+
+      function onKeyDown( e ) {
+        if ( _enterKeyActivity && __keyboardAvoidElements.indexOf( e.target.nodeName ) === -1 && ( e.which === 13 || e.keyCode === 13 ) ) {
+          _activities[ _enterKeyActivity ]( e );
+        }
+        else if ( _escapeKeyActivity && __keyboardAvoidElements.indexOf( e.target.nodeName ) === -1 && ( e.which === 27 || e.keyCode === 27 ) ) {
```
dseif: can we split this line and line 27 up a bit

src/dialog/dialog.js
```diff
@@ -0,0 +1,186 @@
...
+          _activities[ _escapeKeyActivity ]( e );
+        }
+      }
+
+      var _internal = {
+        rootElement: _rootElement,
```
dseif: does this ever get used? Looks like we can just use _rootElement

src/dialog/dialog.js
```diff
@@ -0,0 +1,186 @@
...
+        },
+
+        enableElements: function(){
+          for( var i=0; i<arguments.length; ++i ){
+            _rootElement.querySelector( arguments[ i ] ).removeAttribute( "disabled" );
+          }
+        },
+
+        disableElements: function(){
+          for( var i=0; i<arguments.length; ++i ){
```
dseif: can store a reference to arguments.length for some extra goodness, on line 81 as well

src/dialog/dialog.js
```diff
@@ -0,0 +1,186 @@
...
+        open: function () {
+          for( var e in _listeners ){
+            if( _listeners.hasOwnProperty( e ) ){
+              _external.listen( e, _listeners[ e ] );
+            }
+          }
+          _modal = new Modal( _rootElement );
+          setTimeout( function() {
+            _external.focus();
+          }, 0 );
```
dseif, line 264: Why do we need to do the setTimeout 0, is there a way around this?

src/dialog/dialog.js
```diff
@@ -0,0 +1,186 @@
...
+        _external.dispatch( "submit" );
+        _external.close();
+      });
+
+      dialogCtor( _internal, spawnOptions.data );
+
+      // make this happen a tiny bit later so variables get initialized before opening
+      setTimeout( function(){
+        _external.open();
+      }, 10 );
```
dseif: anyway we could do this in a callback instead?

src/dialog/dialog.js
```diff
@@ -0,0 +1,186 @@
...
+        else {
+          __openDialogs[ name ] = __dialogs[ name ]( spawnOptions );
+          __openDialogs[ name ].listen( "close", function () {
+            __openDialogs[ name ] = null;
+          });
+        }
+        return __openDialogs[ name ];
+      }
+      else {
+        throw "Dialog '" + name + "' does no exist.";
```
dseif: should this be throw new error?

src/dialog/dialog.js
```diff
@@ -0,0 +1,186 @@
...
+      }
+    }
+
```
dseif: get rid of this whitespace

src/dialog/dialog.js
```diff
@@ -0,0 +1,186 @@
...
+  };
+
```
dseif: same here

src/dialog/dialogs/delete-track.html
```diff
@@ -0,0 +1,16 @@
...
+    <div class="content vbox center">
+      <h1>Are you sure you want to delete <span class="track-name"></span>?</h1>
+    </div>
+  </div>
+  <div class="buttons" class="vbox center">
+    <button class="yes">Yes</button>
+    <button class="no">No</button>
+  </div>
+  <div class="close-button"></div>
+</div>
```
dseif: This is soooo much better than iframes, is the plan to move the editors to something like this as well?

src/dialog/dialog.js
dseif: never mind I see it being used later in the actual dialogs, my bad.

src/dialog/dialogs/export.js
```diff
@@ -0,0 +1,44 @@
...
+define([ "text!dialog/dialogs/export.html", "dialog/dialog" ],
+  function( LAYOUT_SRC, Dialog ){
+
+  Dialog.register( "export", LAYOUT_SRC, function( dialog, exportData ) {
+
+    var jsonButton = dialog.rootElement.querySelector( ".json-button" ),
+        htmlButton = dialog.rootElement.querySelector( ".html-button" ),
+        jsonExport = dialog.rootElement.querySelector( ".json-export" ),
+        htmlExport = dialog.rootElement.querySelector( ".html-export" ),
+        title = dialog.rootElement.querySelector( ".title" );
```
dseif: can you create a reference to dialog.rootElement

src/dialog/dialogs/delete-track.js
```diff
@@ -0,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the MIT license
+ * If a copy of the MIT license was not distributed with this file, you can
+ * obtain one at http://www.mozillapopcorn.org/butter-license.txt */
+
+define([ "text!dialog/dialogs/delete-track.html", "dialog/dialog" ],
```
dseif, line 5: can you explain the text!dialog part of this, I don't get what it's doing

src/dialog/dialogs/load.js
```diff
@@ -0,0 +1,31 @@
...
+    var list = dialog.rootElement.querySelector( ".project-list" );
+
+    dialog.registerActivity( "load", function( e ){
+      dialog.send( "submit", {
+        id: list.value,
+        name: list.querySelectorAll( 'option' )[ list.selectedIndex ].innerHTML
+      });
+    });
+
+    for( var i=0; i<projects.length; ++i ){
```
dseif: store reference to projects.length

src/dialog/dialogs/delete-track.js
dseif, line 5: does that mean it get's returned as plain text as opposed to an object?

src/dialog/dialogs/save-as.html
```diff
@@ -0,0 +1,16 @@
+<!--  This Source Code Form is subject to the terms of the MIT license
+      If a copy of the MIT license was not distributed with this file, you can
+      obtain one at http://www.mozillapopcorn.org/butter-license.txt -->
+
+<div class="butter-dialog">
+  <h1>Please give your project a<span class="better"> better</span> name:</h1>
```
dseif: haha what is the span there for?

src/dialog/dialogs/save-as.js
```diff
@@ -0,0 +1,27 @@
...
+
+  Dialog.register( "save-as", LAYOUT_SRC, function( dialog, name ) {
+    var nameInput = dialog.rootElement.querySelector( ".name-input" );
+
+    dialog.registerActivity( "save", function( e ){
+      if( nameInput.value.replace( /\s/g, "" ) !== "" ){
+        dialog.send( "submit", nameInput.value );
+      }
+      else{
+        dialog.rootElement.querySelector( ".better" ).style.display = "inline";
```
dseif, line 16: ah I see what that span was being used for now

src/dialog/dialogs/track-data.js
```diff
@@ -0,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the MIT license
+ * If a copy of the MIT license was not distributed with this file, you can
+ * obtain one at http://www.mozillapopcorn.org/butter-license.txt */
+
+define([ "text!dialog/dialogs/track-data.html", "dialog/dialog" ],
+  function( LAYOUT_SRC, Dialog ){
+
+  Dialog.register( "track-data", LAYOUT_SRC, function ( dialog, track ) {
+    var trackName = dialog.rootElement.querySelector( ".track-name" ),
+        trackData = dialog.rootElement.querySelector( ".track-data" );
```
dseif: store reference to rootElement please

src/dialog/modal.js
```diff
@@ -6,34 +6,30 @@ define( [], function(){
...
-    } //if
-
+    _element.classList.add( "layer" );
     __container.appendChild( _element );
 
     // need to wait an event-loop cycle to apply this class
     // ow, opacity transition fails to render
     setTimeout( function(){
-      _element.className += " fade-in";
+      _element.classList.add( "fade-in" );
```
dseif, line 24: we should probably use our shim for this so once we add IE9 support we don't run into issues

src/dialogs.js
```diff
@@ -0,0 +1,25 @@
...
+    "delete-track",
+    "export",
+    "quit-confirmation",
+    "save-as",
+    "load",
+    "share"
+  ];
+
+  var include = [];
+  for ( var i = 0; i < dialogs.length; ++i ) {
```
dseif: store reference to dialogs.length

src/ui/header.js
```diff
@@ -129,14 +110,9 @@ define( [ "dialog/iframe-dialog" ], function( IFrameDialog ){
     });
 
     function showErrorDialog( message, callback ){
-      var dialog = new IFrameDialog({
-        type: "iframe",
-        modal: true,
-        url: butter.ui.dialogDir + "error-message.html",
+      var dialog =Dialog.spawn( "error-message", {
```
dseif: add space after =

dseif:
When you open the load dialog, the "x" button in the corner is just a blue box

dseif:
Going to need tests for this, don't know if you want to write them now or do what Humph mentioned in an earlier email and stage this then set a tag to "needs-tests" or whatever.

src/dialog/dialog.js
secretrobotron: Two things to consider here:

1) I didn't rewrite this code, I only ported it, mostly through copy & paste.
2) While I do like optimizing, these are probably instances where it will just become more confusing to read, rather than add a whole lot of value, no?

src/dialog/dialog.js
secretrobotron, line 264: Like I said in the comment above, I've just copied and pasted most of this code. However, I can see why this might be here: to let the user-agent do what it needs to do before applying the changes from `focus()`. I think we added this to prevent some problems with the user-agent grabbing focus improperly otherwise. Regardless, I don't see problems with these kind of `setTimout` blocks because they're usually UI-oriented.

src/dialog/dialog.js
secretrobotron: How do you mean?

src/dialog/dialog.js
secretrobotron: I thought so too, at first, but it doesn't look like it's necessary: https://developer.mozilla.org/en/JavaScript/Reference/Statements/throw

src/dialog/dialogs/delete-track.html
secretrobotron: Yep :D.

src/dialog/dialogs/delete-track.js
secretrobotron, line 5: Yeah, it's how the layout system works. `text!` will deliver a piece of text and not try to run it as code. Great for including HTML in your projects :D.

src/dialog/dialogs/load.js
secretrobotron: Again, I'd like to make the case that readability is good here. This doesn't need to be a hot loop, right?

src/dialog/dialogs/save-as.html
secretrobotron: Try saving with a blank filename :). Again, copy & pasted.

src/dialog/dialogs/save-as.js
secretrobotron, line 16: Perfect :D.

src/dialog/modal.js
secretrobotron, line 24: It should already be included, actually. To be absolutely certain though probe Chris about the status of this, https://webmademovies.lighthouseapp.com/projects/65733/tickets/1231-insert-shims-on-startup.

src/dialogs.js
secretrobotron: Same as above. This gets run once, and there aren't thousands of dialogs. Hot loop necessary?

secretrobotron:
Aye. I'll add tests.

secretrobotron:
Did you get that blue box in Chrome? FF? Both? It happened for me too, but my gfx card was already flipping out. Couldn't tell if it was just Chrome or not.

src/dialog/dialog.js
dseif: If you didn't write the code then it's not a huge deal.

src/dialog/dialog.js
dseif, line 264: Same as above, if you didn't write the code then this is fine.

src/dialog/dialog.js
dseif: Is there a way to call `_external.open(); after the variables get initialized? ( reading comment ). If this is another case of you not writing this code then ignore this comment.

src/dialog/dialog.js
dseif: Cool

src/dialog/dialogs/delete-track.js
dseif, line 5: Yea that's pretty awesome

src/dialog/dialogs/load.js
dseif: I dunno `i = 0, l = projects.length; i < l;` is still pretty readable. Hot loop? If this is another case of you not writing this code then ignore it.

src/dialog/modal.js
dseif, line 24: Yea I stole the ticket, will review after this

dseif:
It was in Chrome that I got the blue box, will try again and see if I can reproduce it.

ghost:
#### From: Bot.io (butyr.org)

---

# Received

Command **cmd_check** from @dseif received. Current queue size: 0

Live output at: http://botio.butyr.org:8000/973f183196eacff/output.txt

ghost:
#### From: Bot.io (butyr.org)

---

# Success

Full output at http://botio.butyr.org:8000/973f183196eacff/output.txt

Total script time: 0.33 mins

ghost:
#### From: Bot.io (butyr.org)

---

# Received

Command **cmd_preview** from @dseif received. Current queue size: 0

Live output at: http://botio.butyr.org:8000/a3a6cb4ed58814e/output.txt

ghost:
#### From: Bot.io (butyr.org)

---

# Success

Full output at http://botio.butyr.org:8000/a3a6cb4ed58814e/output.txt

Total script time: 0.62 mins

Server is running on http://pr363.butyr.org/templates/test.html

dseif:
Yea the blue box issue on the load dialog looks to be fixed

src/dialog/dialog.js
secretrobotron: No, I think you're right in pointing this out, but I'm not sure it's bad/wrong. My intent was for you to be able to call `spawn()` and just have the dialog appear without waiting for it to initialize. Think that's too confusing?

dseif: I like the idea of calling `spawn()` and it opening the dialog for us, I think that's awesome. I think the setTimeout bothers me because 10 milliseconds just seems like an arbitrary number and gives us no guarantee that everything will be ready in time.

secretrobotron: What about this? https://developer.mozilla.org/en/DOM/window.setTimeout#Minimum_delay_and_timeout_nesting

I just picked `10` because it's ... yeah. Probably should be `0`.

dseif: Alright cool if this is just being done so it doesn't block the main thread then +1. After changing it to 0 this can go to SR

ghost:
#### From: Bot.io (butyr.org)

---

# Received

Command **cmd_preview** from @dseif received. Current queue size: 0

Live output at: http://botio.butyr.org:8000/dfb520adf3d4353/output.txt

ghost:
#### From: Bot.io (butyr.org)

---

# Success

Full output at http://botio.butyr.org:8000/dfb520adf3d4353/output.txt

Total script time: 0.93 mins

Server is running on http://pr363.butyr.org/templates/test.html

css/dialog.css
```diff
@@ -61,11 +34,11 @@ html, body {
...
...GgoAAAANSUhEUgAAAC4AAAArCAYAAAAHdLqEAAAAGXRFWHRTb2Z0d2FyZQBBZG9iZSBJbWFnZVJlYWR5ccllPAAAAylJREFUeNrUWTFvEzEUdiIWkFoVtUuhlZhgYLkwIMGSIpAYEJAOULYmUvgBRazQCz8gaRYYiJSObVkCqAMSqO0AlZAgtzCUKVKLuoAoBcEY3guv1dlNuGef7675JOsS587+/PL58zs71W63RT/iiI1GvryvnYJLDkoWCn52utzmUVmD0jh5vrgTps9UmIgD4Qm4zEKZMHi8AaUKA1iNjTgQxohWDAmrQOIFGEBL56G0AWkXLk1LpAW104R285FEHBoegkudtBwV5iHyBWvEifRKj0mXCHmuVOoxkUbkIVD10MRJ0zkRL/JBmv+vVMg9mgmtMejzmV5uk2ZIJCkMkeXqSYX+KkckixwtcloRnz4kack0W+O62l7/uCHWP2xIdcU7V8TgwLH97+WnL6Tfx06MiNvXLnK7OK7mNr2SLC0XGRsdEbXFx2L35+/9ut1ff4Q7M9X5vLT8TpRrL6Vnnj25ryUZ9HeOVLI6rY6PDovi1GWprrbwWmxuf+sMRo32LYj0hXNndLrIcjWunYfcu3tDnD09LtW5lYXOALZgAHtA+ZTon9CAE0iccmsjlB/IK/WrNQ8k9Eaqq8A9fu1bI04vAkbAiOOk9MOve5TH1ayZw1K+ZJ7WBkqmeB0m6/CBeoxy+WEhTNNOpMR/gJugo6jAyG9tf7XWj/2IP6pL8pAn62KYpluREUcHwcWoFz593jxgjVyoyVa6yw1GL6/o2eoig36tTla8BwdgkCmypOLptoye7ZfInl/jZFXtz0AyXiTE0a+xqO6ChLG4yoKDckJZaeA5d0MIb2S/dS8tv5WW8MGBo5JEMJnCJMy/giJ5VUYBezC8NyAw/O+UzCcND+ZdRscOq4ckH6/q+vhct9kcM1oQ7Xkt4pS4lxImXjDeEAKtrwh7223WNoY4K+dkApJBb50JlauQZC7FSB77mQzaP2flKtCIFxP5Tj+cLWd2kkXkMybpABOrRJrVvunGviv+nUTYkkYJCM/pPGR8lELvprM6qUEXwg0i3dJ9OBX21M13cHWTaZsNyoVCHWClbB8X+vb6HMp1WlR2uPpNhHhcSIs+Rd8S/yvAAKaXQVlKJQmeAAAAAElFTkSuQmCC" );
```
humphd: Why not let the browser cache this file?  With icons we load as separate files, why not here?

css/media-view.css
```diff
@@ -0,0 +1,55 @@
...
+.butter-media-properties.open .edit-message {
+  visibility: hidden;
+}
+.butter-media-properties .url {
+  width: 224px;
+  overflow: hidden;
+}
+.butter-media-properties .url input,
+.butter-media-properties .url button {
+  transition: opacity 0.2s;
```
humphd, line 26: Needs prefix versions.

src/dialog/dialog.js
```diff
@@ -0,0 +1,197 @@
...
+        rootElement: _rootElement,
+
+        activity: function( activityName ){
+          _activities[ activityName ]();
+        },
+
+        enableCloseButton: function(){
+          var closeButton = _rootElement.querySelector( ".close-button" );
+          if( closeButton ){
+            closeButton.addEventListener( "click", function( e ){
```
humphd: Remove this on fire?

closeButton.addEventListener( "click", function clickHandler( e ){
this.removeEventListener( "click", clickHandler, false);
...

src/dialog/dialog.js
```diff
@@ -0,0 +1,197 @@
...
+          _escapeKeyActivity = activityName;
+        },
+
+        registerActivity: function( name, callback ){
+          _activities[ name ] = callback;
+        },
+
+        assignButton: function( selector, activityName ){
+          var element = _rootElement.querySelector( selector );
+          element.addEventListener( "click", _activities[ activityName ], false );
```
humphd: Same deal here, won't we leak this?

src/dialog/dialog.js
humphd: Caching arguments.length is advised, yes.  Whether you wrote this code or not, you've touched it now, so we have to deal with it here.

src/dialog/dialog.js
humphd, line 264: This is exactly why I wanted async events in the EventMangerWrapper stuff I did.  By having an async event you solve this issue.  Firefox has gotten into a lot of problem using this hack, see http://blog.mozilla.org/tglek/2012/04/10/argh-at-our-unresponsive-tab-strip-settimeoutfoo-0-can-be-very-harmful/  I think we need to move to async event for this, and drop setTimeout.  That will need #1130 to land, so if you want to file a bug blocked on that, I can fix it later.

src/dialog/dialog.js
secretrobotron: I'm really not entirely sure. While I'll remove the listener for the sake of good practice, will this just get GC'd since references removed completely?

src/dialog/dialog.js
```diff
@@ -0,0 +1,197 @@
...
+
+      _internal.registerActivity( "default-ok", function(){
+        _external.dispatch( "submit" );
+        _external.close();
+      });
+
+      dialogCtor( _internal, spawnOptions.data );
+
+      // make this happen a tiny bit later so variables get initialized before opening
+      setTimeout( function(){
```
humphd: Same deal here, we want an async event instead.

src/dialog/dialogs/error-message.js
```diff
@@ -0,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the MIT license
+ * If a copy of the MIT license was not distributed with this file, you can
+ * obtain one at http://www.mozillapopcorn.org/butter-license.txt */
+
+define([ "text!dialog/dialogs/error-message.html", "dialog/dialog" ],
+  function( LAYOUT_SRC, Dialog ){
+
+  Dialog.register( "error-message", LAYOUT_SRC, function( dialog, data ) {
+    var message = dialog.rootElement.querySelector( ".message" );
+    message.innerHTML = data;
```
humphd, line 10: Kind of sucks that you still have to work with innerHTML here.  Should we provide some nicer way to not have to work directly with a dom node?

src/dialog/dialogs/export.js
```diff
@@ -0,0 +1,46 @@
...
+
+    var rootElement = dialog.rootElement;
+
+    var jsonButton = rootElement.querySelector( ".json-button" ),
+        htmlButton = rootElement.querySelector( ".html-button" ),
+        jsonExport = rootElement.querySelector( ".json-export" ),
+        htmlExport = rootElement.querySelector( ".html-export" ),
+        title = rootElement.querySelector( ".title" );
+
+    title.innerHTML = "HTML Export";
```
humphd, line 18: Same thing here...I'm wondering if we want to be working against a "dialog" object vs. a dialog-html-fragment?

src/dialog/dialogs/load.js
humphd: Have to agree with Dave, cache, cache, cache.

src/dialog/modal.js
```diff
@@ -4,36 +4,31 @@
...
     }, 10 );
 
+    _element.appendChild( childElement );
+
     this.destroy = function(){
       __container.removeChild( _element );
```
humphd, line 30: also _element = null;

src/dialogs.js
humphd: Even better:

var i = dialogs.length;
while( i-- ){
  ... dialogs[ i ]...
}

src/main.js
```diff
@@ -76,6 +80,8 @@
 
       EventManagerWrapper( _this );
 
+      this.dialog = Dialog;
```
humphd: Will you later just get this via require instead of holding an instance?

secretrobotron: It was my intent to leave it open for people to use `butter.dialog.register()` during execution instead of having to have your whole stack during compile time.

humphd: OK, that's awesome.  Maybe leave a comment to that effect above this line.

src/dialog/dialog.js
humphd: No, let's leave it.  Based on irc discussion with you/rick, I think it's fine.  My hesitation comes from historical issues.

src/dialogs.js
secretrobotron: Yeah, that's a lot nicer. I'm curious about why is this an issue to fix up here though? Should we always ensure that lengths are cached in loops during review? I like being speedy, but catching all of these seems more pedantic than necessary imo.

Happy to change it regardless.

src/dialog/dialogs/error-message.js
secretrobotron, line 10: I agree. However, I think that's another enhancement on top of this ticket. File a new one?

src/dialog/dialog.js
```diff
@@ -0,0 +1,197 @@
...
+      spawnOptions = spawnOptions || {};
+
+      var _listeners = spawnOptions.events || {},
+          _activities = {},
+          _rootElement = LangUtils.domFragment( layoutSrc ),
+          _enterKeyActivity,
+          _escapeKeyActivity,
+          _modal;
+
+      // make sure have a handle to the butter-dialog div
```
humphd: Typo: "make sure ^we have a handle..."

src/dialog/dialog.js
```diff
@@ -0,0 +1,197 @@
...
+          _activities[ _enterKeyActivity ]( e );
+        }
+        else if ( _escapeKeyActivity &&
+                  __keyboardAvoidElements.indexOf( e.target.nodeName ) === -1 &&
+                  ( e.which === 27 || e.keyCode === 27 ) ) {
+          _activities[ _escapeKeyActivity ]( e );
+        }
+      }
+
+      var _internal = {
```
humphd, line 82: Please add an explanatory comment to _internal and _external, and how they are used.  In general, this code needs a bit more documentation.

src/dialog/dialog.js
```diff
@@ -0,0 +1,197 @@
...
+            _rootElement.querySelector( arguments[ i ] ).setAttribute( "disabled", true );
+          }
+        },
+
+        send: function( message, data ){
+          _external.dispatch( message, data );
+        }
+      };
+
+      var _external = {
```
humphd, line 239: Same deal here.  Explain a bit.

jbuck:
Staged on master: https://github.com/mozilla/butter/compare/510b88cd8039...bd9990522eb2

PR #379, opened 2012-06-04
[#740] Added in logger.warn. Made it so errors and warnings always get t...

dseif:
...hrown


jbuck:
wont-fixed in Lighthouse

PR #380, opened 2012-06-04
T1278 Backend for Manual Test Results

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/1278-add-database-backend-for-manual-tests

### DIFF
--- a/cornfield/app.js
+++ b/cornfield/app.js
@@ -26,6 +26,7 @@ var mongoose = require('mongoose'),
     });
     Schema = mongoose.Schema,
 
+
     Project = new Schema({
       name: String,
       html: String,
@@ -45,7 +46,7 @@ var mongoose = require('mongoose'),
       testURL: String,
       result: Boolean,
       userAgent: String,
-      popVersion: String,
+      popcornVersion: String,
       butterVersion: String
     }),
     TestResultModel = mongoose.model( "TestResult", TestResult );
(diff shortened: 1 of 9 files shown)


cornfield/app.js
```diff
@@ -25,7 +25,8 @@ var mongoose = require('mongoose'),
       }
     });
     Schema = mongoose.Schema,
-    
+
+
```
humphd, line 29: Spurious white space change.

cornfield/app.js
```diff
@@ -269,6 +281,46 @@ app.get('/api/whoami', function( req, res ) {
...
+    testURL: req.body.testURL,
+    result: req.body.result,
+    userAgent: req.body.ua,
+    popcornVersion: req.body.popcornVersion,
+    butterVersion: req.body.butterVersion
+  });
+
+  result.save( function( err ) {
+    if ( !err ) {
+      res.json( { error: "okay" } );
```
humphd, line 302: error okay?  This seems odd.

test/manual-tests/manual-test.js
```diff
@@ -4,29 +4,70 @@
...
 
     function getInfo() {
       return {
-        ua: navigator.userAgent,
-        popcornVersion: Popcorn.version,
-        butterVersion: Butter.version,
         testTitle: document.title,
-        testURL: /[^\/]+$/.exec(location.pathname)[0]
+        butterVersion: Butter.version,
+        popcornVersion: Popcorn.version,
```
humphd, line 21: If these are the plain @VERSION@ strings, I wonder if we should use Date.now() too?  Something like:

butterVersion: Butter.version === "@VERSION@" ? Butter.version + " - " + Date.now() : Butter.version;

And same for Popcorn.version.  This way we at least get some historical sense of what you were looking at when you ran the test.

test/manual-tests/manual-test.js
```diff
@@ -4,29 +4,70 @@
...
+
+      xhr.open( "POST", server + "/api/tests/", false );
+
+      xhr.onreadystatechange = function() {
+        if ( this.readyState === 4 ) {
+          var response;
+          try {
+            response = JSON.parse( this.responseText );
+          } catch ( err ) {
+            console.error( "failed to parse data from server: \n" + err );
```
humphd, line 39: Should this be more than just in the console?  Maybe we want something in the page or an alert?

test/manual-tests/manual-test.js
```diff
@@ -4,29 +4,70 @@
...
+
     function passFn() {
-      parent.postMessage("PASS", "*");
+      result = true;
+      var data = JSON.stringify( getInfo() );
+
+      postResults( data );
+
+      parent.postMessage( "PASS", "*" );
+      disableButtons();
```
humphd, line 55: I think this is wrong.  You might go on irc, have a discussion, and change your mind.  It would be better to maybe highlight the one they clicked with a color, or have a label beside that shows their choice.  I think you want to be able to change it.

test/manual-tests/manual-test.js
```diff
@@ -4,29 +4,70 @@
...
 
     function failFn() {
-      parent.postMessage("FAIL", "*");
+      result = false;
+      var data = JSON.stringify( getInfo() );
+
+      postResults( data );
+
+      parent.postMessage( "FAIL", "*" );
+      disableButtons();
```
humphd, line 65: Same deal here.

test/manual-tests/results.html
```diff
@@ -0,0 +1,74 @@
...
+                data = JSON.parse( this.responseText ),
+                result;
+
+            data = data.testResults;
+            data.reverse();
+
+            for ( x in data ) {
+              result = data [ x ];
+
+              tr = document.createElement( "tr" );
```
humphd, line 28: Let's use something like http://tablesorter.com/docs/ to give us a usable table (e.g., sorting) without wasting cycles on UI here.  Something we can use with external JS vs. pulling it into our tree.

mjschranz:
Ticket was closed in Lighthouse

PR #381, opened 2012-06-04
[#565] When going from a flash player -> html5 media, we discard the wra...

dseif:
...pper div correctly now


src/core/popcorn-wrapper.js
```diff
@@ -169,7 +190,18 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
               _onConstructing();
             }
             // wait for the media to become available and notify the user, or timeout
-            waitForMedia( _onPrepare, timeoutWrapper );
+            waitForMedia(function() {
```
mjschranz, line 193: Whiiiiiiiiiitespace

dseif, line 193: Wrong

jbuck:
wont-fixed in Lighthouse

PR #382, opened 2012-06-04
[#1343] Trackevents use plugin defaults as soon as first update is calle...

dseif:
...d, which is right after they are created

### DIFF
--- a/src/core/trackevent.js
+++ b/src/core/trackevent.js
@@ -53,15 +53,11 @@ define( [
     _this.popcornOptions = _popcornOptions;
 
     function defaultValue( item ) {
-      var val;
-
       if ( item.default ) {
-        val = item.default;
+        return item.default;
       } else {
-        val = item.type === "number" ? 0 : "";
+        return item.type === "number" ? 0 : "";
       }
-
-      return val;
     }
 
     if( !_type ){
--- a/src/core/trackevent.js
+++ b/src/core/trackevent.js
@@ -52,6 +52,18 @@ define( [
 
     _this.popcornOptions = _popcornOptions;
 
+    function defaultValue( item, val ) {
+      // Don't print "undefined" or the like
+      if ( val === undefined || typeof val === "object" ) {
+        if ( item.default ) {
+          val = item.default;
+        } else {
+          val = item.type === "number" ? 0 : "";
+        }
+      }
+      return val;
+    }
+
     if( !_type ){
       _logger.log( "Warning: " + _id + " has no type." );
     } //if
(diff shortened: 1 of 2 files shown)


src/core/trackevent.js
```diff
@@ -52,6 +52,18 @@ define( [
 
     _this.popcornOptions = _popcornOptions;
 
+    function defaultValue( item, val ) {
```
humphd: Code duplication from editors/default-editor.js.  If you're going to do this in order for us to land on 0.5.x, you need to file a bug in 0.6 to refactor so we don't have duplicate code in the tree.

src/core/trackevent.js
```diff
@@ -98,13 +110,17 @@ define( [
...
-          if ( updateOptions.hasOwnProperty( prop ) ) {
-            _popcornOptions[ prop ] = updateOptions[ prop ];
-          } //if
-        } //for
+      } else {
+        var _manifest = Popcorn.manifest[ _type ].options;
+        for ( var prop in _manifest ) {
+          if ( _manifest.hasOwnProperty( prop ) ) {
+            if ( updateOptions[ prop ] === undefined ) {
+              _popcornOptions[ prop ] = defaultValue( _manifest[ prop ] );
```
humphd, line 113: All callers of this version of defaultValue() pass only a single argument, which means val === undefined is always true.  Simplify your defaultValue to get rid of the other code path.

ghost:
#### From: Bot.io (butyr.org)

---

# Received

Command **cmd_preview** from @humphd received. Current queue size: 0

Live output at: http://botio.butyr.org:8000/f71419d0885b99b/output.txt

ghost:
#### From: Bot.io (butyr.org)

---

# Success

Full output at http://botio.butyr.org:8000/f71419d0885b99b/output.txt

Total script time: 0.74 mins

Server is running on http://pr382.butyr.org/templates/test.html

src/core/trackevent.js
dseif: Is there anywhere you would suggest putting a function like this so it's accesible to both the default editor and the trackEvent class?

humphd: For 0.5.x there isn't, no.  For 0.6 we probably want to see what Bobby does with editors, since there are tickets to allow editors to work with Butter objects, methods, etc.

dseif: Alright I'll file a follow up ticket in 0.6 then, thanks.

test/core.js
```diff
@@ -348,6 +348,37 @@
...
+          end: 6
+        });
+      });
+
+      butter.listen( "trackeventupdated", function( e ) {
+        equal( e.data.popcornOptions.text, "Popcorn.js", "Trackevent is given proper plugin defaults" );
+        start();
+      });
+
+      m = butter.addMedia({ url: "../external/popcorn-js/test/italia.ogg", target: "mediaDiv" }),
```
mjschranz, line 370: Needs some spacing fixed between the () brackets.

dseif, line 370: we don't need to do that for object literals, see https://webmademovies.lighthouseapp.com/projects/63272/styleguide ctrl+f for  Paren whitespace exceptions

mjschranz, line 370: _shrug_

It's that way in the rest of the file.

PR #386, opened 2012-06-05
[#1316] Template Configs can now specify a project ID and have butter lo...

mjschranz:
...ad it from cornfield instead of only project JSON

https://webmademovies.lighthouseapp.com/projects/65733/tickets/1316-unable-to-specify-the-users-project-id-when-it-is-loading

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -570,7 +570,7 @@
           if( callback ){
             callback();
           } //if
-          
+
           _this.dispatch( "pageready" );
         });
       }; //preparePage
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -675,7 +675,17 @@
             catch( e ){
               _this.dispatch( "loaddataerror", "Saved data not formatted properly." );
             }
-            _this.importProject( savedData );
+
+            if ( !savedData.projectID ) {
+              _this.importProject( savedData );
+            }
+            else {
+              _this.cornfield.load( savedData.projectID, function( e ) {
```
secretrobotron: This is interesting, but might not be necessary since `savedData` is already a JSON blob with some project data in it. As a result, `load` would be redundant unless the data on the server changed.

It's assumed that the data loaded here is actually _from_ cornfield already.

PR #392, opened 2012-06-05
[#1285] Made sure frameAnimation is not turned on when we export. Added ...

dseif:
...a todo flag to update this to be optional later

### DIFF
--- a/src/core/media.js
+++ b/src/core/media.js
@@ -262,9 +262,10 @@
         }
 
         /* TODO: determine if we need to turn on frameAnimation or not before calling generatePopcornString
-         * for now we default to off when exporting by passing an empty options object
+         * for now we default to off when exporting by setting frameAnimation to false
          */
-        return _popcornWrapper.generatePopcornString( {}, _url, _target, null, callbacks, scripts, collectedEvents );
+        _popcornOptions.frameAnimation = false;
+        return _popcornWrapper.generatePopcornString( _popcornOptions, _url, _target, null, callbacks, scripts, collectedEvents );
       };
 
       Object.defineProperties( this, {
--- a/src/core/media.js
+++ b/src/core/media.js
@@ -261,7 +261,10 @@
           collectedEvents = collectedEvents.concat( _tracks[ i ].trackEvents );
         }
 
-        return _popcornWrapper.generatePopcornString( _popcornOptions, _url, _target, null, callbacks, scripts, collectedEvents );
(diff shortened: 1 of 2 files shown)


src/core/media.js
```diff
@@ -261,7 +261,10 @@
           collectedEvents = collectedEvents.concat( _tracks[ i ].trackEvents );
         }
 
-        return _popcornWrapper.generatePopcornString( _popcornOptions, _url, _target, null, callbacks, scripts, collectedEvents );
+        /* TODO: determine if we need to turn on frameAnimation or not before calling generatePopcornString
+         * for now we default to off when exporting by passing an empty options object
+         */
+        return _popcornWrapper.generatePopcornString( {}, _url, _target, null, callbacks, scripts, collectedEvents );
```
humphd: This worries me, since it loses anything else that might have been on _popcornOptions.  Why not just explicitly set rAF on that options object to false?

dseif: I don't think we set anything else on the options object but you're right in that I probably shouldn't ignore the entire object. I'll update now.

src/core/media.js
```diff
@@ -261,6 +261,10 @@
           collectedEvents = collectedEvents.concat( _tracks[ i ].trackEvents );
         }
 
+        /* TODO: determine if we need to turn on frameAnimation or not before calling generatePopcornString
```
humphd, line 264: File a bug on this TODO so we don't lose it.  Put the bug # in the comment.

src/core/popcorn-wrapper.js
```diff
@@ -283,7 +283,7 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
       // special case for basePlayer, since it doesn't require as much of a harness
       if( _mediaType === "baseplayer" ) {
         popcornString +=  "Popcorn.player( 'baseplayer' );\n" +
-                          "var popcorn = Popcorn.baseplayer( '#" + target + "', " + popcornOptions + " );\n";
+                          "var popcorn = Popcorn.baseplayer( '#" + target + "' " + popcornOptions + " );\n";
```
cadecairos, line 286: Why did you remove that comma?

dseif, line 286: because it shouldn't be there. Popcorn options inserts it one for us, so having two would be wrong. I just saw it and fixed it, as it's going to have to get fixed sooner or later for the baseplayer to work.

PR #401, opened 2012-06-06
T1317 Export HTML is not taking into consideration the domain

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/1317-export-html-is-not-taking-into-consideration-the-domain

### DIFF
--- a/cornfield/app.js
+++ b/cornfield/app.js
@@ -25,7 +25,7 @@ var mongoose = require('mongoose'),
       }
     });
     Schema = mongoose.Schema,
-    
+
     Project = new Schema({
       name: String,
       html: String,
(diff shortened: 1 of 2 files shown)


cornfield/app.js
```diff
@@ -88,19 +88,6 @@ function publishRoute( req, res ){
             url = PUBLISH_PREFIX + "/" + id + ".html",
             data = doc.projects[ i ].html;
 
-        var template = TEMPLATES[ doc.projects[ i ].template ];
```
secretrobotron, line 91: Probably don't need `TEMPLATES` at all anymore, right?

PR #408, opened 2012-06-06
added check-templates to lint CSS and JS of templates dir, removed templ...

k88hudson:
...ates dir from check

### DIFF
--- a/make.js
+++ b/make.js
@@ -14,6 +14,62 @@ var JSLINT = './node_modules/jshint/bin/hint',
 
 require('shelljs/make');
 
+
+function checkCSS( dirs, warnings, errors ) {
+  // dirs must be passed as an array
+  echo('### Linting CSS files');
+  if (dirs instanceof Array) {
+    dirs = dirs.join(' ');
+  }
+
+  // see cli.js --list-rules.
+  warnings = warnings || [
+//    "important",
+//    "adjoining-classes",
+//    "duplicate-background-images",
+//    "qualified-headings",
+    "fallback-colors",
+//    "empty-rules",
+//    "shorthand",
+//    "overqualified-elements",
+//    "import",
+    "regex-selectors",
+//    "rules-count",
+//    "font-sizes",
+//    "universal-selector",
+//    "unqualified-attributes",
+    "zero-units"
+  ].join(",");
+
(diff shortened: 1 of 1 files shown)


make.js
```diff
@@ -14,6 +14,62 @@ var JSLINT = './node_modules/jshint/bin/hint',
 
 require('shelljs/make');
 
+
+function checkCSS( dirs, warnings, errors ) {
+  // dirs must be passed as an array
```
jbuck: No they don't

make.js
```diff
@@ -14,6 +14,62 @@ var JSLINT = './node_modules/jshint/bin/hint',
...
+  echo('### Linting CSS files');
+  if (dirs instanceof Array) {
```
jbuck, line 20: find() can handle arrays just fine, just delete this block

make.js
```diff
@@ -14,6 +14,62 @@ var JSLINT = './node_modules/jshint/bin/hint',
...
+    dirs = dirs.join(' ');
+  }
+
+  // see cli.js --list-rules.
+  warnings = warnings || [
```
jbuck: Personally, I'd prefer to have a single set of warnings that we follow, rather than being able to override it somehow. so this would just be `var warnings = [`, and the checkCSS function signature above would change.

jbuck: Unless there's somewhere you're overriding this... but I don't think you are

make.js
```diff
@@ -14,6 +14,62 @@ var JSLINT = './node_modules/jshint/bin/hint',
...
+//    "import",
+    "regex-selectors",
+//    "rules-count",
+//    "font-sizes",
+//    "universal-selector",
+//    "unqualified-attributes",
+    "zero-units"
+  ].join(",");
+
+  errors = errors && errors.join(",") || [
```
jbuck: Same as above for warnings

make.js
k88hudson: Not doing anything with this, but I figured it might be useful. I can take it out

make.js
k88hudson, line 20: CSSlint doesn't use find though, it just takes a string

make.js
k88hudson: Heh, true.

PR #414, opened 2012-06-07
T1369 Fixed track data not populating editors upon opening an existing event

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/1369-pop-up-template-after-editing-pop-plugin-text-dissapears#ticket-1369-9

### DIFF
--- a/templates/supported/popup/editors/js/editor-state.js
+++ b/templates/supported/popup/editors/js/editor-state.js
@@ -373,11 +373,17 @@
 		fieldset = field.fieldset || select;
 
 		if (targets.length <= 1) {
+			if ( fieldset ) {
+				fieldset.style.display = 'none';
+			}
+
 			if (targets.length) {
 				this.trackEvent.target = targets[0];
 			}
 		} else {
-			fieldset.style.display = '';
+			if ( fieldset ) {
+				fieldset.style.display = '';
+			}
 			select = field.element;
 
 			for (i = 0; i < targets.length; i++) {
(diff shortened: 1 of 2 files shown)


ghost:
#### From: Bot.io (butyr.org)

---

# Received

Command **cmd_preview** from @secretrobotron received. Current queue size: 0

Live output at: http://botio.butyr.org:8000/c498fed1f5a61a9/output.txt

ghost:
#### From: Bot.io (butyr.org)

---

# Failed

Full output at http://botio.butyr.org:8000/c498fed1f5a61a9/output.txt

Total script time: 0.20 mins

templates/supported/popup/editors/js/editor-state.js
```diff
@@ -368,23 +368,20 @@
 		if (typeof field.filter === 'function') {
 			field.filter(field, targets);
 		}
-		
+
 		select = field.element;
 		fieldset = field.fieldset || select;
-		
-		if (targets.length <= 1) {
-			fieldset.style.display = 'none';
```
secretrobotron, line 376: Ahhhh I see now! Good eye!

However, it's probably more correct to check for the existence of `fieldset` before attempting to make its display none instead of removing this line. Notice that the `else` block here tries to use `fieldset` as well, which should also be guarded.

The point is: we should only try to manipulate aspects of the element that this code has tried to automatically grab if it actually found something :P.

PR #420, opened 2012-06-07
T1339

k88hudson:
Draggable hooks


templates/supported/newscaster/butter/template.js
```diff
@@ -23,10 +24,114 @@ document.addEventListener( "DOMContentLoaded", function( e ){
           }
         });
 
+        // EDITING HOOKS
+        // Would be good if these were part of a library in shared, instead of existing here.
+        // Should be updated to a hook syntax like hook("image2", myUpdateFunction) or something
+        // Add listeners for future track events.
+          butter.listen("trackeventadded", updateFunction);
```
secretrobotron: Can you fix this indentation?

templates/supported/newscaster/butter/template.js
```diff
@@ -6,7 +6,8 @@ document.addEventListener( "DOMContentLoaded", function( e ){
   Butter({
     config: "butter/config.json",
     ready: function( butter ){
-      var media = butter.media[ 0 ];
+      var media = butter.media[ 0 ],
+          popcorn = butter.media[ 0 ].popcorn.popcorn;
```
secretrobotron: This won't cause trouble, will it? I'm guessing you only need popcorn for this closure, and not when export happens?

templates/supported/newscaster/butter/template.js
```diff
@@ -23,10 +24,114 @@ document.addEventListener( "DOMContentLoaded", function( e ){
...
+          butter.listen("trackeventadded", updateFunction);
+          butter.listen("trackeventupdated", updateFunction);
+
+        function updateFunction(e) {
+
+          var trackEvent,
+              _container = null,
+              _textEls,
+              _popcornOptions,
+              canvas = document.createElement( "canvas" ),
```
secretrobotron: Why the disparity between _ and no-_? Can you choose one and stick with it here?

templates/supported/newscaster/butter/template.js
```diff
@@ -23,10 +24,114 @@ document.addEventListener( "DOMContentLoaded", function( e ){
...
+            trackEvent = e.target; 
+          } else { 
+            trackEvent = e; 
+          }
+
+          _popcornOptions = trackEvent.popcornOptions;
+
+          // Requires _container in options to work.
+          // This will be better when track events store a reference to popcorn in Butter, but it's ok for now.
+          trackEvent.popcornTrackEvent = popcorn.getTrackEvent( popcorn.getLastTrackEventId() ); //Store a reference
```
secretrobotron: Seriously? Please file a new bug for this if there isn't one already. 0.5.2. This is ridiculous.

templates/supported/newscaster/butter/template.js
k88hudson: No, this is only inside butter. We will eventually have to deal with popcorn code, but for the purposes of this editing code, popcorn is always within the closure.

templates/supported/newscaster/butter/template.js
k88hudson: There is a bug filed, but its in .6 i think

templates/supported/newscaster/butter/template.js
```diff
@@ -1,33 +1,142 @@
...
+            }
+
+            if ( trackEvent.type === "image2" ) {
+              // Prevent default draggable behaviour of images
+              trackEvent.popcornTrackEvent._image.addEventListener( "mousedown", function(e) {
+                e.preventDefault();
+              }, false);
+
+              //Change default text to indicate draggable
+              if( !_popcornOptions.src && window.$ ){
```
humphd, line 64: you pass window.$ in as $ in the anonymous closure already.

templates/supported/newscaster/butter/template.js
```diff
@@ -1,33 +1,142 @@
...
+                _container.innerHTML = "Drag an image from your desktop";
+              }
+
+              //Apply resizable/draggable if jQuery exists
+
+              if( $ ) {
```
humphd, line 70: This can fail if $ is really null.  Why would it be?  Don't you include jquery in the page?

PR #422, opened 2012-06-07
Disabled Stamen-Terrain map type [#1404]

cadecairos:


jbuck:
This was pulled in on v0.5.x.

PR #423, opened 2012-06-07
added speak plugin

k88hudson:


templates/supported/shared/plugins/speak/popcorn.speak.js
```diff
@@ -0,0 +1,233 @@
...
+            elem: "input",
+            type: "number",
+            label: "Pitch:"
+          },
+          speed: {
+            elem: "input",
+            type: "number",
+            label: "Speed:"
+          },
+          pluginPath: {
```
k88hudson, line 59: Is there a better way to figure out the relative path of the worker?

secretrobotron, line 59: What wrong with this, actually? Are you worried that people won't be able to figure out how to specify where the correct file is?

templates/supported/shared/plugins/speak/popcorn.speak.js
```diff
@@ -0,0 +1,233 @@
...
+              handleWav(event.data);
+            };
+            speakWorker.postMessage({ text: text, args: args });
+          }
+        }
+
+      // END SPEAK.JS---------------------------------------
+
+      speakOptions = {};
+      options.amplitude && ( speakOptions.amplitude = options.amplitude );
```
secretrobotron, line 197: Can these not just be `speakOptions.amplitude = options.amplitude`? If `options.amplitude` isn't set, `speakOptions.amplitude` will also just remain unset, right? It'll be `undefined` regardless? This will disregard 0's as well. Is that intended?

PR #443, opened 2012-06-09
T1433 Renamed Image2 Plugin to Image

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/1433-rename-image2-plugin-to-something-more-friendly


templates/supported/newscaster/butter/template.js
```diff
@@ -54,7 +54,7 @@ document.addEventListener( "DOMContentLoaded", function( e ){
               return;
             }
 
-            if ( trackEvent.type === "image2" ) {
+            if ( trackEvent.type === "image" ) {
```
k88hudson: The only thing I'm worried about is that this editing hook won't work on the original image plugin (which doesn't store position data), and if someone tries to apply to more than one template, it will fail. 

Anyone who looks at this code who isn't me/us might be confused by two different plugins having the same name. Could this possible be an issue?

PR #465, opened 2012-06-12
T1477

humphd:
...ter'


templates/supported/popup/config.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "pop",
+  "title": "Popup"
```
cadecairos: butter fails to load because of a missing comma here

PR #467, opened 2012-06-12
Storycamp

humphd:


make.js
```diff
@@ -276,3 +285,46 @@ target.test = function() {
...
+target.storycamp = function(){
+  echo('### Making single file version of Butter + Popcorn');
+
+  build( 'storycamp' );
+  target['buttered-popcorn']();
+
+  var storyCamp = DIST_DIR + '/storycamp-butter.js',
+      storyCampMin = DIST_DIR + '/storycamp-butter.min.js';
+
+  cat( './LICENSE_HEADER', BUTTERED_POPCORN, DIST_DIR + '/butter.js' ).to( storyCamp );
```
jbuck: Didn't you fix this in Lighthouse #1391?

humphd: It's slightly different in this case with combining them.  Only one header block can be ! important, it seems.  This is a hack to get them both to show, since it's popcorn + butter.  It's good enough for now, I think.

make.js
```diff
@@ -276,3 +285,46 @@ target.test = function() {
...
+
+  // Write out dist/storycamp-butter.js
+  exec( UGLIFY + ' --output ' + storyCampMin + ' ' + storyCamp );
+
+  echo('### Cleaning temp files');
+  var cwd = pwd();
+  cd(DIST_DIR);
+  rm('-f', 'butter.js', 'storycamp-butter.js', 'butter.min.js', 'buttered-popcorn.js', 'buttered-popcorn.min.js');
```
jbuck: Can we have a flag (perhaps set by an environment variable) that determines whether to use minified code, or unminified code for everything? Maybe a `minString` variable that's `''` or `'.min'` and you just append it to all file names.

It sounds more complicated, but if we're using this for storymaker-dev, then I want it to be as easy to debug as possible

make.js
```diff
@@ -276,3 +285,46 @@ target.test = function() {
...
+  mkdir('css');
+  mv('*.css', './css');
+
+  // Copy other assets over
+  cd(cwd);
+  cp('-R', 'resources', DIST_DIR);
+  cp('-R', 'dialogs', DIST_DIR);
+  cp('-R', 'editors', DIST_DIR);
+  cp('-R', 'templates', DIST_DIR);
+
```
jbuck, line 334: Can you add `cp('-R', 'cornfield', DIST_DIR);` as well?

PR #479, opened 2012-06-12
T1491

secretrobotron:


jbuck:
Add loading of project data from URL [#1491]

PR #487, opened 2012-06-13
T1511

humphd:


jbuck:
I think ben just removed the low value templates from the listing in the storycamp repo.

PR #497, opened 2012-06-14
[#1527] Fixed cornfield exporting issues

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/1527-make-dirs-section-in-cornfield-config-all-relative

### DIFF
--- a/cornfield/app.js
+++ b/cornfield/app.js
@@ -9,7 +9,7 @@ const express = require('express'),
       CONFIG = require('config'),
       TEMPLATES_DIR =  CONFIG.dirs.templates,
       PUBLISH_DIR = CONFIG.dirs.publish,
-      PUBLISH_PREFIX = CONFIG.dirs.publishPrefix,
+      PUBLISH_PREFIX = CONFIG.dirs.hostname,
       WWW_ROOT = path.resolve( CONFIG.dirs.wwwRoot || path.join( __dirname, ".." ) ),
       VALID_TEMPLATES = CONFIG.templates,
       EXPORT_ASSETS = CONFIG.exportAssets;
@@ -111,19 +111,33 @@ function publishRoute( req, res ){
               currentTrack,
               currentTrackEvent,
               mediaPopcornOptions,
+              templateBase = VALID_TEMPLATES[ template ],
+              templateURL,
+              baseString,
+              headStartTagIndex,
+              templateScripts,
+              startString,
               j, k;
-
+          
+          templateURL = templateBase.substring( templateBase.indexOf( '/templates' ), templateBase.lastIndexOf( '/' ) );
+          baseString = '\n  <base href="' + PUBLISH_PREFIX + templateURL + '/"/>';
+          
           // look for script tags with data-butter-exclude in particular (e.g. butter's js script)
           data = data.replace( /\s*<script[\.\/='":_-\w\s]*data-butter-exclude[\.\/='":_-\w\s]*><\/script>/g, '' );
-
+          
+          // Adding 6 to cut out the actual head tag
+          headStartTagIndex = data.indexOf( '<head>' ) + 6;
           headEndTagIndex = data.indexOf( '</head>' );
           bodyEndTagIndex = data.indexOf( '</body>' );
+          
+          templateScripts = data.substring( headStartTagIndex, headEndTagIndex );
+          startString = data.substring( 0, headStartTagIndex);
 
           for ( i = 0; i < EXPORT_ASSETS.length; ++i ) {
-            externalAssetsString += '\n<script src="' + EXPORT_ASSETS[ i ] + '"></script>';
+            externalAssetsString += '\n<script src="' + PUBLISH_PREFIX + '/' + EXPORT_ASSETS[ i ] + '"></script>';
           }
 
-          popcornString += '<script>'
+          popcornString += '<script>';
 
           for ( i = 0; i < projectData.media.length; ++i ) {
             currentMedia = projectData.media[ i ];
@@ -143,7 +157,7 @@ function publishRoute( req, res ){
           }
           popcornString += '</script>';
 
-          data = data.substring( 0, headEndTagIndex ) + externalAssetsString + data.substring( headEndTagIndex, bodyEndTagIndex ) + popcornString + data.substring( bodyEndTagIndex );
+          data = startString + baseString + templateScripts + externalAssetsString + data.substring( headEndTagIndex, bodyEndTagIndex ) + popcornString + data.substring( bodyEndTagIndex );
 
           fs.writeFile( projectPath, data, function(){
             if( err ){
--- a/cornfield/app.js
+++ b/cornfield/app.js
@@ -32,7 +32,8 @@ var mongoose = require('mongoose'),
       name: String,
       html: String,
       data: String,
-      template: String
+      template: String,
+      popcornString: String
     }),
     ProjectModel = mongoose.model( 'Project', Project ),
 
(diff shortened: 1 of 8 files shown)


templates/supported/newscaster/butter/config.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "newscaster",
   "title": "Newscaster",
+  "template": "../templates/supported/newscaster/index.html",
```
dseif: should we use baseDir in the path as we do in the editors?

cornfield/config/default.json
```diff
@@ -14,15 +14,15 @@
...
   },
   "templates": {
-    "test": "../templates/test.html",
-    "robots": "../templates/supported/robots/index.html",
-    "popup": "../templates/supported/popup/index.html",
-    "newscaster": "../templates/supported/newscaster/index.html"
+    "test": "../templates/test-config.json",
+    "robots": "../templates/supported/robots/butter/config.json",
+    "popup": "../templates/supported/popup/config.json",
+    "newscaster": "../templates/supported/newscaster/config.json"
```
dseif: can we use templates like we do baseDir, so {{templates}} since we alread write it out above

cornfield/app.js
```diff
@@ -102,57 +102,81 @@ function publishRoute( req, res ){
...
+                currentTrack = currentMedia.tracks[ j ];
+                for ( k = 0; k < currentTrack.trackEvents.length; ++k ) {
+                  currentTrackEvent = currentTrack.trackEvents[ k ];
+                  popcornString += '\npopcorn.' + currentTrackEvent.type + '(';
+                  popcornString += JSON.stringify( currentTrackEvent.popcornOptions, null, 2 );
+                  popcornString += ');';
+                }
               }
+              popcornString += '}());';
             }
```
dseif: Don't we already do all of this work? Is there some way we can tack it on to the projectData object so we don't have to repeat this step?

cornfield/app.js
```diff
@@ -102,57 +102,81 @@ function publishRoute( req, res ){
...
+                j, k;
+
+            templateURL = templateBase.substring( templateBase.indexOf( '/templates' ), templateBase.lastIndexOf( '/' ) );
+            baseString = '\n  <base href="' + PUBLISH_PREFIX + templateURL + '/"/>';
+
+            // look for script tags with data-butter-exclude in particular (e.g. butter's js script)
+            data = data.replace( /\s*<script[\.\/='":_-\w\s]*data-butter-exclude[\.\/='":_-\w\s]*><\/script>/g, '' );
+            
+            // Adding 6 to cut out the actual head tag
+            headStartTagIndex = data.indexOf( '<head>' ) + 6;
```
dseif: 6?

cornfield/config/default.json
```diff
@@ -14,15 +14,15 @@
     "wwwRoot": "../",
     "templates": "../templates",
     "publish": "view",
-    "publishPrefix": "http://localhost:8888/cornfield/view"
+    "hostname": "http://localhost:8888"
```
dseif, line 17: will this be fine once it gets hosted somewhere?

templates/supported/newscaster/butter/config.json
dseif: nvm I see this is in an interal butter folder or something

cornfield/app.js
mjschranz: 6 is the length of the <head> tag bracket.

cornfield/config/default.json
mjschranz, line 17: It's something that has to be manually changed to the correct value to wherever it's hosted. AFAIK this is how it was before.

templates/supported/newscaster/butter/config.json
mjschranz: I don't think I can use the same baseDir that's in there for everything else. I tried a little and unless I'm using path.resolve wrong with node it's basically because they are from two different places. The baseDir stuff that's there is relative to butter, where as I am relative to cornfield.

cornfield/config/default.json
mjschranz: probably could do something like that.

cornfield/app.js
secretrobotron: Yes, but we need it here now. It was inserted in a different ticket, so I don't actually know why it's all green unless it went through an indentation change.

cornfield/config/default.json
secretrobotron: That would be an awesome idea.

templates/supported/robots/butter/config.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "robots",
   "title": "Robots",
+  "template": "../templates/supported/robots/index.html",
```
secretrobotron: These should probably be relative, or use {{baseDir}} if they can, no?

cornfield/app.js
```diff
@@ -102,57 +102,81 @@ function publishRoute( req, res ){
...
+            headEndTagIndex = data.indexOf( '</head>' );
+            bodyEndTagIndex = data.indexOf( '</body>' );
+            
+            templateScripts = data.substring( headStartTagIndex, headEndTagIndex );
+            startString = data.substring( 0, headStartTagIndex);
```
secretrobotron: Add some spacing here before the last bracket.

cornfield/app.js
```diff
@@ -102,57 +102,81 @@ function publishRoute( req, res ){
...
-            popcornString += '\n(function(){';
-            popcornString += '\nvar popcorn = Popcorn.smart("#' + currentMedia.target + '", "' + currentMedia.url + '", ' + JSON.stringify( mediaPopcornOptions ) + ');';
-            for ( j = 0; j < currentMedia.tracks.length; ++ j ) {
-              currentTrack = currentMedia.tracks[ j ];
-              for ( k = 0; k < currentTrack.trackEvents.length; ++k ) {
-                currentTrackEvent = currentTrack.trackEvents[ k ];
-                popcornString += '\npopcorn.' + currentTrackEvent.type + '(';
-                popcornString += JSON.stringify( currentTrackEvent.popcornOptions, null, 2 );
-                popcornString += ');';
+        fs.readFile( VALID_TEMPLATES[ template ], function(err, conf){
```
secretrobotron: Looks like the `utf8` argument was dropped here. Any reason?

jbuck: Yeah, that'll need to be added back.

cornfield/app.js
jbuck: It did go through an indentation change in comment #r985775 above

cornfield/config/default.json
jbuck, line 17: Yeah, it can be override in the runtime.json file.

templates/supported/robots/butter/config.json
jbuck: Uhhh, this should just be "../index.html"

templates/supported/newscaster/butter/config.json
jbuck: ../index.html

templates/test-config.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "test",
+  "template": "../templates/test.html",
```
jbuck: test.html

templates/supported/popup/config.json
```diff
@@ -1,6 +1,7 @@
 {
-  "name": "pop",
+  "name": "popup",
   "title": "Popup",
+  "template": "../templates/supported/popup/index.html",
```
jbuck: index.html

cornfield/app.js
```diff
@@ -100,59 +101,67 @@ function publishRoute( req, res ){
...
+            
+            templateScripts = data.substring( headStartTagIndex, headEndTagIndex );
+            startString = data.substring( 0, headStartTagIndex);
+
+            for ( i = 0; i < EXPORT_ASSETS.length; ++i ) {
+              externalAssetsString += '\n<script src="' + PUBLISH_PREFIX + '/' + EXPORT_ASSETS[ i ] + '"></script>';
+            }
+            
+            // If the template has custom plugins defined in it's config, add them to our exported page
+            if ( templateConfig.plugin.plugins ) {
```
jbuck: `if ( templateConfig.plugin.plugins ) {` => `if ( templateConfig.plugin && templateConfig.plugin.plugins ) {`

src/ui/header.js
```diff
@@ -146,6 +146,7 @@ define( [
       function execute(){
         butter.project.html = butter.getHTML();
         butter.project.data = butter.exportProject();
+        butter.project.popcornString = butter.getPopcornString();
```
secretrobotron: We removed this stuff on purpose so that people can't submit their own popcornString and circumvent our safety measures here. Can you put the old server-side Popcorn string generation back in?

cornfield/app.js
```diff
@@ -100,59 +101,68 @@ function publishRoute( req, res ){
...
+                popcornString = '',
+                templateURL,
+                baseString,
+                headStartTagIndex,
+                templateScripts,
+                startString,
+                j, k;
+
+            templateURL = templateFile.substring( templateFile.indexOf( '/templates' ), templateFile.lastIndexOf( '/' ) );
+            baseString = '\n  <base href="' + PUBLISH_PREFIX + templateURL + '/"/>';
```
secretrobotron, line 127: You don't really need to escape that last quote, right?

mjschranz, line 127: no escaping of quotes here!

PR #549, opened 2012-06-18
Don't show directory listings in production environment

jbuck:


cornfield/app.js
```diff
@@ -49,16 +49,21 @@ if ( !path.existsSync( PUBLISH_DIR ) ) {
...
-  }))
-  .use(express.static( WWW_ROOT ))
-  .use(express.static( PUBLISH_DIR ))
-  .use(express.directory( WWW_ROOT, { icons: true } ) );
+app.configure( function() {
+  app.use( express.logger( CONFIG.logger ) )
+    .use( express.static( WWW_ROOT ))
+    .use( express.static( PUBLISH_DIR ))
+    .use( express.bodyParser())
+    .use( express.cookieParser())
```
mjschranz: If we are going to have whitespace on the left here we should in the right as well for all four of these, should we not?

ScottDowne: Just the whitespace, and this is good.

PR #580, opened 2012-06-19
T1644

humphd:


templates/supported/popup/plugins/popcorn.pop.js
```diff
@@ -16,6 +16,19 @@ todo: animate top, left and other styles (color, font size, etc.)
...
+  /**
+   * Stub implementation of words for API compat in 0.5.x
+   */
+  function nop(){}
+  Popcorn.plugin( 'words', {
+    _setup:    nop,
+    start:     nop,
+    frame:     nop,
+    end:       nop,
+    _teardown: nop
```
ScottDowne: Popcorn.plugin( "words" ) is safe, you don't need the nops.

PR #592, opened 2012-06-19
T1656

ScottDowne:
...be sure.


templates/supported/popup/plugins/popcorn.pop.js
```diff
@@ -19,7 +19,12 @@ todo: animate top, left and other styles (color, font size, etc.)
   /**
    * Stub implementation of words for API compat in 0.5.x
    */
-  Popcorn.plugin( 'words', {} );
+  Popcorn.plugin( 'words', {
+    _setup: Popcorn.nop,
+    start: Popcorn.nop,
+    end: Popcorn.nop,
```
humphd, line 25: Add frame in here too, since Brian had it before.

PR #643, opened 2012-07-10
[t1775] init/before/after scripts in export

secretrobotron:
1. Added support for using functions directly (only works client-side).
2. Added support for #id script insertion in cornfield.
3. Added tests for (1).
4. Exposed scripts/callbacks on Butter objects.


jbuck:
wont-fixing for now, while Scott tries the monkey-patch approach.

PR #674, opened 2012-07-24
[#1849] Allowed for UnloadDialog to be turned off via config

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/1849-allow-are-you-sure-you-want-to-leave-dialog-to-be-turned-off-via-config-option

### DIFF
--- a/src/default-config.json
+++ b/src/default-config.json
@@ -7,6 +7,7 @@
   "title": "Popcorn Maker",
   "ui": {
     "enabled": true,
+    "onLeaveDialog": true,
     "trackEventHighlight": "click"
   },
   "mediaDefaults": {
--- a/src/ui/ui.js
+++ b/src/ui/ui.js
@@ -78,6 +78,7 @@ define( [ "core/eventmanager", "./toggler", "./logo-spinner", "./context-button"
         _state = true,
         _logoSpinner,
         _uiConfig = butter.config,
+        _uiOptions = _uiConfig.value( "ui" ),
         _this = this;
 
     EventManagerWrapper( _this );
@@ -114,8 +115,10 @@ define( [ "core/eventmanager", "./toggler", "./logo-spinner", "./context-button"
     _element.appendChild( _areas.work.element );
     _element.appendChild( _areas.tools.element );
 
-    if( _uiConfig.value( "ui" ).enabled !== false ){
-      UnloadDialog( butter );
+    if ( _uiOptions.enabled ) {
+      if ( _uiOptions.onLeaveDialog ) {
+        UnloadDialog( butter );
+      }
       document.body.classList.add( "butter-header-spacing" );
       document.body.classList.add( "butter-tray-spacing" );
       document.body.appendChild( _element );
@@ -147,13 +150,13 @@ define( [ "core/eventmanager", "./toggler", "./logo-spinner", "./context-button"
     };
 
     this.load = function( onReady ){
-      if( _uiConfig.value( "ui" ).enabled !== false ){
+      if( _uiOptions.enabled ){
         var loadOptions = {};
 
         // Determine if we should load a pre-built CSS file for Butter (e.g.,
         // the deployment case, post `node make`), or whether we need to load
         // the LESS file directly and parse it into CSS (e.g., the dev case).
-        if( _uiConfig.value( "cssRenderClientSide" ) === true ){
+        if( _uiConfig.value( "cssRenderClientSide" ) ){
           loadOptions.type = "less";
           loadOptions.url = BUTTER_LESS_FILE;
         } else {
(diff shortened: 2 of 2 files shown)


ghost:
#### From: Bot.io (butyr.org)

---

# Received

Command **cmd_preview** from @ScottDowne received. Current queue size: 0

Live output at: http://botio.butyr.org:8000/34dba1da7676fa1/output.txt

ghost:
#### From: Bot.io (butyr.org)

---

# Received

Command **cmd_check** from @ScottDowne received. Current queue size: 1

Live output at: http://botio.butyr.org:8000/6b231f1e3d19a99/output.txt

ghost:
#### From: Bot.io (butyr.org)

---

# Success

Full output at http://botio.butyr.org:8000/34dba1da7676fa1/output.txt

Total script time: 0.77 mins

Server is running on http://pr674.butyr.org/templates/test.html

ghost:
#### From: Bot.io (butyr.org)

---

# Success

Full output at http://botio.butyr.org:8000/6b231f1e3d19a99/output.txt

Total script time: 0.44 mins

src/default-config.json
```diff
@@ -3,6 +3,7 @@
   "baseDir": "../",
   "cssRenderClientSide": false,
   "snapshotHTMLOnReady": true,
+  "onLeaveDialog": true,
   "scrapePage": true,
   "title": "Popcorn Maker",
   "ui": {
```
ScottDowne: Should move onLeaveDialog into the ui block here.

src/ui/ui.js
```diff
@@ -114,8 +114,10 @@ define( [ "core/eventmanager", "./toggler", "./logo-spinner", "./context-button"
     _element.appendChild( _areas.work.element );
     _element.appendChild( _areas.tools.element );
 
-    if( _uiConfig.value( "ui" ).enabled !== false ){
-      UnloadDialog( butter );
+    if ( _uiConfig.value( "ui" ).enabled !== false ) {
+      if ( _uiConfig.value( "onLeaveDialog" ) !== false ) {
```
ScottDowne: Given the above comment, you just need:

_uiConfig.value( "ui" ).onLeaveDialog !== false

dseif: Couldn't this also just be shortened to check `_uiConfig.value( "ui" ).onLeaveDialog` seeing as it's going to return true or false

mjschranz: What if someone enables the UI, but doesn't want the UnloadDialog? (IE, like me for UI testing)

In this case, we need both checks. The onLeaveDialog only determines if we actually do setup for it, where as the UI enabled check will still handle adding in the header/tray elements.

Also, I only did checks like that because that's how they were present before. I figured Scott wanted to ensure that people actually put false/true as the values and not just truthy/false things.

ScottDowne: Right.

So if ui is enabled, leaveDialog can be disabled or enabled.

But with ui off, I think it makes sense to have the leaveDialog inside the ui block, because we don;t use the leaveDialog with ui off, so people might get confused as to why it is not showing. By putting it in the ui, it is obvious it is off if ui is off, but with ui on, it can be on or off. This was my intent, and I think you produced that.

dseif: Maybe I wasn't clear with what I said, but the following lines can be shortened to:

```
if ( _uiConfig.value( "ui" ).enabled ) {
  if ( _uiConfig.value( "ui" ).onLeaveDialog ) {
```

We should also store a reference to `_uiConfig.value( "ui" )`

ScottDowne: I think this line comment is not actually part of the diff, because he's already updated it, if I understand you.

https://github.com/mozilla/butter/pull/674/files#L1R117

src/ui/ui.js
```diff
@@ -469,7 +472,7 @@ define( [ "core/eventmanager", "./toggler", "./logo-spinner", "./context-button"
       _this.visible = true;
       _toggler.visible = true;
       ContextButton( butter );
-      if( _uiConfig.value( "ui" ).enabled !== false ){
+      if( _uiOptions.enabled !== false ){
```
dseif: can you update these last two occurences to look like the ones on 118 and 119

ghost:
#### From: Bot.io (butyr.org)

---

# Received

Command **cmd_preview** from @dseif received. Current queue size: 0

Live output at: http://botio.butyr.org:8000/03cb88b3427ffcf/output.txt

ghost:
#### From: Bot.io (butyr.org)

---

# Success

Full output at http://botio.butyr.org:8000/03cb88b3427ffcf/output.txt

Total script time: 1.09 mins

Server is running on http://pr674.butyr.org/templates/test.html

PR #691, opened 2012-07-30
[t1874] Have CSS Lint ignore -moz-appearance rule

humphd:


make.js
```diff
@@ -40,11 +40,49 @@ var path = require( "path" ),
...
+    ' --errors=' + errors +
+    ' --quiet --format=compact' +
+    ' ' + filename, { silent: true } ).output.split( /\r?\n/ )
+    .forEach( function( line ) {
+      if( !line ) {
+        return;
+      }
+      var number = "|" + lineRegex.exec( line )[1] + "|";
+      if( ignoreLines.indexOf( number ) === -1 ) {
+        console.log( line );
```
k88hudson: We mostly use `echo` in this file, maybe use that instead

make.js
```diff
@@ -77,10 +115,17 @@ function checkCSS() {
...
+  var files = ls( dir );
+  files.forEach( function( filename ) {
+    filename = join( dir, filename );
+    if( /\.css$/.test( filename ) ) {
+      checkCSSFile( filename, warnings, errors );
+    }
+  });
+
+  // CSS Lint seems to want to print 2 blank lines when done.
+  console.log('\n\n');
```
k88hudson: Same as above

PR #697, opened 2012-07-31
T1424

humphd:


make.js
```diff
@@ -166,6 +163,28 @@ target.submodules = function() {
   exec('git submodule update --init --recursive');
 };
 
+function checkHTMLFile( filename ) {
+  out = exec( HTML5LINT + " -h " + filename, { silent: true } ).output;
+  if( out ) {
```
jbuck: Spacing

make.js
```diff
@@ -166,6 +163,28 @@ target.submodules = function() {
...
+    echo( out );
+  }
+}
+
+target["check-html"] = function() {
+  echo('### Linting HTML - requires network access');
+
+  var files = find([
+    EDITORS_DIR,
+    join( SRC_DIR, "layouts" ),
```
jbuck: I'd probably take this directory out, unless we can force fragment parsing for these files; they'll never be standalone valid html files.

make.js
```diff
@@ -166,6 +163,28 @@ target.submodules = function() {
   exec('git submodule update --init --recursive');
 };
 
+function checkHTMLFile( filename ) {
+  out = exec( HTML5LINT + " -h " + filename, { silent: true } ).output;
```
jbuck: I think that adding the `-e` switch for errors only would be good because it'd get rid of the many `Warning: The character encoding of the document was not declared.` messages.

PR #712, opened 2012-08-02
[t1899] Added onchanged handlers in te editor.

secretrobotron:


src/editor/trackevent-editor.js
```diff
@@ -165,6 +173,14 @@ define( [ "util/lang", "util/keys", "./base-editor",
...
         trackEvent.update( updateOptions );
       }, false );
+
+      if ( element.type === "number" ) {
+        element.addEventListener( "change", function( e ) {
+          var updateOptions = {};
+          updateOptions[ propertyName ] = element.value;
+          trackEvent.update( updateOptions );
+        }, false );
+      }
```
dseif, line 183: Why only for type === number?

secretrobotron, line 183: Just so it's guarded. I only cared about the `number` type since everything else is already covered, afaik. What else would benefit from it?

dseif, line 183: I didn't realize we already handled textboxes and dropdowns, so this looks great.

PR #721, opened 2012-08-03
Bug 1920 - Copy editors directory to dist folder

jbuck:


make.js
```diff
@@ -401,6 +397,7 @@ target.package = function() {
 
   target.build();
 
+  cp('-R', 'editors', DIST_DIR);
```
humphd, line 400: Given that you do this in 3 places, can you make into a function?

PR #734, opened 2012-08-07
[1930] Windows machines run python code with the 'python' keyword

stanica:
Had to preface the normalize function with 'python' so it's recognized in Windows.


jbuck:
Staged on master: https://github.com/mozilla/butter/commit/1cda2e5862a54093875984da9148042f8e56af29

PR #756, opened 2012-08-09
[#1871] Plugins are no longer droppable in between tracks

dseif:


jbuck:
Staged on master: https://github.com/mozilla/butter/commit/5171f970dc0d0e7cc358d253c947600278424fcf

PR #788, opened 2012-08-14
T1980

humphd:


src/embed.js
```diff
@@ -1,9 +1,15 @@
-/* This Source Code Form is subject to the terms of the MIT license
```
jbuck, line 1: I don't think we should delete the license

PR #811, opened 2012-08-17
[#1721] Complete implementation of URL config for embed of iframe in 3rd...

humphd:
... party sites


jbuck:
Closing since humph is away, and put the ticket into the slush pile

PR #841, opened 2012-08-21
[#2074] Make sure LESS is compiled before linting

cadecairos:
Lighthouse ticket: https://webmademovies.lighthouseapp.com/projects/65733-popcorn-maker/tickets/2074-transition-css-is-linted-before-its-compiled


make.js
```diff
@@ -156,8 +156,9 @@ function checkJS(){
 
 target.all = function() {
   target.submodules();
-  target.check();
```
jbuck, line 159: Why not just switch this line with `target.build()` below?

PR #880, opened 2012-08-28
[#2117] Hacky fix to make opera happy when seeking the timeline, video i...

dseif:
...s still janky

### DIFF
--- a/src/core/media.js
+++ b/src/core/media.js
@@ -47,7 +47,6 @@
                 _this.dispatch( "mediavolumechange", _popcornWrapper.volume );
               },
               timeupdate: function(){
-                _currentTime = _popcornWrapper.currentTime;
                 _this.dispatch( "mediatimeupdate", _this );
               },
               pause: function(){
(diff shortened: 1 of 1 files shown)


src/core/media.js
```diff
@@ -47,7 +47,6 @@
                 _this.dispatch( "mediavolumechange", _popcornWrapper.volume );
               },
               timeupdate: function(){
-                _currentTime = _popcornWrapper.currentTime;
```
secretrobotron, line 50: Is it possible to get `timeupdate` to occur without touching `play` or `seeking`? If so, it'll be a problem to remove this line.

dseif, line 50: I don't think so, seeking seems to occur from setting the currentTime (either programmatically or through a manual seek) and playing handles when it is playing so I think we should be fine

secretrobotron, line 50: Fine with me. Let's just add a comment note here to make sure it's clear what's happening when we open up the hood on this issue in the future.

PR #881, opened 2012-08-29
[#2123] Editor tests failing due to breadcrumb additions

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/2123-editor-tests-failing


src/editor/trackevent-editor.js
```diff
@@ -121,7 +121,9 @@ define( [ "util/lang", "util/keys", "./base-editor",
         editorTitle.appendChild( document.createTextNode( trackEvent.type ) );
       }
 
-      oldTitleEl.parentNode.replaceChild( breadcrumbsLayout, oldTitleEl );
```
secretrobotron: Actually, I'd argue this `if` block can wrap even more code. If we're not going to use any of the things created above because `oldTitlteE1` doesn't exist, let's not even bother.

PR #885, opened 2012-08-29
[#2129] Lowered the timeout time to recheck if the map was ready to some thing much more reasonable

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/2129-google-maps-takes-a-moment-to-load


templates/assets/plugins/googlemap/popcorn.googlemap.js
```diff
@@ -379,7 +379,7 @@ var googleCallback;
           } else {
             setTimeout(function () {
               isMapSetup();
-            }, 500);
```
cadecairos, line 382: missing a space

PR #889, opened 2012-08-30
[#2137] Added zindex to wikipedia, webpage, titles, and image. Fixed bug...

ScottDowne:
... where a track event can be updated before media is set.

### DIFF
--- a/src/core/trackevent.js
+++ b/src/core/trackevent.js
@@ -117,7 +117,8 @@ define( [
 
       var newStart = _popcornOptions.start,
           newEnd = _popcornOptions.end,
-          manifestOptions;
+          manifestOptions,
+          media;
 
       if ( !isNaN( updateOptions.start ) ) {
         newStart = TimeUtil.roundTime( updateOptions.start );
@@ -136,11 +137,12 @@ define( [
       if ( newStart >= newEnd ) {
         throw new TrackEventUpdateException( "start-greater-than-end", "[start] must be less than [end]." );
       }
-      if ( _track && _track._media && _track._media.ready ) {
-        var media = _track._media;
-        if( ( newStart > media.duration ) ||
-            ( newEnd > media.duration ) ||
-            ( newStart < 0 ) ) {
+      if ( _track && _track._media ) {
+        media = _track._media;
+        if( media.ready &&
+            ( newStart > media.duration ||
+            newEnd > media.duration ||
+            newStart < 0 ) ) {
           throw new TrackEventUpdateException( "invalid-times", "[start] or [end] are not within the duration of media" );
         }
       }
@@ -162,8 +164,8 @@ define( [
           if ( !( "target" in manifestOptions ) && updateOptions.target ) {
             _popcornOptions.target = updateOptions.target;
           }
-          if ( "zindex" in manifestOptions ) {
-            _popcornOptions.zindex = _track._media.maxPluginZIndex - _track.order;
+          if ( "zindex" in manifestOptions && media ) {
+            _popcornOptions.zindex = media.maxPluginZIndex - _track.order;
           }
         }
       }
--- a/src/editor/trackevent-editor.js
+++ b/src/editor/trackevent-editor.js
@@ -439,7 +439,11 @@ define([ "util/lang", "util/keys", "util/time", "./base-editor",
             element.checked = popcornOptions[ option ];
           }
           else {
-            element.value = popcornOptions[ option ];
+            if ( typeof popcornOptions[ option ] !== "undefined" ) {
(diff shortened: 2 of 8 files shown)


src/core/trackevent.js
```diff
@@ -117,7 +117,7 @@ define( [
 
       var newStart = _popcornOptions.start,
           newEnd = _popcornOptions.end,
-          manifestOptions;
+          manifestOptions, media;
```
secretrobotron: On a new line please (for convention).

src/core/trackevent.js
```diff
@@ -136,11 +136,12 @@ define( [
...
-      if ( _track && _track._media && _track._media.ready ) {
-        var media = _track._media;
-        if( ( newStart > media.duration ) ||
+      if ( _track && _track._media ) {
+        media = _track._media;
+        if( _track._media.ready &&
+            ( ( newStart > media.duration ) ||
             ( newEnd > media.duration ) ||
-            ( newStart < 0 ) ) {
+            ( newStart < 0 ) ) ) {
```
secretrobotron: There is an excess of brackets here. Do you mind cleaning this if clause up?

templates/assets/plugins/titles/popcorn.titles.js
```diff
@@ -180,7 +183,7 @@
       target.appendChild( container );
 
       container.style.fontStyle = options.fontItalics ? "italic" : "normal";
-      container.style.color = validateHexColor( options.fontColor ) ? options.fontColor : "#668B8B";
+      container.style.color = options.fontColor && validateHexColor( options.fontColor ) ? options.fontColor : "#668B8B";
```
secretrobotron: Are you just sneaking this in here? Doesn't look like it has anything to do with z-index.

ScottDowne: This was exposed by all of this.

validateHexColor assumes fontColor is always defined. It is not always defined.

I'll take a closer look as to why it is undefined when it probably shouldn't be.

src/core/trackevent.js
ScottDowne: I don't know what you mean. I actually think this is right.

Can you elaborate?

ghost:
#### From: Bot.io (butyr.org)

---

# Received

Command **cmd_check** from @ScottDowne received. Current queue size: 2

Live output at: http://botio.butyr.org:8000/ee3c92267dcca6a/output.txt

src/core/trackevent.js
secretrobotron: Each one of the lines in this `if` evaluation has brackets. That's just unnecessary and confusing, no?

templates/assets/plugins/titles/popcorn.titles.js
```diff
@@ -11,6 +11,8 @@
   
    **/
 
+  var FONT_COLOR = "#000";
```
secretrobotron: Perhaps `DEFAULT_FONT_COLOR`?

src/editor/trackevent-editor.js
```diff
@@ -424,7 +424,7 @@ define( [ "util/lang", "util/keys", "./base-editor",
             element.checked = popcornOptions[ option ];
           }
           else {
-            element.value = popcornOptions[ option ];
+            element.value = popcornOptions[ option ] || manifestOptions[ option ].default || "";
```
secretrobotron: I don't think this will work. Try using the flickr plugin and erasing 'kittens' from the `Tag Search` textbox. As soon as you unfocus, it reappears.

templates/assets/plugins/titles/popcorn.titles.js
ScottDowne: Perhaps!

src/core/trackevent.js
ScottDowne: Haha, my bad.

That was how it already was. For some reason they were all in their own bracket.

From here, I thought you were saying I should remove what I added.

I did clean it up, though.

src/core/trackevent.js
```diff
@@ -136,11 +137,12 @@ define( [
...
         throw new TrackEventUpdateException( "start-greater-than-end", "[start] must be less than [end]." );
       }
-      if ( _track && _track._media && _track._media.ready ) {
-        var media = _track._media;
-        if( ( newStart > media.duration ) ||
-            ( newEnd > media.duration ) ||
-            ( newStart < 0 ) ) {
+      if ( _track && _track._media ) {
+        media = _track._media;
+        if( _track._media.ready &&
```
mjschranz: Really only a nit. Couldn't you just do media.ready here as well?

PR #1011, opened 2012-09-22
Lawnchair

humphd:


src/ui/unload-dialog.js
```diff
@@ -2,42 +2,22 @@ define([], function() {
...
-    var eventFunction = function() {
-      if ( !changed ) {
-        changed = true;
-        window.onbeforeunload = areYouSure;
-      }
-    };
-
-    butter.listen( "mediaready", mediaReady );
+    butter.listen( "projectchanged", function() {
+      changed = true;
```
secretrobotron: I think you eviscerated the change detection mechanism here such that `changed` is pointless.

src/core/project.js
```diff
@@ -0,0 +1,247 @@
...
+          _template = template;
+        },
+        enumerable: true
+      },
+      data: {
+        get: function(){
+          return buildDataObject();
+        }
+//,
+        // XXXhumph: can we dump this?  Seems odd to have it.  Should we only build internally?
```
secretrobotron: For now, yes. No need for `set` here.

src/core/project.js
```diff
@@ -0,0 +1,247 @@
...
+    };
+
+    // Export project data as JSON string (e.g., for use with project.import())
+    _this.export = function() {
+      return JSON.stringify( buildDataObject() );
+    };
+
+    // XXXhumph this looks like dead code to me...
+    // Clear project data
+/**
```
secretrobotron: I think it is. It's saved in history. No need to keep it around.

src/core/project.js
```diff
@@ -0,0 +1,247 @@
...
+          _isDirty = false;
+          __butterStorage.remove( "butter-backup-project", function(){});
+        }
+        callback( e );
+      });
+    };
+
+    Object.defineProperties( _this, {
+      id: {
+        get: function(){
```
secretrobotron: Why do all of these properties have simple accessors? Can they just be properties on `_this` instead? i.e. `_this.id = null;` ... `butter.project.id = 'something';`

secretrobotron: Or was the intent to fire a `projectchanged` when they are changed?

src/core/project.js
```diff
@@ -0,0 +1,247 @@
...
+        // XXXhumph: can we dump this?  Seems odd to have it.  Should we only build internally?
+//        set: function( data ){
+//          _p.data = data;
+//          console.log('set data!');
+//        },
+//        enumerable: true
+      }
+    });
+
+    // XXXhumph: no logical place to kill this interval...
```
secretrobotron: In conjunction with https://github.com/mozilla/butter/pull/1011/files#L2R117, I think you're safe not to bother storing `_backupInterval`.

src/core/project.js
```diff
@@ -0,0 +1,247 @@
...
+    // the app's data states changing, and want to track.
+    butter.listen( "mediaready", function mediaReady() {
+      butter.unlisten( "mediaready", mediaReady );
+
+      function invalidate() {
+        // Project is dirty, needs backup
+        _isDirty = true;
+
+        // Let consumers know that the project changed
+        _this.dispatch( "projectupdated" );
```
secretrobotron: The event everything else listens to is `projectchanged` (e.g. https://github.com/mozilla/butter/pull/1011/files#L5R763).

src/core/project.js
```diff
@@ -0,0 +1,290 @@
...
+    // Manage access to project properties.  Some we only want
+    // to be read (and managed by db/butter), others we want to
+    // affect save logic.
+    Object.defineProperties( _this, {
+      "id": {
+        get: function() {
+          return _id;
+        },
+        enumerable: true,
+        configurable: false
```
secretrobotron: I don't think you need to set `configurable` for each one of these entries. It's set to `false` by default: https://developer.mozilla.org/en-US/docs/JavaScript/Reference/Global_Objects/Object/defineProperty#Description

src/core/project.js
```diff
@@ -0,0 +1,290 @@
...
+define( [ 'core/eventmanager', 'core/media', 'lawnchair' ],
+        function( EventManager, Media, Lawnchair ){
+
+  // Local storage provider for backups
+  var __butterStorage = new Lawnchair( { name: "Butter-Storage" }, function(){} );
+
+  function Project( butter ) {
+
+    var _this = this,
+        // Whether or not to save (project data has changed)
```
secretrobotron: Just a suggestion: my preference for style in blocks like these is to either separate commented variables entirely with whitespace between them and others, or to place comments a few tabs over to the right on the same line as the code they describe.

src/core/project.js
```diff
@@ -0,0 +1,290 @@
...
+      var projectJSON = JSON.stringify({
+        id: _id,
+        name: _name,
+        template: _template,
+        author: _author,
+        data: buildDataObject()
+      });
+
+      // Save to local storage first in case network is down.
+      backupData();
```
secretrobotron: Think it's worth it to save a bit of computing power here and only call `buildDataObject` once? It gets called when compiling the JSON object, and through `backupData` right afterward. Not sure if that would produce harder-to-manage code.

src/core/project.js
```diff
@@ -0,0 +1,290 @@
...
+          __butterStorage.remove( "butter-backup-project", function(){} );
+
+          // Grab id generated by server and store
+          _id = e.project._id;
+
+          // Now Publish and get URLs for embed
+          butter.cornfield.publish( _id, function( e ) {
+            if( e.error === "okay" ) {
+              // Save + Publish is OK
+              _isSaved = true;
```
secretrobotron: `_isSaved` is really `_isPublished`, right? Might be better to rename that variable to discern between `project.isSaved`, `_isSaved`, and `_isDirty`.

src/core/project.js
```diff
@@ -0,0 +1,290 @@
...
+    if( _backupIntervalMS > 0 ) {
+      setInterval( backupData, _backupIntervalMS );
+    }
+  }
+
+  // Check for an existing project that was autosaved but not saved.
+  // Returns new Project object if found, otherwise null.
+  Project.checkForBackup = function( butter, callback ) {
+    // See if we already have a project autosaved from another session.
+    __butterStorage.get( "butter-backup-project", function( obj ){
```
secretrobotron: Spacing.

src/core/project.js
```diff
@@ -0,0 +1,290 @@
...
+      data.backupDate = (new Date()).toJSON();
+      __butterStorage.save({ key: "butter-backup-project", value: JSON.stringify( data ) });
+    };
+
+    // Save and Publish a project.  Saving only happens if project data needs
+    // to be saved (i.e., it has been changed since last save, or was never
+    // saved before).
+    _this.save = function( callback ) {
+      // Don't save if there is nothing new to save.
+      if( _this.isSaved ) {
```
secretrobotron: Spacing.

src/core/project.js
```diff
@@ -0,0 +1,290 @@
...
+    };
+
+    // Export project data as JSON string (e.g., for use with project.import())
+    _this.export = function() {
+      return JSON.stringify( buildDataObject() );
+    };
+
+    // Expose backupData() to make testing possible
+    var backupData = _this.backupData = function() {
+      if( !_isDirty ) {
```
secretrobotron: Spacing.

src/core/project.js
```diff
@@ -0,0 +1,290 @@
...
+          _id = e.project._id;
+
+          // Now Publish and get URLs for embed
+          butter.cornfield.publish( _id, function( e ) {
+            if( e.error === "okay" ) {
+              // Save + Publish is OK
+              _isSaved = true;
+              _publishUrl = e.url;
+              _iframeUrl = _publishUrl.replace( "/v/", "/e/" );
+            }
```
secretrobotron: If a publish failure does happen here for some reason, what happens? I _think_ the "Preview" button will stay hidden, and the "Save" button will be disabled. So, to see your published project, you have to change something in order to try to save again, right?

That seems broken, but probably worthy of a completely separate ticket.

src/core/project.js
```diff
@@ -0,0 +1,290 @@
...
+      setInterval( backupData, _backupIntervalMS );
+    }
+  }
+
+  // Check for an existing project that was autosaved but not saved.
+  // Returns new Project object if found, otherwise null.
+  Project.checkForBackup = function( butter, callback ) {
+    // See if we already have a project autosaved from another session.
+    __butterStorage.get( "butter-backup-project", function( obj ){
+      if( !obj ){
```
secretrobotron: Spacing.

tools/build.js
```diff
@@ -23,6 +24,9 @@
   // provides. Path is relative to baseUrl.
   name: '../tools/almond',
 
+  // Strip extra licenses
+  preserveLicenseComments: false,
```
secretrobotron, line 27: Why do you want this change here? Won't this strip out licenses that might matter, or is everything the exact same license?

src/core/project.js
humphd: I don't think we want to introduce the idea of publish vs. save.  When we "save" a project, part of that process is writing out embeds (i.e., publish), but it's a detail.  You can't save to the db separate from writing the embed files, so the distinction is not helpful.  I think just having isSaved is useful and correct.

tools/build.js
humphd, line 27: It ends up having our license more than once, plus any license block in any of our files.  The result is a mess of 40+ lines of comments that look weird.  This was a nice way to dump it all.  If we don't do this, we need to do some work to clean-up our license header end-result.  New bug or leave this?

src/core/project.js
```diff
@@ -0,0 +1,295 @@
...
+        _backupIntervalMS = butter.config.value( "backupInterval" )|0;
+
+    function invalidate() {
+      // Project is dirty, needs backup
+      _isDirty = true;
+
+      // Let consumers know that the project changed
+      _this.dispatch( "projectchanged" );
+
+console.log("Project is dirty");
```
jbuck: nuke it

src/core/project.js
```diff
@@ -0,0 +1,295 @@
...
+
+        // Whether or not the project is saved to the db and published.
+        // The notion of "saving" to consumers of this code is unware of
+        // the save vs. publish distinction. As such, we use isSaved externally
+        // and isPublished internally, where Publish follows Save and is
+        // more correct.
+        _isPublished = false,
+
+        // How often to backup data in ms. If 0, no backups are done.
+        _backupIntervalMS = butter.config.value( "backupInterval" )|0;
```
jbuck, line 33: `|| 0` right?

src/core/project.js
```diff
@@ -0,0 +1,295 @@
...
+
+    // Expose backupData() to make testing possible
+    var backupData = _this.backupData = function() {
+      if ( !_isDirty ) {
+        return;
+      }
+      var data = _this.data;
+      data.backupDate = (new Date()).toJSON();
+      __butterStorage.setItem( "butter-backup-project", JSON.stringify( data ) );
+console.log("Projet backed-up to localStorage");
```
jbuck: nuke it

tools/build.js
jbuck, line 27: We can just add our license header to /tools/wrap.start and /tools/wrap.end too

humphd, line 27: It's already in wrap.start

src/core/project.js
humphd, line 33: No, |0 to get it to be an integer always, and 0 if it's anything else.

PR #1046, opened 2012-10-01
[t2355] Resizing blocked in both directions

secretrobotron:
1. Added resize event code path to dragndrop and track-container
2. saved-data has two events for ease of testing
3. Fixed and ammended findOverlappingTrackEvent for adjustable input rect

### DIFF
--- a/src/timeline/track-container.js
+++ b/src/timeline/track-container.js
@@ -139,6 +139,26 @@ define( [ "core/logger", "util/dragndrop", "./ghost-manager" ],
       }
     }
 
+    function onTrackEventResized( trackEvent, x, w, resizeEvent, direction ) {
+      var trackView = trackEvent.track.view,
+          overlappingTrackEvent,
+          leftOffset = _container.getBoundingClientRect().left,
+          trackEventView = trackEvent.view,
+          trackEventSpacing = trackEventView.element.clientLeft;              // Compensate for border width
+
+      overlappingTrackEvent = trackView.findOverlappingTrackEvent( trackEventView, x + leftOffset + trackEventSpacing, w );
+
+      if ( overlappingTrackEvent ) {
+        if ( direction === 'right' ) {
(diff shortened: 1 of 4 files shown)


ghost:
#### From: Bot.io (butyr.org)

---

# Received

Command **cmd_preview** from @brett-etherworks received. Current queue size: 2

Live output at: http://botio.butyr.org:8000/acc816c1939c65f/output.txt

ghost:
#### From: Bot.io (butyr.org)

---

# Success

Full output at http://botio.butyr.org:8000/acc816c1939c65f/output.txt

Total script time: 2.86 mins

Server is running on http://pr1046.butyr.org/templates/basic/
Run the Test Suite: http://pr1046.butyr.org/test/

src/timeline/track-container.js
```diff
@@ -139,6 +139,25 @@ define( [ "core/logger", "util/dragndrop", "./ghost-manager" ],
...
+          overlappingTrackEvent,
+          leftOffset = _container.getBoundingClientRect().left,
+          trackEventView = trackEvent.view;
+
+      overlappingTrackEvent = trackView.findOverlappingTrackEvent( trackEventView, x + leftOffset, w );
+
+      if ( overlappingTrackEvent ) {
+        if ( direction === 'right' ) {
+          // Use clientLeft to compensate for border (https://developer.mozilla.org/en-US/docs/DOM/element.clientLeft)
+          resizeEvent.block( overlappingTrackEvent.view.element.offsetLeft - overlappingTrackEvent.view.element.clientLeft * 2 );
```
ScottDowne: How safe do you feel about the use of clientLeft?

You felt the need to add the docs for it, so, I wonder what the risk is for using it are.

If you wanted to not use clientLeft, you can instead use offsetWidth - clientWidth to get the total padding. Divide that by 2 to get the left or right padding assuming left and right have the same padding. Would this make things safer? ClientWidth is also not standard, but we use that all over the place. Is it generally better than clientLeft?

Anyway, not going to fail it on this alone. Just want the thoughts on this.

PR #1047, opened 2012-10-01
[#2389] Stop growing of track events on drop and resize.

ScottDowne:


css/tray-timeline.less
```diff
@@ -98,7 +98,7 @@
   cursor: move;
   display: block;
   height: @_trackEventHeight;
-  padding: 1px;
+  padding: 1px 0px;
```
ScottDowne: Self reminder, change the 0px to 0.

PR #1057, opened 2012-10-02
[#2334] Moving a trackevent between tracks now handles editor state prop...

dseif:
...erly


mjschranz:
#1085 is being used instead.

PR #1062, opened 2012-10-03
[#2394] Added a basic color input handler for the trackevent editors.

mjschranz:
This way we can provide more elegant support/error handling for input type=color on browsers that don't support it

https://webmademovies.lighthouseapp.com/projects/65733/tickets/2394-add-error-handling-for-incorrect-colours-in-editors

### DIFF
--- a/src/editor/trackevent-editor.js
+++ b/src/editor/trackevent-editor.js
@@ -9,17 +9,35 @@ define([ "util/lang", "util/keys", "util/time", "./base-editor", "ui/widget/tool
 
   var __defaultLayouts = LangUtils.domFragment( DEFAULT_LAYOUT_SNIPPETS ),
       __googleFonts = [
-                        "Gentium Book Basic",
-                        "Lato",
-                        "Vollkorn",
-                        "Merriweather",
-                        "Gravitas One",
-                        "PT Sans",
-                        "Open Sans",
-                        "Bangers",
-                        "Fredoka One",
-                        "Covered By Your Grace"
-                      ];
+        "Gentium Book Basic",
+        "Lato",
+        "Vollkorn",
+        "Merriweather",
+        "Gravitas One",
+        "PT Sans",
+        "Open Sans",
+        "Bangers",
+        "Fredoka One",
+        "Covered By Your Grace"
+      ],
+      __colorHexCodes = {
+        "black": "#000000",
+        "silver": "#c0c0c0",
+        "gray": "#808080",
+        "white": "#ffffff",
+        "maroon": "#800000",
+        "red": "#ff00000",
+        "purple": "#800080",
+        "fuchsia": "#ff00ff",
+        "green": "#008000",
+        "lime": "#00ff00",
+        "olive": "#808000",
+        "yellow": "#ffff00",
+        "navy": "#000080",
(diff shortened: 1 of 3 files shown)


templates/assets/editors/popup/popup-editor.js
```diff
@@ -157,6 +157,20 @@
             else if ( option.elementType === "input" ) {
               if ( option.element.type === "checkbox" ) {
                 _this.attachCheckboxChangeHandler( option.element, option.trackEvent, key, updateTrackEventWithoutTryCatch );
+              }
```
ScottDowne: trailing whitespace.

templates/assets/editors/popup/popup-editor.js
```diff
@@ -157,6 +157,20 @@
             else if ( option.elementType === "input" ) {
               if ( option.element.type === "checkbox" ) {
                 _this.attachCheckboxChangeHandler( option.element, option.trackEvent, key, updateTrackEventWithoutTryCatch );
+              } 
+              else if ( key === "fontColor" ) {
+                _this.attachColorChangeHandler( option.element, option.trackEvent, key, function( te, prop, message ) {
```
ScottDowne: don't make functions in a loop.

templates/assets/editors/text/text-editor.js
```diff
@@ -101,6 +101,19 @@
             else if ( option.elementType === "input" ) {
               if ( option.element.type === "checkbox" ) {
                 _this.attachCheckboxChangeHandler( option.element, option.trackEvent, key, updateTrackEventWithoutTryCatch );
+              } else if ( key === "fontColor" ) {
+                _this.attachColorChangeHandler( option.element, option.trackEvent, key, function( te, prop, message ) {
```
ScottDowne: don't make functions in a loop.

src/editor/trackevent-editor.js
```diff
@@ -187,6 +205,53 @@ define([ "util/lang", "util/keys", "util/time", "./base-editor",
...
+        if ( value.indexOf( "#" ) === -1 ) {
+
+          for ( i in __colorHexCodes ) {
+            if ( __colorHexCodes.hasOwnProperty( i ) ) {
+              if ( i === value.toLowerCase() ) {
+                flag = false;
+                break;
+              } else {
+                flag = true;
+              }
```
dseif: we can just set flag to `true` on line 214 instead of having this else statement as it breaks and sets it to false

templates/assets/editors/popup/popup-editor.js
```diff
@@ -128,6 +128,19 @@
           }, false );
         }
 
+        function colorCallback( te, prop, message ) {
+          var updateOptions = {
+                fontColor: prop.fontColor
+              };
```
dseif: don't need to declare update options

templates/assets/editors/text/text-editor.js
```diff
@@ -87,12 +87,24 @@
         var key,
             option;
 
+        function colorCallback( te, prop, message ) {
+          var updateOptions = {
+                fontColor: prop.fontColor
+              };
```
dseif: same here

PR #1085, opened 2012-10-05
[#2334] Ugly fix for preventing an open trackevent editor from being closed when moving a trackevent to a new layer

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/2334-moving-an-event-to-a-new-track-will-close-its-editor


src/timeline/trackhandles.js
```diff
@@ -69,6 +69,15 @@ define( [ "dialog/dialog", "util/dragndrop", "util/lang", "text!layouts/track-ha
           events: {
             submit: function( e ){
               if( e.data === true ){
+                var trackEvents = track.trackEvents,
+                    currentEditor = butter.editor.currentEditor,
+                    isTrackEventEditor = currentEditor.getTrackEvent;
+                for ( var i = 0, l = trackEvents.length; i < l; i++ ) {
```
secretrobotron: You should use `track.getTrackEventById( trackEvent.id )` here!

secretrobotron: Actually, why do you need this here at all if `closeEditor` does it too?

src/editor/module.js
```diff
@@ -78,8 +78,18 @@ define( [ "core/eventmanager", "core/trackevent", "./editor",
      *
      * Closes the currently opened editor and opens the default one in it's place
      *
+     * @param {TrackEvent} trackEvent: When specified will check if it matches the current editors track event and
+     *                                 will prevent it from closing if so.
      */
-    _this.closeEditor = function() {
+    _this.closeEditor = function( trackEvent ) {
```
secretrobotron: Since the functionality of `closeEditor` is warping pretty drastically from what someone would expect it to do given the simple name `closeEditor`, let's rename it to something more appropriate. Perhaps `closeTrackEventEditor` or something more concise.

src/editor/module.js
```diff
@@ -78,8 +78,18 @@ define( [ "core/eventmanager", "core/trackevent", "./editor",
...
+      var currentEditor,
+          isTrackEventEditor;
```
secretrobotron: Looks like this function isn't quite done. `isTrackEventEditor` should check `_currentEditor`, no?

src/timeline/trackhandles.js
mjschranz: I'm fairly certain I can't use track.getTrackEventById here. I don't have reference to any track events so I don't know which one that's on a track is opened.

Still, this can be improved.

src/editor/module.js
```diff
@@ -86,6 +86,24 @@ define( [ "core/eventmanager", "core/trackevent", "./editor",
     };
 
     /**
+     * Member: closeTrackEventEditor
+     *
+     * Closes the currently opened editor and opens the default one in it's place
```
secretrobotron: "it's" shouldn't have an apostrophe here.

src/editor/module.js
```diff
@@ -86,6 +86,24 @@ define( [ "core/eventmanager", "core/trackevent", "./editor",
...
+     *
+     * @param {TrackEvent} trackEvent: When specified will check if it matches the current editors track event and
```
secretrobotron: Comma after "When specified", and apostrophe on "current editors". I also don't really understand this comment. Namely, the "...and will prevent it from closing if so" part. Can you re-word it?

src/core/trackevent.js
```diff
@@ -73,15 +71,18 @@ define( [ "./logger", "./eventmanager", "util/lang", "util/time", "./views/track
...
     /**
-     * Member: setPopcornWrapper
+     * Member: bind
      *
-     * Sets the PopcornWrapper object. Subsequently, PopcornWrapper can be used to directly manipulate Popcorn track events.
+     * Binds the TrackEvent to its dependencies.
      *
-     * @param {Object} newPopcornWrapper: PopcornWrapper object or null
+     * @param {Object} track: The track the trackevent will inhabit.
+     * @param {Object} popcornWrapper: a reference to a popcornWrapper object the wraps various functionality for modifying Popcorn data.
```
secretrobotron: Two small corrections here: `a reference to a popcornWrapper object the wraps` -> `a reference to a PopcornWrapper object that wraps`

PR #1088, opened 2012-10-05
[#2416]

cadecairos:
1. Create a 'createTrackEvent' method on each Butter Target object
2. Set a default target for the basic template
3. Make double clicking on a plugin list element call 'createTrackEvent' on the defaultTarget, if it exists

Ticket in Lighthouse: https://webmademovies.lighthouseapp.com/projects/65733/tickets/2416

### DIFF
--- a/src/plugin/plugin-list.js
+++ b/src/plugin/plugin-list.js
@@ -45,9 +45,12 @@ define( [ "util/dragndrop", "util/lang", "editor/editor", "text!layouts/plugin-l
         }
       });
 
-      // request a new trackEvent on the currentMedia
+      // Create a new trackEvent on the defaultTarget
       element.addEventListener( "dblclick", function() {
-        butter.currentMedia.dispatch( "trackeventrequested", element );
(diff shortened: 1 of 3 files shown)


src/plugin/plugin-list.js
```diff
@@ -45,9 +45,12 @@ define( [ "util/dragndrop", "util/lang", "editor/editor", "text!layouts/plugin-l
         }
       });
 
-      // request a new trackEvent on the currentMedia
+      // Create a new trackEvent on the defaultTarget
       element.addEventListener( "dblclick", function() {
-        butter.currentMedia.dispatch( "trackeventrequested", element );
+        // Make sure it exists ( It doesn't if it's not specified in the template's html )
```
secretrobotron, line 50: This isn't totally true. It will just default to the first target if one isn't specified.

secretrobotron, line 50: ... when #2024 lands. :/

cadecairos, line 50: ahha! :P

PR #1109, opened 2012-10-10
[t2464] Better TrackEvent update error detection and recovery

secretrobotron:
1. Created src/core/observer to give classes ability to notify subscribers of events (synchronous).
2. Created src/strings to start holding Butter-wide strings.
3. Removed independent setErrorState, updateTrackEvent\* from editors.
4. Added centralized updateTrackEventSafe to TrackEventEditor.
5. Rerouted TrackEvent update mechanisms to use updateTrackEventSafe in editors.
6. Added trackevent update subscriber to src/core/track to reject an update if overlap occurs.

### DIFF
--- a/src/core/observer.js
+++ b/src/core/observer.js
@@ -1,0 +1,159 @@
+/* This Source Code Form is subject to the terms of the MIT license
+ * If a copy of the MIT license was not distributed with this file, you can
+ * obtain one at http://www.mozillapopcorn.org/butter-license.txt */
+
+define([], function(){
+
+  /**
+   * Notification
+   *
+   * A Notification object is passed to subscribers when a notification occurs. It
+   * describes the notification, encompassing references to the notification origin,
+   * the name of the notification, and some data to assist. Notifications can be
+   * cancelled by calling the `cancel` function, and a reason can be specified to
+   * pass on to the body which issued the notification.
+   *
+   * @param {Object} origin: The object which issued the notification.
+   * @param {String} type: The type of notification.
+   * @param {Object} data: Arbitrary data to associate with the notification.
+   */
+  function Notification( origin, type, data ) {
+    var _cancelledReason;
+
+    /**
+     * cancel
+     *
+     * Cancels a notification and records a reason for doing so.
+     *
+     * @param {String} reason: The reason for canceling the notification.
+     */
+    this.cancel = function( reason ) {
+      _cancelledReason = reason || true;
+    };
+
+    Object.defineProperties( this, {
+      origin: {
+        value: origin,
+        enumerable: true
+      },
+      type: {
+        value: type,
+        enumerable: true
+      },
+      data: {
+        value: data,
+        enumerable: true
+      },
+      cancelledReason: {
+        enumerable: true,
+        get: function() {
+          return _cancelledReason;
+        }
+      },
+      cancelled: {
+        enumerable: true,
+        get: function() {
+          return !!_cancelledReason;
+        }
+      }
+    });
+  }
+
+  /**
+   * __subscribe
+   *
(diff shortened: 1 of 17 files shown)


src/core/observer.js
```diff
@@ -0,0 +1,156 @@
...
+      }
+    }
+  }
+
+  /**
+   * __notify
+   *
+   * A class function which calls all the subscribers of a given notification type.
+   *
+   * @param {String} type: The type of notification to subscribe to.
```
humphd: "to subscribe to"?  Maybe, "the type of notification to determine which subscribers to notify"?

src/core/observer.js
```diff
@@ -0,0 +1,156 @@
...
+   * @param {Object} object: The object issuing the notification.
+   */
+  function __notify( type, data, subscriberDict, object ) {
+    var i, l,
+        subscribers = subscriberDict[ type ],
+        notification = new Notification( object, type, data );
+
+    if ( subscribers ) {
+      for ( i = 0, l = subscribers.length; i < l; ++i ) {
+        subscribers[ i ]( notification );
```
humphd, line 117: Is it worth checking whether the notification was cancelled by one of the subscribers?  If so, do you want to bail and not bother with the rest?

src/core/observer.js
```diff
@@ -0,0 +1,156 @@
...
+   * Gives an object the functionality to record and notify subscribers for typed notifications
+   * (simple implementation of Observer pattern).
+   *
+   * @param {Object} object: The object to extend with Observer functionality.
+   */
+  function extendObject( object ) {
+    var _subscribers = {};
+
+    if ( object.subscribe ) {
+      throw "Object already has Observer properties.";
```
humphd, line 139: Just return, there's no harm if it already exists, right?

src/core/trackevent.js
```diff
@@ -100,18 +105,29 @@ define( [ "./logger", "./eventmanager", "util/lang", "util/time", "./views/track
...
+        throw new TrackEventUpdateException( "invalid-end-time" );
       }
 
       if ( newStart >= newEnd ) {
-        throw new TrackEventUpdateException( "start-greater-than-end", "[start] must be less than [end]." );
+        throw new TrackEventUpdateException( "start-greater-than-end" );
+      }
+
+      // Synchronously notify observers that an update is happening.
+      // This action gives observers are chance to stop the trackevent from updating
```
humphd: s/are/a/

src/strings/trackevent-errors.js
```diff
@@ -0,0 +1,10 @@
+/* This Source Code Form is subject to the terms of the MIT license
```
humphd: I wonder at the value of separating things out like this, given that we don't do this anywhere else.  What's your thinking?

src/core/observer.js
secretrobotron, line 139: I did want to keep this `throw` as a protective measure in case someone is adding redundant code.

src/strings/trackevent-errors.js
secretrobotron: I thought it was a good idea to start doing it somewhere as a test. These messages were the easiest thing to pluck out of the UI because they were set mostly by inline strings in the code. Considering I set it up to transition easily into `require`'s l18n system, do you think it's not worth doing?

src/core/observer.js
secretrobotron, line 117: Done (just below here).

src/strings/trackevent-errors.js
humphd: Personally, I'd wait and do it en masse.  I don't think this provides us much value yet.

PR #1110, opened 2012-10-10
T2380

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/2380-when-you-try-to-select-in-image-it-clicks-through-and-loads-up-the-host-url

### DIFF
--- a/templates/assets/editors/image/image-editor.js
+++ b/templates/assets/editors/image/image-editor.js
@@ -252,6 +252,16 @@
           });
         });
 
+        _this.createTooltip( _linkInput, {
+          name: "image-link-tooltip" + Date.now(),
+          element: _linkInput.parentElement,
+          message: "Links will be clickable when shared.",
+          top: "105%",
+          left: "50%",
+          hidden: true,
+          hover: false
+        });
+
         _this.attachInputChangeHandler( _linkInput, trackEvent, "linkSrc", function( te, prop ) {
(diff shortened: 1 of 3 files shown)


templates/assets/editors/image/image-editor.js
```diff
@@ -378,6 +388,22 @@
...
+    }
+
+    function addLinkListeners() {
+      var i,
+          ln,
+          links = _trackEvent.popcornTrackEvent._container.querySelectorAll( "a" );
+
+      if ( links ) {
+        for ( i = 0, ln = links.length; i < ln; i++ ) {
+          links[ i ].addEventListener( "click", linkClick, false );
```
ScottDowne: instead of using click events with prevent default, let's go with onclick returns false.

Using preventdefault does more than just stop the click. It does other, unexpected things.

If we return false inside the onclick, it'll return and cancel the click.

I think you have to use onclick, and not addeventlistener, like so:

elem.onclick = function() { return false; };

templates/assets/editors/image/image-editor.js
```diff
@@ -423,6 +449,7 @@
           _trackEvent = e.target;
           calcImageTime();
           _this.updatePropertiesFromManifest( _trackEvent );
+          addLinkListeners();
```
ScottDowne: Why are you creating a function that is only called in one spot?

Let's reduce the call stack and just pu the code in here.

PR #1117, opened 2012-10-11
[#2460] Disable preview button if you have unsaved changes. This way you...

ScottDowne:
... cannot preview. One disabled state for buttons. Changing title opens share and saves.

### DIFF
--- a/css/buttons.less
+++ b/css/buttons.less
@@ -37,6 +37,12 @@
   }
 }
 
+
+.butter-btn.butter-disabled {
+  opacity: 0.5;
+  cursor: not-allowed;
+}
+
 // Colors
 .btn-light {
   background: @baseLight;
@@ -60,12 +66,6 @@
   box-shadow: inset 0 1px 0 lighten( @green, 10% );
   &:hover {
     color: #FFF;
-  }
-
-  &.disabled {
-    color: #EEE;
-    background: @greyGreen;
-    border: none;
   }
 }
 
--- a/src/editor/share-editor.js
+++ b/src/editor/share-editor.js
@@ -79,6 +79,22 @@ define([ "editor/editor", "editor/base-editor", "ui/user-data",
       " frameborder='0' mozallowfullscreen webkitallowfullscreen allowfullscreen></iframe>";
     }
 
+    function togglePreviewButton( on ) {
+      if ( on ) {
+        previewBtn.classList.remove( "butter-disabled" );
+        previewBtn.href = butter.project.publishUrl;
+        previewBtn.onclick = function() {
+          return true;
+        };
+      } else {
+        previewBtn.classList.add( "butter-disabled" );
+        previewBtn.href = "";
+        previewBtn.onclick = function() {
+          return false;
+        };
+      }
+    }
+
     function displayEditor() {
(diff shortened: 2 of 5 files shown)


css/buttons.less
```diff
@@ -37,6 +38,12 @@
   }
 }
 
+
+.butter-btn.butter-disabled {
+  opacity: 0.5;
+  cursor: not-allowed;
+}
```
ScottDowne, line 44: One disabled state to rule them all.

css/buttons.less
```diff
@@ -61,12 +68,6 @@
   &:hover {
     color: #FFF;
   }
-
-  &.disabled {
-    color: #EEE;
-    background: @greyGreen;
-    border: none;
-  }
```
ScottDowne, line 69: Removing second disabled state, now just have one for all buttons to use.

src/editor/share-editor.js
```diff
@@ -100,10 +115,7 @@ define([ "editor/editor", "editor/base-editor", "ui/user-data",
 
       projectName.value = project.name;
       projectURL.value = project.publishUrl;
-      previewBtn.href = project.publishUrl;
-
-      headerPreviewBtn.classList.remove( "butter-hidden" );
-      headerPreviewBtn.href = project.publishUrl;
```
ScottDowne, line 106: Removing all references to header preview inside share editor. Header  code handles those buttons now.

src/ui/header.js
```diff
@@ -139,6 +151,7 @@ define([ "dialog/dialog", "util/lang", "ui/user-data", "ui/widget/tooltip" ],
       _projectName.textContent = node.value || _projectTitlePlaceHolderText;
       if( checkProjectName( _projectName.textContent ) ) {
         butter.project.name = _projectName.textContent;
+        login( prepare );
```
ScottDowne, line 154: This makes sure our share editor is opened after you change the title in the header, so you do not get confused when you open the share editor and it is asking for the title again.

src/editor/share-editor.js
```diff
@@ -79,6 +79,22 @@ define([ "editor/editor", "editor/base-editor", "ui/user-data",
       " frameborder='0' mozallowfullscreen webkitallowfullscreen allowfullscreen></iframe>";
     }
 
+    function togglePreviewButton( on ) {
+      if ( on ) {
+        previewBtn.classList.remove( "butter-disabled" );
+        previewBtn.href = butter.project.publishUrl;
+        previewBtn.onclick = function() {
+          return true;
```
cadecairos, line 87: is this line necessary? includes the `return false` below too

src/ui/header.js
```diff
@@ -61,12 +61,24 @@ define([ "dialog/dialog", "util/lang", "ui/user-data", "ui/widget/tooltip" ],
...
-        _saveButton.classList.add( "btn-light" );
+      }
+    }
+
+    function togglePreviewButton( on ) {
+      if ( on ) {
+        _previewBtn.classList.remove( "butter-disabled" );
+        _previewBtn.href = butter.project.publishUrl;
+        _previewBtn.onclick = function() {
+          return true;
```
cadecairos, line 74: same question as above

PR #1124, opened 2012-10-11
T2417

secretrobotron:


src/core/eventmanager.js
```diff
@@ -243,9 +243,13 @@ define( [], function(){
         }
       }
 
+      if ( !listener ) {
```
humphd, line 246: Move this above the `if ( listener )` check at 239, then get rid of that check totally, since you'll know listeners is non-null.

PR #1126, opened 2012-10-12
Bug 2297 - Fixed preview button to work on import

jbuck:


src/core/project.js
```diff
@@ -197,6 +213,14 @@ define( [ 'core/eventmanager', 'core/media' ],
           }
         }
       }
+
+      // If we're loading data from the server, then save it again so stuff works
+      // This is a bad hack and I should feel bad
+      if ( json.projectID ) {
+        _this.save(function(e) {
```
humphd: This empty callback is fine, but don't leave it open (empty space) like that, be clear that it's meant to be an empty func.

PR #1127, opened 2012-10-12
T2319

secretrobotron:


cornfield/lib/filter.js
```diff
@@ -8,7 +8,8 @@ filters = {
       next();
     } else {
       res.json({
-        error: 'unauthorized'
+        error: 'unauthorized',
+        csrf: req.session._csrf
```
jbuck, line 12: I don't think we want to send csrf tokens to unauthorized users... but now that I think about it, this call cause logging in to not work, right?

jbuck:
Closing in favour of #1133

PR #1133, opened 2012-10-12
Add CSRF protection

jbuck:


src/util/xhr.js
```diff
@@ -20,21 +20,24 @@
...
+          try {
+            __csrf_token = JSON.parse( this.response || this.responseText ).csrf;
+          } catch (e) {}
+        }
 
-    if ( document.readyState !== "loading" ) {
-      setCSRFToken();
-    } else {
-      document.addEventListener( "DOMContentLoaded", setCSRFToken, false );
+        callback.apply( this, arguments );
```
mjschranz: might want to ensure callback is a function here.

src/util/xhr.js
```diff
@@ -20,21 +20,24 @@
     return s.join("&").replace("/%20/g", "+");
   }
 
+  var __csrf_token = "";
```
secretrobotron: Silly note: why is this `__csrf_token`? CamelCase it.

src/util/xhr.js
secretrobotron: Not a bad idea. Callback should always be something here though. If it's not, it's most likely a mistake we'd want to hear about.

PR #1134, opened 2012-10-12
T2337

secretrobotron:


cornfield/lib/user.js
```diff
@@ -1,11 +1,23 @@
 "use strict";
 
+function defaultDBReadyFunction( err ) {
+  if ( err ) {
+    err = Array.isArray( err ) ? err[ 0 ] : err;
+    console.warn( "lib/user.js: DB setup error\n", err.number, err.sqlState, err.message );
```
mjschranz: when are some of these properties provided? When I tried locally to get this to popup number and sqlState were always undefined.

cornfield/test/lib.user.test.js
```diff
@@ -13,7 +13,26 @@ var test = require( "tap" ).test,
     id,
     callback;
 
-test( "db setup", function( t ) {
+test( "sqlite db setup", function( t ) {
```
mjschranz: Might be better to call this something that shows it won't fail with options that aren't ment for sqlite.

README.md
```diff
@@ -185,7 +199,15 @@ Testing
...
-* Run `node make server` and navigate to http://localhost:8888/test to run the browser tests
+* Run `node make server` and navigate to `http://localhost:8888/test` to run the browser tests
+
+If you are contributing changes to cornfield, make sure you run the cornfield tests from the butter root directory this command:
+
+```
+npm test
+```
+
+Cornfield will attempt to run tests for using a mysql backend, but they will not run if mysql is not set up properly. To ensure that they run, use the `DB_HOST`, `DB_DATABASE`, `DB_USERNAME`, and `DB_PASSWORD` enrionment variables to tell the tests which host, database, username, and password to use respectively. If `DB_HOST` is left blank, `localhost` is assumed.
```
mjschranz: <code>enrionment</code> -> <code>environment</code>

PR #1146, opened 2012-10-15
[#2482] Guard loading of autosaves from reading in null values from loca...

cadecairos:
...lStorage

Lighthouse Ticket: https://webmademovies.lighthouseapp.com/projects/65733/tickets/2482-local-saving-can-throw-on-page-load-if-there-is-a-key-with-no-data

### DIFF
--- a/src/core/project.js
+++ b/src/core/project.js
@@ -309,19 +309,18 @@ define( [ 'core/eventmanager', 'core/media' ],
     var obj = __butterStorage.getItem( "butter-backup-project" ),
         p;
 
-    if ( !obj ){
-      callback();
-      return;
-    }
-
-    if ( butter.config.value( "recover" ) !== "purge" ) {
-      p = new Project( butter );
-      p.import( obj );
-    }
+    try {
+      obj = JSON.parse( obj );
+      if ( obj && butter.config.value( "recover" ) !== "purge" ) {
+        p = new Project( butter );
+        p.import( obj );
+      }
+    } catch( e ) { }
 
     // Delete since user can save if he/she wants.
     __butterStorage.removeItem( "butter-backup-project" );
     callback( p );
+
   };
 
   return Project;
--- a/src/main.js
+++ b/src/main.js
@@ -791,9 +791,11 @@
                 // (i.e., we won't resotre again after this).
                 Project.checkForBackup( _this, function( project ) {
                   function useProject( project ) {
-                    project.template = project.template || _config.value( "name" );
-                    _this.project = project;
-                    _this.chain( project, [ "projectchanged", "projectsaved" ] );
+                    if ( project ) {
+                      project.template = project.template || _config.value( "name" );
+                      _this.project = project;
+                      _this.chain( project, [ "projectchanged", "projectsaved" ] );
+                    }
 
                     // Fire the ready event
                     _this.dispatch( "ready", _this );


src/core/project.js
```diff
@@ -309,19 +309,29 @@ define( [ 'core/eventmanager', 'core/media' ],
     var obj = __butterStorage.getItem( "butter-backup-project" ),
         p;
 
-    if ( !obj ){
-      callback();
+    function finish() {
+      // Delete since user can save if he/she wants.
+      __butterStorage.removeItem( "butter-backup-project" );
+      callback( p );
       return;
```
humphd: You don't need this return

src/core/project.js
```diff
@@ -309,19 +309,29 @@ define( [ 'core/eventmanager', 'core/media' ],
...
     }
 
-    if ( butter.config.value( "recover" ) !== "purge" ) {
-      p = new Project( butter );
-      p.import( obj );
+    try {
+      obj = JSON.parse( obj );
+
+      if ( !obj ) {
```
humphd: The logic here is wrong.  In all code paths, you call finish, so move it down below the try/catch block, and don't bother calling it inside try/catch.  Change this if check to check if obj does exist, and if so, do the block at 326:

``` javascript
if ( obj && butter.config.value( "recover" ) !== "purge" ) {
  p = new Project( butter );
  p.import( obj );
}
```

src/core/project.js
```diff
@@ -309,19 +309,29 @@ define( [ 'core/eventmanager', 'core/media' ],
...
+      if ( butter.config.value( "recover" ) !== "purge" ) {
+        p = new Project( butter );
+        p.import( obj );
+        finish();
+      }
+    } catch( e ) {
+      finish();
     }
 
-    // Delete since user can save if he/she wants.
```
humphd: I actually don't know if the move to the finish function gains you much here, since you need to do things no matter what at this point.

src/main.js
```diff
@@ -788,9 +788,11 @@
                 // (i.e., we won't resotre again after this).
                 Project.checkForBackup( _this, function( project ) {
                   function useProject( project ) {
-                    project.template = project.template || _config.value( "name" );
-                    _this.project = project;
-                    _this.chain( project, [ "projectchanged", "projectsaved" ] );
+                    if ( project ) {
```
humphd, line 794: I'm confused about this.  We have to have some kind of project at this point, no?  Won't we leave the app in a state of chaos if there is no project created (even if empty) before we dispatch ready?

cadecairos, line 794: In short, no.

Byt this point, Butter has loaded in it's default saved-data project. If the conversion of the localStorage value fails, `checkForBackup` will return undefined, so we know to not try and load in the recovered project.

src/core/project.js
```diff
@@ -309,19 +309,17 @@ define( [ 'core/eventmanager', 'core/media' ],
...
-    }
+    try {
+      obj = JSON.parse( obj );
+      if ( obj && butter.config.value( "recover" ) !== "purge" ) {
+        p = new Project( butter );
+        p.import( obj );
+      }
+    } catch( e ) { }
 
-    // Delete since user can save if he/she wants.
```
humphd: Restore these 2 lines (comment + removeItem), and have callback(p) go last.

PR #1155, opened 2012-10-15
[t2434] WIP - using transform: translate

secretrobotron:


src/util/dragndrop.js
```diff
@@ -35,17 +39,39 @@ define( [ 'core/eventmanager' ], function( EventManager ) {
...
+  function __getScrollGroup( scrollElement ) {
+    var i, newScrollGroup;
+
+    if ( scrollElement ) {
+      for ( i = __scrollGroups.length - 1; i >= 0; --i ) {
+        if ( __scrollGroups[ i ].scrollElement === scrollElement ) {
+          return __scrollGroups[ i ];
+        }
+      }
+      newScrollGroup = new ScrollGroup.ScrollGroup( scrollElement );
```
ScottDowne, line 53: new ScrollGroup.ScrollGroup?

That doesn't look right.

secretrobotron, line 53: No? What should it be?

templates/basic/saved-data.json
```diff
@@ -70,33 +70,6 @@
...
-                  "italics": false,
-                  "underline": false
-                },
-                "left": 25,
-                "top": 0,
-                "width": 50,
-                "zindex": 1000
-              },
-              "track": "Layer",
-              "name": "TrackEvent4"
```
ScottDowne: Is there a reason you deleted this?

jbuck: Yeah, I don't see the reason for deleting this. Please re-add

src/util/lang.js
```diff
@@ -156,6 +156,23 @@ define( [ "./shims" ], function(){
       element.removeEventListener( "transitionend", listener, false );
       element.removeEventListener( "oTransitionEnd", listener, false );
       element.removeEventListener( "webkitTransitionEnd", listener, false );
+    },
+
+    setTransformProperty: function( element, transform ) {
```
jbuck, line 174: I wonder if it wouldn't be better to figure out what prefix to use ahead of time. MDN suggestions a solution if you're interested: https://developer.mozilla.org/en-US/docs/CSS/CSS_animations/Detecting_CSS_animation_support

PR #1159, opened 2012-10-16
[#2536] Adjust draggable and resizeable z-index value

cadecairos:
Lighthouse Ticket: https://webmademovies.lighthouseapp.com/projects/65733/tickets/2536-popup-doesnt-have-re-positioning-handle

### DIFF
--- a/templates/assets/css/jquery-ui/jquery.ui.butter.less
+++ b/templates/assets/css/jquery-ui/jquery.ui.butter.less
@@ -24,12 +24,13 @@
 }
 
 
-@highlight: #3FB58E; //3FB58E;
+@highlight: #3FB58E;
 @handleWidth: 4px;
 @offset: 0;
 @handlesmall: 8px;
 @handlecorner: 18px;
 @handlelarge: 25px;
+@handleZIndex: 1000;
 
 .ui-resizable,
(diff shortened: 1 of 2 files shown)


templates/assets/css/jquery-ui/jquery.ui.butter.less
```diff
@@ -24,12 +24,13 @@
...
 
 
-@highlight: #3FB58E; //3FB58E;
+@highlight: #3FB58E;
 @handleWidth: 4px;
 @offset: 0px;
 @handlesmall: 8px;
 @handlecorner: 18px;
 @handlelarge: 25px;
+@zIndex: 5;
```
k88hudson: Because this is the same stack as all the plugin content, in this case it's actually a good idea to use an arbitrarily high number, like `1000`.

Also, consider naming this `@handleZIndex`, since that's what it is.

templates/assets/css/jquery-ui/jquery.ui.butter.less
```diff
@@ -24,12 +24,13 @@
 }
 
 
-@highlight: #3FB58E; //3FB58E;
```
k88hudson, line 27: Thanks, btw

PR #1164, opened 2012-10-16
[#2444] Have webmaker nav login menu display/close based around mouse events

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/2444-closing-the-user-data-dropdown-in-the-header-requires-clicking-on-the-users-email-again

### DIFF
--- a/src/ui/webmakernav/webmakernav.js
+++ b/src/ui/webmakernav/webmakernav.js
@@ -10,7 +10,9 @@ define( [ "util/lang", "text!./webmakernav.html", "text!./webmakernav.css" ],
       // Added to body when secondary nav is expanded
       EXPANDED_CLASS = "webmaker-expanded",
        // The class prefix for each individual tab
-      TAB_PREFIX = "tab-";
+      TAB_PREFIX = "tab-",
+      // Transition used for the user menu dropdown
+      USER_MENU_TRANSITION = "tooltip-no-transition-on";
 
   return function( options ) {
     options = options || {};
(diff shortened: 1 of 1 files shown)


src/ui/webmakernav/webmakernav.js
```diff
@@ -36,6 +36,14 @@ define( [ "util/lang", "text!./webmakernav.html", "text!./webmakernav.css" ],
         personaBtnGroup.style.display = "none";
         usernameContainer.style.display = "";
         usernameInner.innerHTML = butter.cornfield.username();
+
+        if ( userMenu.classList.contains( "tooltip-no-transition-on" ) ) {
```
secretrobotron: Can you make `tooltip-no-transition-on` a const? Kate did a nice job of organizing this file that way.

src/ui/webmakernav/webmakernav.js
```diff
@@ -36,6 +36,14 @@ define( [ "util/lang", "text!./webmakernav.html", "text!./webmakernav.css" ],
...
+          userMenu.classList.remove( "tooltip-no-transition-on" );
+        }
+
+        if ( username.classList.contains( BTN_ACTIVE_CLASS ) ) {
```
secretrobotron: When does this code run when `BTN_ACTIVE_CLASS` is actually on the `classList` of `username` (and same with `userMenu` and `tooltip-no-transition-on` above)? Do you really need these checks?

src/ui/webmakernav/webmakernav.js
```diff
@@ -91,8 +91,13 @@ define( [ "util/lang", "text!./webmakernav.html", "text!./webmakernav.css" ],
         e.stopPropagation();
       }, false );
       
-      username.addEventListener( "click", function() {
-        userMenu.classList.toggle( "tooltip-no-transition-on" );
+      username.addEventListener( "mouseout", function() {
+        userMenu.classList.toggle( USER_MENU_TRANSITION );
```
k88hudson: I think it's better to use add/remove -- if the `mouseout` event ever doesn't get fired, you'll end up with the reverse functionality, right?

PR #1166, opened 2012-10-16
[#2511] Share editor wasn't listening for a login event

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/2511-share-editor-does-not-respond-to-saved-project-state

### DIFF
--- a/src/editor/share-editor.js
+++ b/src/editor/share-editor.js
@@ -208,8 +208,17 @@ define([ "editor/editor", "editor/base-editor",
       }
     }, false );
 
+    function checkProjectState() {
+      if ( !butter.project.name || !butter.project.id ) {
+        displaySave();
+      } else {
+        displayEditor();
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


src/editor/share-editor.js
```diff
@@ -210,6 +210,13 @@ define([ "editor/editor", "editor/base-editor",
 
     butter.listen( "logout", displayLogin );
     butter.listen( "projectupdated", login );
+    butter.listen( "authenticated", function loggingIn(){
```
k88hudson: spacing

k88hudson: Also, `displayLogin` v.s. `loggingIn` are pretty misleading,   Could we maybe name them something more appropriate for what they are and group those functions together?

mjschranz: You've been just patiently waiting to get me on a spacing mistake haven't you?

PR #1171, opened 2012-10-16
[#2510] Prevent save button listeners from being active all the time

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/2510-save-button-is-grey-when-loading-a-remix-project-but-is-valid-and-clickable

### DIFF
--- a/src/ui/header.js
+++ b/src/ui/header.js
@@ -36,11 +36,29 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
       document.body.classList.toggle( "tabzilla-open" );
     }, false );
 
+    function saveProject( e ) {
+      if ( !butter.cornfield.authenticated() ) {
+        _userData.authenticationRequired();
+      }
+      else if ( butter.project.isSaved ) {
+        return;
+      }
+      else if ( checkProjectName( butter.project.name ) ) {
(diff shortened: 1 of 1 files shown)


src/ui/header.js
```diff
@@ -70,11 +88,25 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
...
+
         _previewBtn.style.display = "";
         _projectTitle.style.display = "";
         _saveButton.innerHTML = "Save";
+
+        if ( !butter.project.id ) {
+          saveState = true;
+        }
+
+        // Preview button isn't display if someone loads a dashboard project then logs out
```
ScottDowne: Shouldn't that be "isn't displayed"?

src/ui/header.js
```diff
@@ -70,11 +88,25 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
...
+        // and back in.
+        if ( butter.project.isSaved ) {
+          previewState = true;
+        }
+
+        togglePreviewButton( previewState );
```
ScottDowne: Why not

togglePreviewButton( butter.project.isSaved ):
toggleSaveState( !butter.project.id );

PR #1173, opened 2012-10-16
[t2496] webmakernav require entry-point

secretrobotron:


src/ui/webmakernav/webmakernav.less
```diff
@@ -1,4 +1,5 @@
 @import "../../../css/globals";
+@import "../../../css/editors";
```
mjschranz, line 2: Why does this need to be imported?

PR #1174, opened 2012-10-16
[#2551] Fixed popup so it wasn't getting double sanitized

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733-popcorn-maker/tickets/2551-escaped-characters-are-still-appearing-on-export

### DIFF
--- a/templates/assets/plugins/popup/popcorn.popup.js
+++ b/templates/assets/plugins/popup/popcorn.popup.js
@@ -225,19 +225,17 @@
           width = options.width + "%",
           top = options.top + "%",
           left = options.left + "%",
-          i,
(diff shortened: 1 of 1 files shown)


templates/assets/plugins/popup/popcorn.popup.js
```diff
@@ -15,6 +15,11 @@
   innerDivTriangles.speech = document.createElement( "canvas" );
   innerDivTriangles.thought = document.createElement( "canvas" );
 
+  function newlineToBreak( string ) {
```
jbuck: Hm, if you're only using this function once, why not do it inline on line 245?

PR #1178, opened 2012-10-17
[#2521] Text plugin width resizable. Text plugin uses percentages for si...

ScottDowne:
...zes.

### DIFF
--- a/css/embed-shell.less
+++ b/css/embed-shell.less
@@ -16,7 +16,7 @@ iframe {
 }
 
 
-@media only screen and ( max-width: @medium ) {
+@media only screen and ( max-width: @small ) {
   iframe {
     .height-width( @small );
   }
--- a/css/embed-shell.less
+++ b/css/embed-shell.less
@@ -16,7 +16,7 @@ iframe {
 }
 
 
-@media only screen and ( max-width: @small ) {
+@media only screen and ( max-width: @medium ) {
   iframe {
     .height-width( @small );
   }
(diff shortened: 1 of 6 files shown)


templates/assets/plugins/text/popcorn.text.css
```diff
@@ -76,13 +78,68 @@
...
 .text-custom {
   display: inline-block;
-  max-width: 500px;
 }
 
 .text-custom > div > span {
-  padding: 0;
+  display: block;
+  font-size: 1em;
+  width: 100%;
```
mjschranz, line 89: I don't know if I like the default being 100% personally. It felt like... "a lot" when switching to custom when I saw the dragging handles.

templates/assets/editors/editorhelper.js
```diff
@@ -60,6 +60,8 @@
           }
         },
         stop: function( event, ui ) {
+          media = mediaContainer.getBoundingClientRect();
```
mjschranz, line 67: Hmm. I guess it makes sense that you would want to wait until the dragging stops here.

ScottDowne, line 67: Are we good here?

I did this because the media size can change after initial creation of the editor.

By doing this, we assure we always have up to date sizes.

css/embed-shell.less
```diff
@@ -16,7 +16,7 @@ iframe {
 }
 
 
-@media only screen and ( max-width: @medium ) {
+@media only screen and ( max-width: @small ) {
```
secretrobotron: Why did you change this? It should be `medium` to cover the spectrum of sizes. Note that it says `max-width` and not `min-width` like the others.

PR #1180, opened 2012-10-17
[#2544] Make all API calls in plugins not specify a protocol

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/2544-cant-change-font-in-chrome

### DIFF
--- a/templates/assets/plugins/image/popcorn.image.js
+++ b/templates/assets/plugins/image/popcorn.image.js
@@ -12,6 +12,10 @@
       jsonBits = "&format=json&jsoncallback=flickr",
       FLICKR_SINGLE_CHECK = "flickr.com/photos/";
 
+  if ( window.location.protocol === "https:" ) {
+    flickrUrl = "https://secure.flickr.com/services/";
+  }
+
   function searchImagesFlickr( tags, count, userId, ready ) {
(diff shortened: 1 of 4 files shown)


templates/assets/plugins/image/popcorn.image.js
```diff
@@ -5,7 +5,7 @@
 (function ( Popcorn ) {
 
   var APIKEY = "&api_key=b939e5bd8aa696db965888a31b2f1964",
-      flickrUrl = "http://api.flickr.com/services/",
+      flickrUrl = "//api.flickr.com/services/",
```
jbuck: Uh oh, flickr doesn't seem to have an https version of their API

templates/assets/plugins/twitter/popcorn.twitter.js
```diff
@@ -234,7 +234,7 @@
       if ( !CACHED_RESULTS[ query ] ) {
         if ( options.username ) {
           Popcorn.xhr({
-            url: "https://api.twitter.com/1/account/rate_limit_status.json",
+            url: "//api.twitter.com/1/account/rate_limit_status.json",
```
jbuck, line 237: Heh, I like how https was used here, but http was used on the other requests.

PR #1181, opened 2012-10-17
Bug 2537 - Lint all CSS files in tree 

jbuck:


templates/assets/css/jquery-ui/jquery.ui.butter.less
```diff
@@ -41,7 +41,7 @@
 /*********************************************************
 * Resizable
 */
-@resizeSmall: 10px;
+@resizeSmall: 10;
```
k88hudson: This is not right

templates/assets/css/jquery-ui/jquery.ui.butter.less
```diff
@@ -68,15 +68,15 @@
 .ui-draggable .title {
   position: absolute;
   background: #FFF;
-  bottom: -30px;
+  bottom: -30;
```
k88hudson: Why are you removing all the units?

PR #1182, opened 2012-10-17
[#2547] Remove escape key close function in the first-run dialog

cadecairos:
Lighthouse: https://webmademovies.lighthouseapp.com/projects/65733/tickets/2547-crash-httpdevbutyrorgsrcbutterjs9872-typeerror-_modal-is-null

### DIFF
--- a/src/dialog/dialogs/first-run.js
+++ b/src/dialog/dialogs/first-run.js
@@ -1,72 +1,10 @@
 /* This Source Code Form is subject to the terms of the MIT license
- *  * If a copy of the MIT license was not distributed with this file, you can
- *  * obtain one at http://www.mozillapopcorn.org/butter-license.txt */
+ * If a copy of the MIT license was not distributed with this file, you can
+ * obtain one at http://www.mozillapopcorn.org/butter-license.txt */
 
-define( [ "text!dialog/dialogs/first-run.html", "dialog/dialog",
-          "ui/widget/tooltip", "util/shims" ],
-  function( LAYOUT_SRC, Dialog, ToolTip ) {
-
-
-    var TOOLTIP_DELAY = 3000;
-
-    Dialog.register( "first-run", LAYOUT_SRC, function ( dialog, data ) {
-
-      var editor = document.querySelector( ".butter-editor-area" ),
-          mediaEditorButton = document.querySelector( ".butter-editor-header-media" ),
-          popupTile = document.querySelector( ".butter-plugin-tile[data-popcorn-plugin-type=popup]" ),
-          overlayDiv = document.createElement( "div" ),
-          popupTooltip,
-          mediaTooltip;
-
-      overlayDiv.classList.add( "butter-modal-overlay" );
-      overlayDiv.classList.add( "butter-modal-overlay-dark-bg" );
-      overlayDiv.classList.add( "fade-in" );
-
-      editor.appendChild( overlayDiv );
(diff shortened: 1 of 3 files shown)


src/dialog/dialogs/first-run.js
```diff
@@ -2,71 +2,11 @@
  *  * If a copy of the MIT license was not distributed with this file, you can
  *  * obtain one at http://www.mozillapopcorn.org/butter-license.txt */
```
secretrobotron: Mind cleaning this up?

src/dialog/dialogs/first-run.js
```diff
@@ -2,71 +2,11 @@
...
 
-define( [ "text!dialog/dialogs/first-run.html", "dialog/dialog",
-          "ui/widget/tooltip", "util/shims" ],
-  function( LAYOUT_SRC, Dialog, ToolTip ) {
+define( [ "text!dialog/dialogs/first-run.html", "dialog/dialog", "util/shims" ],
+  function( LAYOUT_SRC, Dialog ) {
```
secretrobotron: And this?

src/first-run.js
```diff
@@ -7,16 +7,91 @@
...
  */
-define( [ "dialog/dialog", "util/cookie" ], function( Dialog, Cookie ) {
+define( [ "dialog/dialog", "util/cookie", "ui/widget/tooltip" ], function( Dialog, Cookie, ToolTip ) {
   return {
     init: function( config ) {
 
-      var _dialog;
+      var TOOLTIP_DELAY = 3000;
+
+      var _dialog,
```
secretrobotron: Let's just make this `dialog` instead of `_dialog` for consistency's sake.

src/dialog/dialogs/first-run.js
```diff
@@ -2,71 +2,11 @@
  *  * If a copy of the MIT license was not distributed with this file, you can
  *  * obtain one at http://www.mozillapopcorn.org/butter-license.txt */
 
-define( [ "text!dialog/dialogs/first-run.html", "dialog/dialog",
-          "ui/widget/tooltip", "util/shims" ],
-  function( LAYOUT_SRC, Dialog, ToolTip ) {
+define( [ "text!dialog/dialogs/first-run.html", "dialog/dialog", "util/shims" ],
```
secretrobotron: Does `shims` need to be here anymore?

src/first-run.js
```diff
@@ -7,16 +7,91 @@
...
+            popupTooltip.parentNode.removeChild( popupTooltip );
+          }
+        }, TOOLTIP_DELAY );
+      }
+
+
+      function onDialogClose() {
+        // Remove Listeners
+        dialog.unlisten( "close", onDialogClose );
+        window.removeEventListener( "click", closeDialog, false );
```
humphd, line 52: Move this into closeDialog so it happens right after you click.

src/first-run.js
```diff
@@ -7,16 +7,91 @@
...
+        popupTile.removeEventListener( "mouseover", closeDialog, false );
```
humphd, line 53: Same here.

src/first-run.js
cadecairos, line 52: The problem with moving the two lines here is that if they're in closeDialog, when the dialog closes itself via the escape key, closeDialog is never called and the listeners won't be removed

PR #1189, opened 2012-10-17
[t2509] export sortedtracks instead of unsorted

secretrobotron:


src/core/media.js
```diff
@@ -494,8 +494,8 @@
         json: {
           get: function(){
             var exportJSONTracks = [];
-            for( var i=0, l=_tracks.length; i<l; ++i ){
-              exportJSONTracks.push( _tracks[ i ].json );
+            for( var i=0, l=_orderedTracks.length; i<l; ++i ){
```
dseif: `for( var i=0, l=_orderedTracks.length; i<l; ++i ){` -> `for ( var i = 0, l = _orderedTracks.length; i < l; ++i ) {`

PR #1206, opened 2012-10-18
T2595

secretrobotron:


src/cornfield/module.js
```diff
@@ -24,10 +25,28 @@ define(['util/xhr'], function(XHR) {
...
+    var sendXHRPost = function() {
+      console.warn( "XHR.post occurred without a CSRF token. Buffering request." );
+      xhrPostQueue.push( arguments );
+    };
+
+    XHR.setCSRFTokenAcquiredCallback( function() {
+      var args;
+
+      while ( xhrPostQueue.length ) {
+        args = xhrPostQueue.pop();
```
jbuck: Change pop() to shift() so that POST requests get processed in the order they were added (Add to back, remove from front).

PR #1216, opened 2012-10-19
[t2613] attribution in embeds for stamen maps

secretrobotron:


css/embed.less
```diff
@@ -204,18 +204,22 @@ body {
 .attribution-author {
   font-style: italic;
 }
-.attribution-media {
+
+
```
k88hudson, line 208: could you maybe take this out

PR #1225, opened 2012-10-19
T2533

cadecairos:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/2533-header-is-too-tall-in-opera

### DIFF
--- a/css/editors.less
+++ b/css/editors.less
@@ -25,10 +25,11 @@
     padding: 0;
     width: 100%;
     list-style: none;
-    .clearfix();
 
-    > li {
+    > .butter-editor-header-li {
       float: left;
+      position: relative;
+      top: 0;
     }
   }
 } // .butter-editor-header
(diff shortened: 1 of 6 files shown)


cadecairos:
I messed this up somehow...

cadecairos:
There we go!

css/editors.less
```diff
@@ -25,10 +25,11 @@
     padding: 0;
     width: 100%;
     list-style: none;
-    .clearfix();
 
-    > li {
+    > li.butter-editor-header-li {
```
k88hudson: you can actually remove the `li` from this, i.e just  `.butter-editor-header-li`
also, just wondering -- what was the reason for this change?

css/normalize.less
```diff
@@ -269,4 +274,12 @@ td {
 
 iframe {
   border: 0;
-}
\ No newline at end of file
+}
```
k88hudson: Remove one of these body declarations, https://github.com/mozilla/butter/pull/1225/files#L2R3

css/editors.less
cadecairos: to avoid the styles being applied to the last `<li>` in the group, which is used for hiding and showing the editor.

PR #1229, opened 2012-10-22
[#2629] Fix play/pause button state when switching media

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/2629-switching-media-sources-while-playing-yields-incorrect-timeline-state

### DIFF
--- a/src/timeline/status.js
+++ b/src/timeline/status.js
@@ -111,6 +111,9 @@ define( [ "util/lang" ], function( util ){
       }
     });
 
+    // Ensure default state is correct
+    _playButton.state = true;
+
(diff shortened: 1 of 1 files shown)


src/timeline/status.js
```diff
@@ -135,6 +138,10 @@ define( [ "util/lang" ], function( util ){
       _playButton.state = true;
     });
 
+    _media.listen( "mediacontentchanged", function( e ){
```
cadecairos, line 141: space before the `{`

mjschranz, line 141: Sorry but the rest of the file is like that. I stuck with that to make it consistent.

PR #1232, opened 2012-10-22
T2627

cadecairos:
Lighthouse Ticket: https://webmademovies.lighthouseapp.com/projects/65733-popcorn-maker/tickets/2627

### DIFF
--- a/src/dependencies.js
+++ b/src/dependencies.js
@@ -61,13 +61,19 @@ define([ 'util/xhr' ], function( XHR ){
       // CSS Loader
       css: function( url, exclude, callback, checkFn, error ){
         var link,
-            interval;
+            interval,
+            img,
+            alreadyFired = false;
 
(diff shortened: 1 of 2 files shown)


src/dependencies.js
```diff
@@ -61,13 +61,18 @@ define([ 'util/xhr' ], function( XHR ){
...
-            interval;
+            interval,
+            img;
 
         checkFn = checkFn || function(){
           return !!link;
         };
 
         function runCheckFn() {
+          if ( runCheckFn.alreadyFired ) {
```
secretrobotron: Just curious: why do you attach `alreadyFired` to `runCheckFn` instead of just making it a flag variable inside this loader function?

cadecairos: There's no real reason why. I can make it a variable to save the extra lookup though.

PR #1235, opened 2012-10-22
[#2594] Default text plugin to use custom position.

ScottDowne:


mjschranz:
If this is going to land( IRC discussions made me think not ) then you are going to need to update the JSON files we have in tree at least because otherwise when opening editors for them they won't have anything selected for the position drop down: https://github.com/ScottDowne/butter/blob/9f0473313482df6408814a09f8c92aa86cad2890/templates/basic/saved-data.json#L56

mjschranz:
To get around the disconnect between the options we store on a popcorn track event object and what Popcorn Maker stores for it's internals, you probably could check what it is inside the editor on setup and fire an update if it's using old values.

templates/assets/plugins/text/popcorn.text.js
```diff
@@ -39,10 +39,19 @@
...
+          "default": "custom"
+        },
+        alignment: {
+          elem: "select",
+          options: [ "Default", "Center", "Left", "Right" ],
+          values: [ "default", "center", "left", "right" ],
+          label: "Text Alignment",
+          // position custom defaults to alignment left,
+          // position fixed defaults to alignment center
+          "default": "Default"
```
mjschranz: the value should the corresponding index in values, IE <code>default</code>

PR #1239, opened 2012-10-22
[t2617] Changed font system to use ems, added global media query mixin

k88hudson:


css/globals.less
```diff
@@ -203,3 +203,43 @@
...
+
+// Heights are calculated based on 16:9 standard video aspect ratio.
+@aspectratio: 0.5625;
+
+// Editor width
+@in-editor-offset-w: @EDITOR_WIDTH;
+@in-editor-offset-h: @TRAY_HEIGHT + @HEADER_SPACING;
+
+@small: ~`"only screen and (min-width : 0px) and (min-width: 0px)"`;
+@medium: ~`"only screen and (min-width : " + parseInt("@{medium-w}") + "px) and (min-height : " + (parseInt("@{medium-w}") * @{aspectratio}) + "px)"`;
```
k88hudson: ~`...` is javascript evaluation in LESS http://lesscss.org/#-javascript-evaluation

have to use `parseInt` because the LESS variables have `px` in them

css/globals.less
```diff
@@ -203,3 +203,43 @@
...
+@xlarge-w: 1280px;
+
+// Heights are calculated based on 16:9 standard video aspect ratio.
+@aspectratio: 0.5625;
+
+// Editor width
+@in-editor-offset-w: @EDITOR_WIDTH;
+@in-editor-offset-h: @TRAY_HEIGHT + @HEADER_SPACING;
+
+@small: ~`"only screen and (min-width : 0px) and (min-width: 0px)"`;
```
k88hudson: This can't actually be replaced by `(max-width: @medium-w) and (max-height: @medium-h)` because there would be a case where `width > @medium-w` (out of range) but `height < @medium-h` (in-range) or vice versa that isn't covered by any media query.

The solution is either to use `(min-width : 0px) and (min-width: 0px)`, or to have a base style that has no media query, i.e.

```
.class {
  width: 10px; // This is for the @small case
  @media @medium {
      width: 13px;
   ...
```

templates/basic/style.less
```diff
@@ -32,7 +59,32 @@ body.editor-minimized .body-wrapper {
 -------------------------------------------------- */
 .video-container {
   position: relative;
-  font-size: 16px;
+  @media @small {
```
k88hudson: Set the base font size on the container so all plugins can make use of it

templates/basic/style.less
```diff
@@ -32,7 +59,32 @@ body.editor-minimized .body-wrapper {
...
+    font-size: 14px;
+  }
+  @media @medium {
+    font-size: 16px;
+  }
+  @media @large {
+    font-size: 18.66px;
```
secretrobotron: Should these not also be `em` now?

templates/assets/plugins/popup/popcorn.popup.js
```diff
@@ -2,6 +2,7 @@
 
 (function ( Popcorn ) {
 
+
```
secretrobotron: Kindly remove.

templates/basic/style.less
k88hudson: No, each base font size is set in pixels, so that all the plugins set font sizes in `em`s relative to that base size.

For example, let's say a plugin sets the font size to `1.5`em.

At the smallest size:
- width: `560px`
- base font size is `14px`
- plugin's font size set by user: `1.5em`
- plugin's computed size: `14px`/1.5  =  `21px`
- plugin's computed size / width = 560/21 = _0.0375_

At medium size:
- width: `640px`
- base font size is `16px` ( medium width / small width  \* small base font size )
- plugin's font size set by user: `1.5em`
- plugin's computed size: `16px*1.5`  =  `24px`
- plugin's computed size / width = 640/24 = _0.0375_

This way, the ratio of the computed font size to width is always the same.

css/globals.less
```diff
@@ -203,3 +203,47 @@
...
+//      font-size: 16px;
+//    }
+//  }
+// }
+
+// These are the widths chosen for each supported size.
+@small-w: 560px;
+@medium-w: 640px;
+@large-w: 853px;
+@xlarge-w: 1280px;
```
ScottDowne: Why not have two variables, one with px and one without. Make the with one just be the witout one + "px"

That way, you do not need all those parseInts.

make.js
```diff
@@ -349,6 +349,12 @@ function buildCSS(compress) {
   });
 
   lessToCSS({
+    lessFile: "templates/basic/style.less",
+    cssFile: "templates/basic/style.css",
+    compress: compress
+  });
+
+  lessToCSS({
```
ScottDowne: Why have two lessToCSS blocks?

templates/assets/plugins/text/popcorn.text.js
```diff
@@ -11,7 +11,8 @@
 
    **/
 
-  var DEFAULT_FONT_COLOR = "#000";
+  var DEFAULT_FONT_COLOR = "#000",
+      FONT_FACTOR = 22.5;
```
ScottDowne: I really do not like this.

Let's explore ways to remove it?

templates/assets/plugins/popup/popcorn.popup.js
```diff
@@ -9,7 +9,8 @@
       _pluginRoot = "/templates/assets/plugins/popup/",
       FILL_STYLE = "rgb(255, 255, 255)",
       innerDivTriangles = {},
-      DEFAULT_FONT = "Tangerine";
+      DEFAULT_FONT = "Tangerine",
+      BASE_FONT_SIZE = 14;
```
ScottDowne: I really do not like this.

Let's explore ways to remove it?

css/globals.less
```diff
@@ -203,3 +203,47 @@
...
+// width > @medium-w` (out of range) but `height < @medium-h` (in-range) or vice versa that isn't covered by any media query.
+@medium: ~`"only screen and (min-width : " + parseInt("@{medium-w}") + "px) and (min-height : " + (parseInt("@{medium-w}") * @{aspectratio}) + "px)"`;
+@large: ~`"only screen and (min-width : " + parseInt("@{large-w}")+ "px) and (min-height : " + (parseInt("@{large-w}")* @{aspectratio}) + "px)"`;
+@xlarge: ~`"only screen and (min-width : " + parseInt("@{xlarge-w}") + "px) and (min-height : " + (parseInt("@{xlarge-w}") * @{aspectratio}) + "px)"`;
+
+@small-in-editor: ~`"only screen and (min-width : 0px) and (min-width: 0px)"`;
+@medium-in-editor: ~`"only screen and (min-width : " + (parseInt("@{medium-w}") + parseInt("@{in-editor-offset-w}") ) + "px) and (min-height : " + (parseInt("@{medium-w}") * @{aspectratio} + parseInt("@{in-editor-offset-h}")) + "px)"`;
+@large-in-editor: ~`"only screen and (min-width : " + (parseInt("@{large-w}")+ parseInt("@{in-editor-offset-w}") ) + "px) and (min-height : " + (parseInt("@{large-w}")* @{aspectratio} + parseInt("@{in-editor-offset-h}")) + "px)"`;
```
ScottDowne: also missing a whitespace between )+

make.js
k88hudson: How else would you do it? There are already like 8. 

If we want to have it run through an array of file names or something instead, that would probably be better, although not this ticket IMO.

css/globals.less
```diff
@@ -203,3 +203,50 @@
...
+@small-w-px: @small-w + 0px;
+@medium-w-px: @medium-w + 0px;
+@large-w-px: @large-w + 0px;
+@xlarge-w-px: @xlarge-w + 0px;
+
+// Heights are calculated based on 16:9 standard video aspect ratio.
+@aspectratio: 0.5625;
+
+// Editor width
+@in-editor-offset-w: ~`parseInt( "@{EDITOR_WIDTH}" )`;
```
k88hudson, line 237: The alternative here would be to make three new variables for all of these constants and it would be really awkward. Since this is only compiled once, it's not so bad if it's just in this one place, right?

make.js
ScottDowne: Ha, right.

For some reason I thought you could make one function call and pass everything in.

templates/assets/plugins/text/popcorn.text.js
```diff
@@ -11,7 +11,7 @@
 
    **/
 
-  var DEFAULT_FONT_COLOR = "#000000",
+  var DEFAULT_FONT_COLOR = "#000",
```
ScottDowne: Cannot revert this.

#000 does not work in chrome, has to be #000000.

It just so happens chrome defaults to black if the colour is invalid, so we never saw it.

Trying changing it to #FFF and see it go to black.

PR #1242, opened 2012-10-23
[#2588] Prevent loading spinner from endlessly spinning

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/2588-visual-can-keep-media-loading-spinner-in-media-editor


src/editor/media-editor.js
```diff
@@ -208,6 +208,14 @@ define( [ "util/lang", "util/keys", "editor/editor", "util/uri", "text!layouts/m
         _media.listen( "mediacontentchanged", onMediaContentChanged );
         _media.listen( "mediafailed", onMediaFailed );
 
+        // Ensure loading spinner is always off, assuming the current media is ready.
```
secretrobotron: How about "Ensure the loading spinner is off when the media is ready. Otherwise, keep it spinning."?

PR #1247, opened 2012-10-23
[t2669] state recording in crash reporter

secretrobotron:


src/dialog/dialogs/crash.js
```diff
@@ -9,6 +9,7 @@ define( [ "text!dialog/dialogs/crash.html", "dialog/dialog", "util/lang" ],
       return "<b>Date</b>: " + report.date + "<br>" +
              "<b>URL</b>: " + report.url + ":" + report.lineno + "<br>" +
              "<b>Error</b>: " + LangUtil.escapeHTML( report.message ) + "<br>" +
+             "<b>Butter State</b>: " + report.stateList.slice().reverse().join( ", " ) + "<br>" +
```
humphd, line 12: Simplify this, just do:

```
report.stateList.slice(-10).reverse().join(", ")
```

This will take the last 10 items (or fewer if less than 10), then you can drop all the list length state handling above.

src/crashreporter.js
```diff
@@ -9,15 +9,41 @@
  */
 define( [ "dialog/dialog", "util/xhr", "util/uri" ], function( Dialog, XHR, URI ) {
 
+  var STATE_EVENT_QUEUE_LENGTH = 10;
```
humphd, line 12: Drop this (see below)

src/crashreporter.js
```diff
@@ -9,15 +9,41 @@
...
   // Only deal with errors we cause (i.e., same origin scripts).
   function shouldHandle( url ) {
     var uriScript = URI.parse( url );
     return uriScript.host === window.location.hostname;
   }
 
+  // Record a handfull of recently-triggered state-changing events.
+  function butterStateChangedOnEvent( e ) {
+    __stateEventQueue.push( e.type );
+    if ( __stateEventQueue.length > STATE_EVENT_QUEUE_LENGTH ) {
```
humphd, line 35: Drop this if block, see below.

PR #1253, opened 2012-10-23
[#2639] Prevent urls from being treated relative to the server

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/2639-links-that-dont-include-protocol-are-relative

### DIFF
--- a/templates/assets/plugins/popup/popcorn.popup.js
+++ b/templates/assets/plugins/popup/popcorn.popup.js
@@ -104,7 +104,7 @@
         },
         linkUrl: {
           elem: "input",
-          type: "url",
+          type: "text",
           label: "Link URL"
         },
         type: {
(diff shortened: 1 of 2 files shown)


templates/assets/plugins/popup/popcorn.popup.js
```diff
@@ -353,9 +354,14 @@
       textContainer.style.fontSize = options.fontSize ? options.fontSize + "px" : "12px";
       textContainer.style.fontWeight = options.fontDecorations.bold ? "bold" : "normal";
 
-      if ( options.linkUrl ) {
+      if ( linkUrl ) {
+
+        if ( linkUrl.indexOf( "http://" ) === -1 && linkUrl.indexOf( "https://" ) === -1 ) {
```
cadecairos: you can achieve this with a single regex.

`/^http(|s):\/\//`

templates/assets/plugins/text/popcorn.text.js
```diff
@@ -212,9 +213,14 @@
           innerDiv.style.zIndex = +options.zindex;
         }
 
-        if ( options.linkUrl ) {
+        if ( linkUrl ) {
+
+          if ( linkUrl.indexOf( "http://" ) === -1 && linkUrl.indexOf( "https://" ) === -1 ) {
```
cadecairos: same regex as above plz

PR #1256, opened 2012-10-24
[#2505] Text shadow and background colour.

ScottDowne:


templates/assets/plugins/text/popcorn.text.js
```diff
@@ -180,8 +198,11 @@
       innerContainer.style.fontStyle = fontDecorations.italics ? "italic" : "normal";
       innerContainer.style.fontWeight = fontDecorations.bold ? "bold" : "normal";
 
+      if ( fontDecorations.background ) {
+        innerDiv.style.background = backgroundColor;
```
k88hudson: use `.backgroundColor`, this will apply extra settings like `background-position`, etc. that are unnecessary

src/editor/trackevent-editor.js
```diff
@@ -355,11 +355,14 @@ define([ "util/lang", "util/keys", "util/time", "./base-editor", "ui/widget/tool
      * @param {TrackEvent} trackEvent: TrackEvent to update
      * @param {String} propertyName: Name of property to update when change is detected
      */
-    extendObject.attachCheckboxChangeHandler = function( element, trackEvent, propertyName ) {
+    extendObject.attachCheckboxChangeHandler = function( element, trackEvent, propertyName, callBack ) {
+      callBack = callBack || function( trackEvent, prop, updateOptions ) {
+        trackEvent.update( updateOptions );
```
k88hudson, line 360: Just wondering but why do you actually have to put this here? Like why not just have `trackEvent.update` where it used to be and do `callBack = callBack || function() {}`

templates/assets/plugins/text/popcorn.text.js
```diff
@@ -180,8 +211,11 @@
       innerContainer.style.fontStyle = fontDecorations.italics ? "italic" : "normal";
       innerContainer.style.fontWeight = fontDecorations.bold ? "bold" : "normal";
 
-      if ( fontDecorations.shadow ) {
-        innerContainer.classList.add( "popcorn-text-shadow" );
+      if ( options.background ) {
```
k88hudson, line 215: Isn't kind of weird to check `options.background` but apply `backgroundColor`?

templates/assets/editors/text/text-editor.js
```diff
@@ -37,16 +38,32 @@
...
         }
+        function checkboxCallback( trackEvent, prop, updateOptions ) {
+          if ( "background shadow".match( prop ) ) {
+            if ( updateOptions[ prop ] ) {
+              pickers[ prop ].classList.remove( "butter-editor-disabled" );
+              pickers[ prop ].onclick = function() {
+                return true;
+              };
+            } else {
+              pickers[ prop ].classList.add( "butter-editor-disabled" );
```
k88hudson: You also want to add `pickers[ prop ].setAttribute( "disabled", "true" )` to actually disable editing, particularly in Firefox

src/editor/trackevent-editor.js
ScottDowne, line 360: That is setting a default callback to do what the old code used to do.

If you notice on line 362, I removed the old update call, and can now call callback in all cases. I remove a couple of checks with this.

templates/assets/plugins/text/popcorn.text.js
ScottDowne, line 215: Hm, not sure I follow.

Background is to say if you want a background or not. backgroundColour is what colour you want it.

k88hudson, line 215: Oh, right that makes sense.. haha

src/editor/trackevent-editor.js
k88hudson, line 360: I was just thinking leaving the update at line 362 instead of having it inside the callback would be better, since `colorCallback` is only handling the UI state for those other editor elements anyway. But this is fine ^_^

PR #1264, opened 2012-10-24
[#2685] Removed functionality for logging in/naming project inside share editor. Cleaned up remaining code.

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/2685-editing-project-name

### DIFF
--- a/src/editor/share-editor.js
+++ b/src/editor/share-editor.js
@@ -3,16 +3,12 @@
  * obtain one at https://raw.github.com/mozilla/butter/master/LICENSE */
 
 define([ "editor/editor", "editor/base-editor",
-          "text!layouts/share-editor.html", "util/social-media", "ui/widget/tooltip" ],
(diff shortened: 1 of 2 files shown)


src/editor/share-editor.js
```diff
@@ -111,83 +61,44 @@ define([ "editor/editor", "editor/base-editor",
...
-        projectName.value = butter.project.name;
-      }
-
-      butter.cornfield.login(function() {
-        if ( !butter.project.name ) {
-          displaySave();
+          // if any of the buttons haven't loaded, or if we aren't logged in
+          if ( !shareTwitter.childNodes.length ||
+               !shareGoogle.childNodes.length ||
+               !butter.cornfield.authenticated() ) {
```
jbuck: We're already logged in at this point, so I don't think we need this condition in the if

PR #1265, opened 2012-10-24
T2492

humphd:


src/core/project.js
```diff
@@ -310,31 +326,34 @@ define( [ 'core/eventmanager', 'core/media' ],
...
+  // Returns project backup data as JS object if found, otherwise null.
+  // NOTE: caller must create a new Project object and call import.
   Project.checkForBackup = function( butter, callback ) {
     // See if we already have a project autosaved from another session.
-    var obj = __butterStorage.getItem( "butter-backup-project" ),
-        p;
+    var projectBackup, backupDate;
+
+    // For testing purposes, we can skip backup recovery
+    if( butter.config.value( "recover" ) === "purge" ) {
```
secretrobotron: Add some spacing after `if`.

src/core/project.js
```diff
@@ -310,31 +326,34 @@ define( [ 'core/eventmanager', 'core/media' ],
...
+      callback( null, null);
```
secretrobotron: Add some spacing after `null`.

src/dialog/dialogs/backup.html
```diff
@@ -0,0 +1,16 @@
+<!-- This Source Code Form is subject to the terms of the MIT license
+If a copy of the MIT license was not distributed with this file, you can
+obtain one at http://www.mozillapopcorn.org/butter-license.txt -->
+
+<div class="butter-dialog">
+  <div class="butter-form">
+    <fieldset>
+      <h2>We found an automatic backup for "<span class="butter-backup-project-name"></span>"</h2>
```
secretrobotron: You probably want a period at the end of this sentence.

src/main.js
```diff
@@ -786,28 +784,45 @@
...
+                          project.import( projectBackup );
+                          useProject( project );
+                        },
+                        discardProject: function() {
+                          projectBackup = null;
+                          attemptDataLoad( useProject );
+                        }
+                      }
+                    });
+                    _dialog.listen( "close", function() { _dialog = null; } );
```
secretrobotron: Just a suggestion: you don't _need_ to do this.

src/util/time.js
```diff
@@ -136,10 +136,48 @@ define( [], function(){
...
+        t, seconds, minutes, hours, days;
+
+    t = ms / 1000;
+    seconds = round( t % 60 );
+    t /= 60;
+    minutes = round( t % 60 );
+    t /= 60;
+    hours = round( t % 24 );
+    t /= 24;
+    days = round( t );
```
secretrobotron, line 163: Instead of doing math to figure this stuff out, could you use Date's functions? https://developer.mozilla.org/en-US/docs/JavaScript/Reference/Global_Objects/Date#Date_instances ? `getUTC*` should give you similar results. Not sure if that's more/less confusing than what you have here.

humphd, line 163: I don't have a date, I have a number of ms, so unless I do work to parse it back into a date, then extract the bits I need...  What I have works well, does it matter--math is fast in js?  Seems like spending time rewriting this is wasting time.

PR #1272, opened 2012-10-25
T2693

humphd:


cornfield/lib/datauri.js
```diff
@@ -0,0 +1,116 @@
+"use strict";
+
+var IMAGE_DATA_URI_PREFIX = "data:image/png;base64,";
+
+var uuid = require( "node-uuid" );
+var crypto = require( "crypto" );
+
+function ImageFile( dataURI, path, url ) {
+  var _filename = path;
```
humphd: Because you pass these in as args, unless you're changing them, you don't need local variables (i.e., the arg will always be available to functions inside).  Same with url/_url.

cornfield/lib/datauri.js
```diff
@@ -0,0 +1,116 @@
...
+  var _data = new Buffer( dataURI, 'base64' );
+  var _hash = crypto.createHash( "sha1" ).update( dataURI, "utf8" ).digest( "hex" );
+
+  this.createDBReference = function( dbModel, projectId, callback ) {
+    dbModel.build({
+      filename: _filename,
+      url: _url,
+      project: projectId,
+      hash: _hash
+    }).save().complete( callback );
```
humphd: Can this fail?  What do you do if it does?

cornfield/lib/datauri.js
```diff
@@ -0,0 +1,116 @@
...
+  };
+
+  this.saveToStore = function( store, callback ) {
+    store.write( _filename, _data, callback );
```
humphd: Where are you dealing with failures?

cornfield/lib/datauri.js
```diff
@@ -0,0 +1,116 @@
...
+}
+
+function collectTrackEvents( projectData ) {
+  var trackEvents = [];
+  if ( projectData.media ) {
+    projectData.media.forEach( function( media ) {
+      if ( media.tracks ) {
+        media.tracks.forEach( function( track ) {
+          if ( track.trackEvents ) {
+            trackEvents = trackEvents.concat( track.trackEvents );
```
humphd, line 63: Sometimes you .push and other times .concat, why one over the other?

cornfield/lib/datauri.js
```diff
@@ -0,0 +1,116 @@
...
+  });
+
+  return unlinkedFiles;
+}
+
+function filterProjectDataURIs( inputData, pathFilter ) {
+  var filesToWrite = [];
+
+  var trackEvents = collectTrackEvents( inputData ).filter( function( trackEvent ) {
+    return trackEvent.popcornOptions && trackEvent.popcornOptions.src && trackEvent.popcornOptions.src.indexOf( IMAGE_DATA_URI_PREFIX ) === 0;
```
humphd: Line to long, break

cornfield/lib/datauri.js
```diff
@@ -0,0 +1,116 @@
...
+  });
+
+  if ( trackEvents.length > 0 ) {
+    trackEvents.forEach( function( trackEvent ) {
+      var file = prepareImageFile( trackEvent.popcornOptions.src.substr( IMAGE_DATA_URI_PREFIX.length ), pathFilter );
```
humphd: Line to long, break

cornfield/lib/user.js
```diff
@@ -38,6 +41,45 @@ module.exports = function( config, dbReadyFn ) {
...
+
+    createImageReferencesForProject: function( imageFiles, projectId, callback ) {
+      var savedImages = 0;
+
+      imageFiles.forEach( function( imageFile ) {
+        ImageReference.build({
+          filename: imageFile.filename,
+          url: imageFile.url,
+          project: projectId
+        }).save().complete( function( err, imageResult ) {
```
humphd, line 81: fail case?

cornfield/lib/datauri.js
secretrobotron, line 63: Here, I use `concat` because I don't have to think about what to do with each item -- I just want all of them. When I need to inspect each one, I use `push` because I'm likely in some sort of decision-making loop over a collection of items.

cornfield/config/default.json
```diff
@@ -42,6 +42,15 @@
       "nameSuffix": ".json"
     }
   },
+  "imageStore": {
+    "type": "local",
+    "options": {
+      "root": "./view",
+      "namePrefix": "images",
+      "nameSuffix": ".png",
+      "contentType": "image/png"
```
jbuck: We need a solution for when the user passes us a `image/jpeg` too. Can S3 do content-type auto-detection, so we don't need to specify this? Or do we need to modify the S3 store somehow?

cornfield/lib/datauri.js
```diff
@@ -0,0 +1,128 @@
...
+
+  if ( inputData ) {
+
+    trackEvents = collectTrackEvents( inputData ).filter( function( trackEvent ) {
+      return  trackEvent.popcornOptions &&
+              trackEvent.popcornOptions.src &&
+              trackEvent.popcornOptions.src.indexOf( IMAGE_DATA_URI_PREFIX ) === 0;
+    });
+
+    if ( trackEvents.length > 0 ) {
```
jbuck: You can just chain the .forEach to the filter above

cornfield/lib/models/image.js
```diff
@@ -0,0 +1,27 @@
...
+    id: {
+      type: DataTypes.INTEGER,
+      primaryKey: true,
+      autoIncrement: true
+    },
+    hash: {
+      type: DataTypes.STRING,
+      allowNull: false
+    },
+    url: {
```
jbuck, line 14: Just to clarify, url is the remotely-addressable url, and filename is the key that's used by the local file store and S3?

cornfield/lib/user.js
jbuck, line 81: complete() handles success or failure so this is fine

cornfield/lib/user.js
```diff
@@ -129,16 +193,29 @@ module.exports = function( config, dbReadyFn ) {
...
           email: email,
           name: data.name,
           author: data.author || "",
           template: data.template,
           latestButterVersion: versions.butter
         })
-        .complete( function(err, result) {
-          callback( err, result );
+        .complete( function( projectUpdateErr, projectUpdateResult ) {
+
```
jbuck: You need to put the callback and early return if the update fails; might just be easier to leave your added code as the .success() promise and add the old code as the .error promise.

cornfield/routes/index.js
```diff
@@ -74,32 +76,92 @@ module.exports = function routesCtor( app, User, filter, sanitizer, stores, EMBE
...
+          stores.images.remove( imageReference.filename );
+        });
+      }
+
       res.json( { error: 'okay' }, 200 );
     });
   });
 
+  function pathFilter( path ) {
+    return EMBED_HOSTNAME + '/' + stores.images.expand( path );
```
jbuck: You could probably use the utils lib here instead

cornfield/lib/models/image.js
secretrobotron, line 14: Yeah, that's right. I think I came up with this naming scheme at midnight...

PR #1285, opened 2012-10-30
[t2377] Tried adding drop-down with new, open, share, preview

k88hudson:


src/editor/header.js
```diff
@@ -2,12 +2,13 @@
  * If a copy of the MIT license was not distributed with this file, you can
  * obtain one at https://raw.github.com/mozilla/butter/master/LICENSE */
 
-define([], function(){
-  
+define([ "ui/widget/tooltip" ], function( Tooltip ){
```
secretrobotron: Spacing before `{`.

src/editor/header.js
```diff
@@ -40,6 +49,23 @@ define([], function(){
       }
     };
 
+    this.views = {
+      unSaved: function() {
+        _loginToShareTooltip.style.display = "";
```
secretrobotron: Please use a class :).

css/header.less
```diff
@@ -86,6 +86,7 @@
   max-width: 55%;
   min-width: 10%;
   height: 100%;
+  position: relative;
```
k88hudson, line 89: fixes the pencil icon position

src/editor/header.js
```diff
@@ -17,6 +18,14 @@ define([], function(){
 
     var _currentFocus;
 
+    // Create a message for the disabled share editor.
+    _loginToShareTooltip = Tooltip.create({
+      title: "login-to-share",
+      message: "Login and save your project to share!",
```
humphd: Drop the ! from this, capitalize 'Save' so it helps point them at the button:

```
Login and Save your project to share
```

src/editor/module.js
```diff
@@ -47,6 +47,22 @@ define( [ "core/eventmanager", "core/trackevent", "./editor",
 
     _header = new Header( _editorAreaDOMRoot, _this );
 
+    function setupHeader() {
+      if ( butter.project && butter.project.isSaved ) {
```
humphd: `butter.project` can't be null, this check is unnecessary.

PR #1291, opened 2012-10-30
[#2743] Only display track event delete warning dialog if more than 1.

ScottDowne:


src/ui/ui.js
```diff
@@ -474,7 +474,15 @@ define( [ "core/eventmanager", "./toggler",
             }
           }
 
-          // Delete the events.
+          // If we have one track event just delete it, otherwise display a warning dialog.
+          if ( selectedEvents.length === 1 ) {
+            selectedEvent = selectedEvents[ 0 ];
+            butter.editor.closeTrackEventEditor( selectedEvent );
+            selectedEvent.track.removeTrackEvent( selectedEvent );
```
k88hudson, line 481: Might want to combine these three lines into a single function since you have to run this twice, right?

src/ui/ui.js
```diff
@@ -474,7 +474,15 @@ define( [ "core/eventmanager", "./toggler",
...
+            return;
+          }
+
+          // Delete the events with warning dialog.
           dialog = Dialog.spawn( "delete-track", {
             data: selectedEvents.length + ( selectedEvents.length === 1 ? " track event" : " track events" ),
```
k88hudson: You don't need this check anymore, it will always be track events. Also, any reason you're writing out `selectedEvents.length`every time  instead of using `l`? Is it just more readable that way?

src/ui/ui.js
ScottDowne, line 481: I don't think it is worth adding a function call to the stack for this. It is fairly straight forward and I don't see it growing any time soon.

secretrobotron, line 481: I try to use a rule of 3's to decide things like this. If it appears in >2 places, put it in a function.

PR #1301, opened 2012-10-31
[#2634] Added soundcloud background image

dseif:


cornfield/app.js
```diff
@@ -235,6 +235,10 @@ app.post( '/api/publish/:id',
           }
         }
 
+        // if this is a soundcloud url, make sure we add a class that gives it a background-image
+        if ( mediaUrlsString.indexOf( "soundcloud" ) > -1 ) {
```
mjschranz, line 239: If you look up on line 129 we have access to the project itself that is being used here.  This means you could check against the url for the media that's being stored in the object and also use the target that's there for the media rather than hard coding in the target below.

templates/basic/style.less
```diff
@@ -111,6 +111,11 @@ body.editor-minimized .body-wrapper {
   background: #FFF;
 }
 
+.video-soundcloud {
+  background: url( "/resources/soundcloud.jpeg" );
+  background-size: contain;
+}
+
```
mjschranz: I'm thinking we should be fine placing this code somewhere inside the butter app LESS rather than what's specific for our template. This way any other templates that are written in the future will have this applied to them as well.

PR #1313, opened 2012-10-31
[t2735] prevent dragging problems

secretrobotron:
1. When our dragging starts to occur, block the browser from starting its own by using `e.preventDefault`.
2. A bit of flag cleanup and reorganization of listeners. `draggedOnce` is unnecessary.

### DIFF
--- a/src/util/dragndrop.js
+++ b/src/util/dragndrop.js
@@ -24,7 +24,6 @@ define( [ "core/eventmanager", "util/lang", "util/scroll-group" ],
       __mouseLast = [ 0, 0 ],
       __scroll = false,
       __helpers = [],
-      __draggedOnce = false,
       __scrollGroups = [],
       __nullScrollGroup = new ScrollGroup.NullScrollGroup();
 
@@ -102,11 +101,16 @@ define( [ "core/eventmanager", "util/lang", "util/scroll-group" ],
     }
   }
 
+  function __onWindowDragStart( e ) {
+    e.preventDefault();
+  }
+
   function __onDraggableDragged( e ) {
+    e.preventDefault();
+
     __mouseLast[ 0 ] = __mousePos[ 0 ];
     __mouseLast[ 1 ] = __mousePos[ 1 ];
     __mousePos = [ e.clientX, e.clientY ];
-    __draggedOnce = true;
 
     var draggables = __selectedDraggables,
         i;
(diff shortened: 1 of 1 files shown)


src/util/dragndrop.js
```diff
@@ -139,47 +148,42 @@ define( [ "core/eventmanager", "util/lang", "util/scroll-group" ],
...
-
-      // Let droppable know that it's about to receive one or more items
-      for ( i = droppables.length - 1; i >= 0; --i ) {
-        droppables[ i ].startDrop();
-      }
-
-      for ( i = selectedDraggables.length - 1; i >= 0; --i ) {
-        selectedDraggable = selectedDraggables[ i ];
-        selectedDraggable.stop();
+    // Collect all the droppables
```
secretrobotron, line 155: All of this is really just an out-dent.

src/util/dragndrop.js
```diff
@@ -139,47 +148,42 @@ define( [ "core/eventmanager", "util/lang", "util/scroll-group" ],
...
+    }
 
-      // Let droppable know that we're done dropping
-      for ( i = droppables.length - 1; i >= 0; --i ) {
-        droppables[ i ].stopDrop();
-      }
+    for ( i = selectedDraggables.length - 1; i >= 0; --i ) {
+      selectedDraggable = selectedDraggables[ i ];
+      selectedDraggable.drop();
+    }
```
dseif, line 177: this for loop and the one above look to be the same. Accidental duplicate?

src/util/dragndrop.js
```diff
@@ -139,47 +148,42 @@ define( [ "core/eventmanager", "util/lang", "util/scroll-group" ],
...
 
-      DragNDrop.dispatch( "dropfinished" );
+    for ( i = selectedDraggables.length - 1; i >= 0; --i ) {
+      selectedDraggable = selectedDraggables[ i ];
+      selectedDraggable.reset();
+    }
```
dseif, line 182: also can these 3 ( or 2 ) for loops not be joined into 1 single one? Seems like they are looping over the same data each time.

src/util/dragndrop.js
secretrobotron, line 177: Nope. One does `drop`, one does `stop`. They happen like this to make sure everything is in a sane state before moving to the next step.

src/util/dragndrop.js
secretrobotron, line 182: Same as above. Everything works like a clock here. Could be optimized later, perhaps, but I'd rather not fiddle with it now. It works as intended, and moving stuff around now might be an unnecessary risk.

I really do appreciate the thorough review, but heed the comment here https://github.com/mozilla/butter/pull/1313/files#L0R151.

PR #1316, opened 2012-10-31
T2775

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/2775-improve-autosave-experience


src/ui/header.js
```diff
@@ -117,6 +123,23 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
...
         toggleShareButton( isSaved );
+
+        if ( butter.autosave ) {
+          var project = butter.autosave,
+              location = window.location.search,
+              queryString = location.substring( 0, location.lastIndexOf( "?" ) );
+
+          _projectAutoSave.classList.remove( "hidden" );
+          _projectAutoSave.href = queryString + "?loadAutoSave=true";
+          _projectAutoSave.innerHTML = "Hey! We found an auto save for <span class='autosave-name'>" + project.name + "</span>, last saved " +
```
secretrobotron: You can probably place the majority of this in the layout, and only manipulate the `innerHTML` of the `autosave-name` span, right?

src/core/project.js
```diff
@@ -289,6 +302,8 @@ define( [ 'core/eventmanager', 'core/media' ],
     // to be saved (i.e., it has been changed since last save, or was never
     // saved before).
     _this.save = function( callback ) {
+      var history = window.history;
```
secretrobotron: Probably doesn't need to be cached. You can just reference `window.history` directly below, right?

src/core/project.js
```diff
@@ -246,9 +241,22 @@ define( [ 'core/eventmanager', 'core/media' ],
         console.warn( "Ignored imported media data. Must be in an Array." );
       }
 
+      if ( json.projectID ) {
+        _id = json.projectID;
+
+        // This means the project we imported as a back up as well for an existing project
+        // However, if the recovered auto save doesn't match the id of the project they loaded
+        // we shouldn't auto save the backup
```
secretrobotron: Missing period.

src/core/project.js
```diff
@@ -246,9 +241,22 @@ define( [ 'core/eventmanager', 'core/media' ],
         console.warn( "Ignored imported media data. Must be in an Array." );
       }
 
+      if ( json.projectID ) {
+        _id = json.projectID;
+
+        // This means the project we imported as a back up as well for an existing project
+        // However, if the recovered auto save doesn't match the id of the project they loaded
```
secretrobotron: s/they/"the user"

src/core/project.js
```diff
@@ -344,35 +365,58 @@ define( [ 'core/eventmanager', 'core/media' ],
...
-    }
+    // Check for an existing project that was autosaved but not saved.
+    // Returns project backup data as JS object if found, otherwise null.
+    _this.checkForBackup = function( readyCB, fallbackCB ) {
+      // See if we already have a project autosaved from another session.
+      var projectBackup,
+          location = window.location.search,
+          loadAutoSave = location.indexOf( "loadAutoSave" ) > -1 ? true : false;
+
+      if ( !fallbackCB ) {
```
secretrobotron: For brevity's sake, you can probably just use `fallbackCB = fallbackCB || function() {};`.

src/core/project.js
```diff
@@ -344,35 +365,58 @@ define( [ 'core/eventmanager', 'core/media' ],
...
+        fallbackCB = function() {};
+      }
+
+      if ( !readyCB ) {
+        readyCB = function() {};
```
secretrobotron: Same as https://github.com/mozilla/butter/pull/1316/files#L1R377

src/ui/header.js
```diff
@@ -1,5 +1,7 @@
-define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data", "ui/webmakernav/webmakernav", "ui/widget/tooltip" ],
-  function( Dialog, Lang, HEADER_TEMPLATE, UserData, WebmakerBar, ToolTip ) {
+define(
+  [ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data", "ui/webmakernav/webmakernav",
+    "ui/widget/tooltip", "util/time" ],
+  function( Dialog, Lang, HEADER_TEMPLATE, UserData, WebmakerBar, ToolTip, TimeUtil ) {
```
secretrobotron: I've been trying to keep the newlines consistent in the function signature with the newlines in the array above it so that you can easily see which objects belong to which source files. Mind cutting this line the way you did the one on top of it?

src/ui/header.js
```diff
@@ -101,6 +104,9 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
         togglePreviewButton( false );
         toggleSaveButton( true );
         toggleShareButton( true );
+
+        butter.project.autosave = _projectAutoSave.onclick = null;
```
secretrobotron: I'm looking at this again, and I think pushing logic for `butter.project.autosave` back into `project` altogether is a good idea. I don't think `header` code should be telling `butter.project.autosave` whether it's `null`, or otherwise.

src/ui/header.js
```diff
@@ -117,6 +123,33 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
...
+            _projectAutoSave.classList.remove( "hidden" );
+            _projectAutoSave.href = queryString + "?savedDataUrl=/api/project/" + project.projectID;
+            project.isAutoSave = true;
+            autoSaveName.innerHTML = project.name;
+            autoSaveTime.innerHTML = TimeUtil.toPrettyString( Date.now() - project.backupDate );
+
+            _projectAutoSave.onclick = function() {
+              // Rebackup the project with a flag attached for loading it
+              butter.project.backupData( project );
+              butter.project.autosave = null;
```
secretrobotron: Same as https://github.com/mozilla/butter/pull/1316/files#L8R108

src/ui/header.js
```diff
@@ -117,6 +123,33 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
...
+          if ( butter.project.autosave ) {
+            var project = butter.project.autosave,
+                location = window.location.search,
+                queryString = location.substring( 0, location.lastIndexOf( "?" ) ),
+                autoSaveName = _projectAutoSave.querySelector( ".autosave-name" ),
+                autoSaveTime = _projectAutoSave.querySelector( ".autosave-time" );
+
+            _projectAutoSave.classList.remove( "hidden" );
+            _projectAutoSave.href = queryString + "?savedDataUrl=/api/project/" + project.projectID;
+            project.isAutoSave = true;
```
secretrobotron: Same as https://github.com/mozilla/butter/pull/1316/files#L8R108.

src/core/project.js
```diff
@@ -149,6 +156,16 @@ define( [ 'core/eventmanager', 'core/media' ],
           return _isPublished && !_isDirty;
         },
         enumerable: true
+      },
+
+      "projectHistoryState": {
```
secretrobotron: This probably doesn't need to be a property defined like this. Anything wrong with just having `this.projectHistoryState`?

src/ui/header.js
```diff
@@ -265,8 +311,8 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
     butter.listen( "authenticated", _this.views.login, false );
     butter.listen( "logout", _this.views.logout, false );
 
-    //Default view
-    _this.views.logout();
+    // Default state is the save button should be active
+    _saveButton.addEventListener( "click", saveProject, false );
```
secretrobotron: Can you use `toggleSaveButton( true );` instead?

src/core/project.js
```diff
@@ -359,35 +399,54 @@ define( [ 'core/eventmanager', 'core/media' ],
...
+        fallbackCB();
+        return;
+      }
+
+      try {
+        projectBackup = __butterStorage.getItem( "butter-backup-project" );
+        projectBackup = JSON.parse( projectBackup );
+      } catch( e ) { }
+
+      if ( projectBackup && projectBackup.isAutoSave ) {
```
secretrobotron: The check for `projectBackup` can just happen once here (if you wrap this in another if block) removing the need for the `else if` below (just becomes an `else`).

src/core/project.js
```diff
@@ -359,35 +399,54 @@ define( [ 'core/eventmanager', 'core/media' ],
...
-
-    // For testing purposes, we can skip backup recovery
-    if ( butter.config.value( "recover" ) === "purge" ) {
-      callback( null, null );
-      return;
-    }
+    // Check for an existing project that was autosaved but not saved.
+    // Returns project backup data as JS object if found, otherwise null.
+    _this.checkForBackup = function( readyCB, fallbackCB ) {
+      // See if we already have a project autosaved from another session.
```
secretrobotron: This comment looks out of place here. Should it be above the `try` below?

src/core/project.js
```diff
@@ -359,35 +399,54 @@ define( [ 'core/eventmanager', 'core/media' ],
...
+      // For testing purposes, we can skip backup recovery
+      if ( butter.config.value( "recover" ) === "purge" ) {
+        fallbackCB();
+        return;
+      }
+
+      try {
+        projectBackup = __butterStorage.getItem( "butter-backup-project" );
+        projectBackup = JSON.parse( projectBackup );
+      } catch( e ) { }
```
secretrobotron: A warning should happen in this catch. It's a bizarre enough situation if a backup fails like this that we'd want to know.

src/core/project.js
```diff
@@ -359,35 +386,59 @@ define( [ 'core/eventmanager', 'core/media' ],
...
-    var projectBackup, backupDate;
-
-    // For testing purposes, we can skip backup recovery
-    if ( butter.config.value( "recover" ) === "purge" ) {
-      callback( null, null );
-      return;
-    }
+    // Check for an existing project that was autosaved but not saved.
+    // Returns project backup data as JS object if found, otherwise null.
+    _this.checkForBackup = function( readyCB, fallbackCB ) {
```
humphd: We use 'callback' vs. 'CB' everywhere else in our tree, do it here too.

src/core/project.js
```diff
@@ -359,35 +386,59 @@ define( [ 'core/eventmanager', 'core/media' ],
...
+        }
+      }
 
       if ( projectBackup ) {
-        backupDate = projectBackup.backupDate;
+        if ( projectBackup.isAutoSave ) {
+          _this.import( projectBackup );
+          readyCB( _this );
+        } else {
+          var id = location.substring( location.lastIndexOf( "/" ) + 1 );
```
humphd: Use util/uri.js to parse this, in case we add more things to the URL later.  Also, after you get the `savedDataUrl` query key, use a regex to clearly show what you're meaning to get (in case our URL structure changes, this will be more clear what you meant by the last /):

```
/\/api\/project\/(\d+)/
```

src/default-config.json
```diff
@@ -5,7 +5,7 @@
   "cssRenderClientSide": false,
   "snapshotHTMLOnReady": false,
   "scrapePage": true,
-  "backupInterval": 30000,
+  "backupInterval": 10000,
```
humphd: Make sure you bump this back to 30

src/main.js
```diff
@@ -784,43 +784,21 @@
           preparePopcornScriptsAndCallbacks( function(){
             preparePage( function(){
               moduleCollection.ready( function(){
+
+                function readyCB( project ) {
```
humphd: Fix naming here too

src/core/project.js
```diff
@@ -271,18 +278,27 @@ define( [ 'core/eventmanager', 'core/media' ],
     };
 
     // Expose backupData() to make testing possible
-    var backupData = _this.backupData = function() {
+    var backupData = _this.backupData = function( backup ) {
+      var data;
+
       // If the project isn't different from last time, or if it's known
       // to not fit in storage, don't bother trying.
```
humphd, line 285: This comment is no longer in sync with the code it precedes, please update to reflect what's really happening.

src/core/project.js
```diff
@@ -271,18 +278,27 @@ define( [ 'core/eventmanager', 'core/media' ],
...
+      if ( !backup ) {
+        data = _this.data;
+        data.projectID = _id;
+        data.name = _name;
+        data.template = _template;
+        data.author = _author;
+        data.backupDate = Date.now();
+      } else {
+        data = backup;
+        data.loadMe = true;
```
humphd, line 299: This name is not great, can you do better?  I'm still confused by it even after all your explanations.  It shouldn't be named in a way that relates to UI, but to a Project and the logic of a Project.

src/core/project.js
```diff
@@ -359,35 +382,45 @@ define( [ 'core/eventmanager', 'core/media' ],
...
+      }
 
       if ( projectBackup ) {
-        backupDate = projectBackup.backupDate;
+        if ( projectBackup.loadMe ) {
+          _this.import( projectBackup );
+          readyCallback( _this );
+          return;
+        } else {
+          _this.autosave = projectBackup;
```
humphd, line 417: Don't use "autosave"

src/layouts/header.html
```diff
@@ -5,21 +5,29 @@
...
-            <li><a class="butter-share-btn butter-btn btn-light" target="_blank"><span class="icon icon-only icon-share"></span> Share</a></li>
+            <li><a class="butter-preview-btn butter-btn btn-light butter-disabled" target="_blank" style="display: none;"><span class="icon icon-only icon-share-alt"></span> Preview</a></li>
+            <li><a class="butter-share-btn butter-btn btn-light butter-disabled" target="_blank"><span class="icon icon-only icon-share"></span> Share</a></li>
           </ul>
         </div>
       </li>
+      <li>
+        <div class="butter-project-autosave">
+          <a class="backup-available hidden" target="_blank">
+            Hey! We found an auto save for <span class="backup-name"></span>
```
humphd, line 26: "We found a backup for {title} from {time} ago."

PR #1317, opened 2012-11-01
[#2731] Popup uses font percentages for font size and image size. Existi...

ScottDowne:
...ng popups do not change, and do not use this feature.

### DIFF
--- a/templates/assets/editors/popup/popup-editor.js
+++ b/templates/assets/editors/popup/popup-editor.js
@@ -11,7 +11,8 @@
 
     var _rootElement = rootElement,
         _trackEvent,
-        _butter;
+        _butter,
+        _popcornOptions;
 
     /**
      * Member: setup
@@ -21,7 +22,8 @@
      * @param {TrackEvent} trackEvent: The TrackEvent being edited
      */
     function setup( trackEvent ) {
-      _trackEvent = trackEvent;
+      _trackEvent = trackEvent,
+      _popcornOptions = _trackEvent.popcornOptions;
 
       var basicContainer = _rootElement.querySelector( ".editor-options" ),
           advancedContainer = _rootElement.querySelector( ".advanced-options" ),
@@ -140,6 +142,12 @@
         }
 
         basicContainer.insertBefore( _this.createStartEndInputs( trackEvent, _this.updateTrackEventSafe ), basicContainer.firstChild );
+      }
+
+      if ( _popcornOptions.fontSize ) {
+        trackEvent.manifest.options.fontPercentage.hidden = true;
+      } else {
+        trackEvent.manifest.options.fontSize.hidden = true;
       }
 
       _this.createPropertiesFromManifest({
--- a/templates/assets/editors/popup/popup-editor.js
+++ b/templates/assets/editors/popup/popup-editor.js
@@ -11,6 +11,7 @@
 
     var _rootElement = rootElement,
         _trackEvent,
+        _manifestOptions,
         _butter,
         _popcornOptions;
 
@@ -23,6 +24,7 @@
      */
     function setup( trackEvent ) {
       _trackEvent = trackEvent;
+      _manifestOptions = _trackEvent.manifest.options;
       _popcornOptions = _trackEvent.popcornOptions;
 
       var basicContainer = _rootElement.querySelector( ".editor-options" ),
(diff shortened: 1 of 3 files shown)


templates/assets/plugins/popup/popcorn.popup.js
```diff
@@ -391,20 +404,26 @@
...
+                img.style.width = width + "px";
+              }
+              img.style.left = -( width - 16 ) + "px";
+              // make sure container is still non-null
+              // if _teardown is called too quickly, it will become null before img loads
+              if ( container && container.offsetHeight ) {
                 img.style.top = ( container.offsetHeight - height ) / 2 - 4 + "px";
               }
+            } else {
+              img.style.height = "110%";
```
k88hudson: Since these are static values, there is no reason to set them here. You should put them in the stylesheet.

templates/assets/plugins/popup/popcorn.popup.js
```diff
@@ -391,20 +404,26 @@
...
+              img.style.right = "100%";
+              img.style.top = "-5%";
+              img.style.marginRight = "-10px";
```
k88hudson: I think you missed this value -- should be -5% or something.

templates/assets/plugins/popup/popcorn.popup.js
```diff
@@ -193,8 +193,16 @@
...
-          "default": 12,
           units: "px",
+          group: "advanced",
+          depricated: true
+        },
+        fontPercentage: {
+          elem: "input",
+          type: "number",
+          label: "Font Size",
+          "default": 5,
```
k88hudson: Make this 4, it computes to between 12-13, which is closer to the original default.

templates/assets/plugins/popup/popcorn.popup.js
ScottDowne: Why should it be -5%?

I just picked a fairly random number that looked decent. You?

ScottDowne: I see what you're saying.

This needed to be px and not % because otherwise when you start to scale up or resize, it'll start to overlap over the left of the main container.

ScottDowne: Wait, margin right is based on % of the parent or the actual element?

I'll see if I can make this percentage, but px is fairly harmless so I won't spend much time.

templates/assets/plugins/popup/popcorn.popup.js
ScottDowne: I wanted to avoid getting mixed up between new and old. How about a classname that sets everything new?

templates/assets/plugins/popup/popcorn.popup.js
k88hudson: Sorry, I assumed the border would be in `%` but it's not ATM. This is fine as is.

templates/assets/plugins/popup/popcorn.popup.js
k88hudson: Classname that sets everything for new popup images is good

templates/assets/editors/popup/popup-editor.js
```diff
@@ -142,6 +144,14 @@
         basicContainer.insertBefore( _this.createStartEndInputs( trackEvent, _this.updateTrackEventSafe ), basicContainer.firstChild );
       }
 
+      if ( _popcornOptions.fontSize ) {
+        trackEvent.manifest.options.fontPercentage.hidden = true;
+        trackEvent.manifest.options.fontSize.hidden = false;
+      } else {
+        trackEvent.manifest.options.fontSize.hidden = true;
+        trackEvent.manifest.options.fontPercentage.hidden = false;
```
mjschranz: Should probably just cache trackEvent.manifest.options here.

templates/assets/plugins/popup/popcorn.popup.js
```diff
@@ -193,8 +193,16 @@
           elem: "input",
           type: "number",
           label: "Font Size",
-          "default": 12,
           units: "px",
+          group: "advanced",
+          depricated: true
```
mjschranz: Doesn't look like we make use of this anywhere in the plugin or it's editor.

ScottDowne: We do not, but I thought it made a lot of sense as a form of documentation.

In retrospec, it probably is a bad idea because we then have to support it.

ScottDowne: I spelled it wrong too :O

PR #1323, opened 2012-11-01
T2712

secretrobotron:


src/core/project.js
```diff
@@ -285,10 +285,14 @@ define( [ 'core/eventmanager', 'core/media' ],
       }
     };
 
+    function savePlaceholder() {
+      console.warn( "Save already in progress. Ignoring." );
```
jbuck: Instead of having this code here, could we just rely on the code in src/cornfield/module.js to prevent too many saves from happening?

src/cornfield/module.js
```diff
@@ -14,6 +14,16 @@ define([ "util/xhr", "jsSHA/sha1" ],
     return location.protocol + "//" + location.hostname + ( location.port ? ":" + location.port : "" );
   }
 
+  function savePlaceholder( id, data, callback ) {
+    console.warn( "Warning: Save is already in progress. Ignoring request." );
```
jbuck: Wrap this is a try catch so when IE9 barfs on this it doesn't crash.

src/cornfield/module.js
```diff
@@ -14,6 +14,16 @@ define([ "util/xhr", "jsSHA/sha1" ],
...
+    callback( { error: "Save is already in progress. Ignoring request." } );
```
jbuck: Should we send back an error? What if we just didn't do anything silently?

src/cornfield/module.js
```diff
@@ -137,6 +150,7 @@ define([ "util/xhr", "jsSHA/sha1" ],
           }
```
jbuck, line 151: Bah, I can't put this on the right line of context, but if the publish fails then you need to change re-assign self.publish as well

src/cornfield/module.js
```diff
@@ -137,6 +150,7 @@ define([ "util/xhr", "jsSHA/sha1" ],
...
 
           callback(response);
+          self.publish = publishFunction;
```
jbuck: Although it doesn't matter in this case, since we don't call anything after save, we should put the function re-assignments before we callback. This is really only important for save(), because in the callback we call publish()

src/cornfield/module.js
secretrobotron: I send back an error so the app can keep moving instead of blocking on waiting for a callback from a save action. Unnecessary?

src/cornfield/module.js
```diff
@@ -14,6 +14,21 @@ define([ "util/xhr", "jsSHA/sha1" ],
     return location.protocol + "//" + location.hostname + ( location.port ? ":" + location.port : "" );
   }
 
+  function savePlaceholder( id, data, callback ) {
+    try {
+      console.warn( "Warning: Save is already in progress. Ignoring request." );
```
humphd: Add a "Popcorn Maker" string to this warning.

src/cornfield/module.js
```diff
@@ -14,6 +14,21 @@ define([ "util/xhr", "jsSHA/sha1" ],
     return location.protocol + "//" + location.hostname + ( location.port ? ":" + location.port : "" );
   }
 
+  function savePlaceholder( id, data, callback ) {
+    try {
```
humphd: Why aren't we just shimming console instead of always assuming it isn't there?

src/cornfield/module.js
```diff
@@ -14,6 +14,21 @@ define([ "util/xhr", "jsSHA/sha1" ],
...
+      console.warn( "Warning: Save is already in progress. Ignoring request." );
+    }
+    catch ( e ) {}
+    callback( { error: "Save is already in progress. Ignoring request." } );
+  }
+
+  function publishPlaceholder( id, callback ) {
+    try {
+      console.warn( "Warning: Publish is already in progress. Ignoring request." );
```
humphd: Same.

src/cornfield/module.js
```diff
@@ -125,18 +141,24 @@ define([ "util/xhr", "jsSHA/sha1" ],
...
+      // Re-route successive calls to `publish` until a complete response has been
+      // received from the server.
+      self.publish = publishPlaceholder;
+
       sendXHRPost(server + "/api/publish/" + id, null, function() {
         if (this.readyState === 4) {
           var response;
+
+          // Reset publish function to its original incarnation.
+          self.publish = publishFunction;
```
humphd, line 144: To make this a bit more clear, pull publishFunction out of here, and move it up under publishPlaceholder:

```
function publishPlaceholder(...) {
  ...
}

function publishFunction(...) {
  ...
}

this.publish = publishFunction;
```

This way it's more clear when you swap.  Same deal for saveFunction and savePlaceholder.

src/cornfield/module.js
```diff
@@ -164,7 +186,11 @@ define([ "util/xhr", "jsSHA/sha1" ],
       });
     };
 
-    this.save = function( id, data, callback ) {
+    this.save = function saveFunction( id, data, callback ) {
```
humphd: Same layout as above.

PR #1324, opened 2012-11-01
Bug 2591 - Prepare projects to be fronted by the clouds

jbuck:
This patch prepares our S3 code to be put behind the CloudFront CDN service by:
- Providing cache-busting URLs to the share editor
- Passing cache-busting URLs through the embed shell to the embed
- Added configuration option to pass arbitrary headers to S3 when writing an object
- Documentation for the S3 changes

### DIFF
--- a/cornfield/lib/file-store.js
+++ b/cornfield/lib/file-store.js
@@ -112,16 +112,27 @@ function S3FileStore( options ) {
 
   // If embedHostname isn't appropriate, a hostname can be provided to override it.
   this.hostname = options.hostname;
+
+  // Append any additional headers to send with a write request
+  if ( options.headers ) {
+    this.headers = options.headers;
+  }
 }
 
 S3FileStore.prototype = Object.create( BaseFileStore );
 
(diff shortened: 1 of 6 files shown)


cornfield/lib/file-store.js
```diff
@@ -107,16 +107,27 @@ function S3FileStore( options ) {
...
 S3FileStore.prototype.write = function( key, data, callback ) {
-  this.client.put( this.expand( key ), {
+  var headers = {
     'x-amz-acl': 'public-read',
     'Content-Length': data.length,
     'Content-Type': this.contentType
-  })
+  };
+
+  Object.keys(this.headers).forEach(function(key) {
```
humphd: Spacing around args-- (<space>this.headers<space>), same with forEach, and key

cornfield/lib/file-store.js
```diff
@@ -107,16 +107,27 @@ function S3FileStore( options ) {
...
+    headers[key] = this.headers[key];
```
humphd: Spacing around key, twice.

PR #1325, opened 2012-11-01
[#2737] Add ua-parser library for more robust user agent detection.

cadecairos:
Lighthouse: https://webmademovies.lighthouseapp.com/projects/65733/tickets/2737-differences-in-ua-string-composition-problem-for-ua-warning-dialog

### DIFF
--- a/external/ua-parser/ua-parser.js
+++ b/external/ua-parser/ua-parser.js
@@ -1,0 +1,344 @@
+// UA-Parser.JS v0.4.6
+// Lightweight JavaScript-based User-Agent string parser
+// https://github.com/faisalman/ua-parser-js
+//
+// Copyright © 2012 Faisalman
+// licensed under MIT
+//
+// commit: https://github.com/faisalman/ua-parser-js/commit/4efb13be3ee21cab86d1b2e9f740572346b497bf
+
+(function (undefined) {
+
+    'use strict';
+
+    var mapper = {
+
+        regex : function () {
+
+            var result, i, j, k, l, m, args = arguments;
+
+            // loop through all regexes maps
+            for (i = 0; i < args.length; i += 2) {
+
+                var regex = args[i],       // odd sequence (0,2,4,..)
+                    props = args[i + 1];   // even sequence (1,3,5,..)
+
+                // construct object barebones
+                if (typeof result === 'undefined') {
+                    result = {};
+                    for (k = 0; k < props.length; k++) {
+                        if (typeof props[k] === 'object') {
+                            result[props[k][0]] = undefined;
+                        } else {
+                            result[props[k]] = undefined;
+                        }
+                    }
+                    if (this.getUA().toString() === '') {
+                        return result;
+                    }
+                }
+
+                // try matching uastring with regexes
+                for (j = 0; j < regex.length; j++) {
+                    l = regex[j].exec(this.getUA());
+                    if (!!l) {
+                        for (k = 0; k < props.length; k++) {
+                            m = l[k + 1];
+                            if (typeof props[k] === 'object' && props[k].length === 2) {
+                                result[props[k][0]] = props[k][1];
+                            } else if (typeof props[k] === 'object' && props[k].length === 3) {
+                                if (typeof props[k][1] === 'function' && !(props[k][1].exec && props[k][1].test)) {
+                                    result[props[k][0]] = m ? props[k][1].call(this, m, props[k][2]) : undefined;
+                                } else {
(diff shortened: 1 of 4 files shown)


src/util/ua-parser.js
```diff
@@ -0,0 +1,395 @@
+// UA-Parser.JS v0.4.5
+// Lightweight JavaScript-based User-Agent string parser
+// https://github.com/faisalman/ua-parser-js
+//
+// Copyright © 2012 Faisalman
+// Dual licensed under GPLv2 & MIT
```
secretrobotron:
```
humph
  just indicate we're using under MIT
humph
  don't mention gpl
```

src/util/ua-parser.js
```diff
@@ -0,0 +1,395 @@
+// UA-Parser.JS v0.4.5
```
secretrobotron: Not that it's a _huge_ deal, but I've generally been putting truly external code in the `external` directory so that we know where it all is and it's not surprising. You don't need to alter any of the code or wrap it in a `define`, since it's really a library we're including. Then, if it needs replacing, we can just replace the whole file, etc. etc. It might be worth keeping the original code in-tact so we don't have to worry that we've changed something by accident.

secretrobotron: And, to make lint happy, I've just done this: https://github.com/mozilla/butter/blob/master/src/cornfield/module.js#L9

src/main.js
```diff
@@ -27,7 +27,7 @@
             EventManager, Logger, Config, Target, Media, Page,
             Modules, Dependencies, Dialogs,
             Dialog, Editor, UI,
-            XHR, Lang, Tutorial,
+            XHR, Lang, Tutorial, UaParser,
```
secretrobotron: This should probably be `UAParser`.

src/main.js
```diff
@@ -52,14 +52,17 @@
...
+          acceptedUA = false;
+
       for ( var uaName in ACCEPTED_UA_LIST ) {
-        if( ACCEPTED_UA_LIST.hasOwnProperty( uaName ) ) {
-          var uaRegex = new RegExp( uaName + "(?:/|\\s)([0-9]+)\\.", "g" ),
-              match = uaRegex.exec( ua );
-          if ( match && match.length === 2 && Number( match[ 1 ] ) >= ACCEPTED_UA_LIST[ uaName ] ) {
-            acceptedUA = uaName + "/" + match[ 1 ];
+        if ( name === uaName ) {
+          console.log( typeof major );
```
secretrobotron: Please remove.

cadecairos: woopsie

src/config.js
```diff
@@ -11,7 +11,8 @@ define(function( require, exports, module ){
     // Paths are relative to the baseUrl
     paths: {
       'text': '../external/require/text',
-      'jsSHA': '../external/jsSHA'
+      'jsSHA': '../external/jsSHA',
+      'UAParser': '../external/ua-parser'
```
secretrobotron, line 15: Make sure you add this change to `tools/build.js` too so deploy doesn't break.

external/ua-parser/ua-parser.js
```diff
@@ -0,0 +1,342 @@
+// UA-Parser.JS v0.4.6
+// Lightweight JavaScript-based User-Agent string parser
+// https://github.com/faisalman/ua-parser-js
+//
+// Copyright © 2012 Faisalman
+// licensed under MIT
```
humphd, line 6: Please include a link to the commit (i.e., version from git) we have in our tree

src/main.js
```diff
@@ -21,7 +21,7 @@
             "util/xhr", "util/lang", "util/tutorial",
             "text!default-config.json", "text!layouts/ua-warning.html",
             "ui/widget/tooltip", "crashreporter", "core/project",
-            "util/shims"                  // keep this at the end so it doesn't need a spot in the function signature
+            "util/shims", "UAParser/ua-parser"    // keep these at the end so they don't need a spot in the function signature
```
humphd: No, shims has to go last.  Move your UAParser line up one.

src/main.js
```diff
@@ -30,9 +30,12 @@
             XHR, Lang, Tutorial,
             DEFAULT_CONFIG_JSON, UA_WARNING_LAYOUT,
             ToolTip, CrashReporter, Project,
-            Shims                         // placeholder
+            Shims               // placeholder
```
humphd: We don't need this, that's the point of having it go last.

PR #1336, opened 2012-11-02
[#2776] Possible for Twitter ticker hack to be applied when there were n...

mjschranz:
...ot tweets retrieved

https://webmademovies.lighthouseapp.com/projects/65733/tickets/2776-crash-typeerror-elem-is-null-httpspopcornwebmakerorgexternalpopcorn-jspopcornjs4100

### DIFF
--- a/templates/assets/plugins/twitter/popcorn.twitter.js
+++ b/templates/assets/plugins/twitter/popcorn.twitter.js
@@ -186,11 +186,15 @@
         // Set layout class for container
         if ( options.layout ) {
           options._container.classList.add( options.layout );
-          if ( options.layout === "ticker" ) {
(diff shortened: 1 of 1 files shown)


templates/assets/plugins/twitter/popcorn.twitter.js
```diff
@@ -186,11 +186,16 @@
         // Set layout class for container
         if ( options.layout ) {
           options._container.classList.add( options.layout );
-          if ( options.layout === "ticker" ) {
+          if ( options.layout === "ticker" && tweetsContainer.childNodes.length ) {
             var elem;
 
             options._tickerInterval = setInterval(function() {
               elem = tweetsContainer.firstChild;
+
```
humphd: Remove this blank line, so the return is clearly paired with this assignment.

PR #1346, opened 2012-11-02
[#2800] Updated UI only after media is ready, to be in synch with what t...

ScottDowne:
...he plugin tray has to wait for.

### DIFF
--- a/src/timeline/media.js
+++ b/src/timeline/media.js
@@ -265,7 +265,7 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
       }
     }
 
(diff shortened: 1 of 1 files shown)


src/timeline/media.js
```diff
@@ -266,7 +265,7 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
       }
     }
 
-    butter.listen( "ready", function(){
+    butter.listen( "mediaready", function(){
```
secretrobotron: This can be left as `ready`. The `onMediaReady` function above already calls `updateUI`.

PR #1348, opened 2012-11-05
[#2808] Potential on project load, remix and previewing for google maps ...

mjschranz:
...api to be queried too quickly. This forces it to retry after a set period

https://webmademovies.lighthouseapp.com/projects/65733-popcorn-maker/tickets/2808-messing-with-maps-doesnt-load-in-firefox#ticket-2808-15

### DIFF
--- a/templates/assets/plugins/googlemap/popcorn.googlemap.js
+++ b/templates/assets/plugins/googlemap/popcorn.googlemap.js
@@ -12,6 +12,9 @@ var googleCallback;
 
   var _mapFired = false,
       _mapLoaded = false,
+      // Store location objects in case the same string location is used multiple times.
+      _cachedGeoCode = {},
+      MAP_FAILURE_TIMEOUT = 100,
       geocoder;
 
   //google api callback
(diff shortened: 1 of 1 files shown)


templates/assets/plugins/googlemap/popcorn.googlemap.js
```diff
@@ -169,26 +170,47 @@ var googleCallback;
       target.appendChild( outerdiv );
     }
 
+    function geoCodeCallback( results, status ) {
+      // second check for innerdiv since it could have disappeared before
+      // this callback is actual run
+      if ( innerdiv && status === google.maps.GeocoderStatus.OK ) {
+        options.lat = results[ 0 ].geometry.location.lat();
+        options.lng = results[ 0 ].geometry.location.lng();
+        _cachedGeoCode[ options.location ] = {
```
humphd: At 205 below, you create a new LatLng object with these scalars; let's just do it here so we don't have the overhead on every use:

```
_cachedGeoCode[ options.location ] = new google.maps.LatLng( options.lat, options.lng );
...
       var cachedLocation = _cachedGeoCode[ options.location ];
        if ( cachedLocation ) {
          location = cachedLocation;
```

templates/assets/plugins/googlemap/popcorn.googlemap.js
```diff
@@ -169,26 +170,47 @@ var googleCallback;
...
+
+        location = new google.maps.LatLng( options.lat, options.lng );
+        map = buildMap( options, innerdiv, that );
+      } else if ( status === google.maps.GeocoderStatus.OVER_QUERY_LIMIT ) {
+        setTimeout(function() {
+          // calls an anonymous google function called on separate thread
+          geocoder.geocode({
+            "address": options.location
+          }, geoCodeCallback );
+        }, 10 );
```
humphd: Why 10?  Also, why on a timeout?  This will already happen async, since you pass in a callback.

templates/assets/plugins/googlemap/popcorn.googlemap.js
```diff
@@ -12,6 +12,7 @@ var googleCallback;
 
   var _mapFired = false,
       _mapLoaded = false,
+      _cachedGeoCode = {},
```
humphd: Add a comment so it's clear what this is for, and so uses of it below are clear.

templates/assets/plugins/googlemap/popcorn.googlemap.js
```diff
@@ -169,26 +170,47 @@ var googleCallback;
...
+        options.lat = results[ 0 ].geometry.location.lat();
+        options.lng = results[ 0 ].geometry.location.lng();
+        _cachedGeoCode[ options.location ] = {
+          lat: options.lat,
+          lng: options.lng
+        };
+
+        location = new google.maps.LatLng( options.lat, options.lng );
+        map = buildMap( options, innerdiv, that );
+      } else if ( status === google.maps.GeocoderStatus.OVER_QUERY_LIMIT ) {
```
humphd: This `else if` looks like it hides a bug.  Above you check if innerDiv is there, and if status is OK.  Could you have a case where status is OK, but innerDiv is null?  If so, you'll enter an infinite loop, I think.  Probably you want to separate your innerDiv check from the status checks.

templates/assets/plugins/googlemap/popcorn.googlemap.js
```diff
@@ -12,6 +12,9 @@ var googleCallback;
 
   var _mapFired = false,
       _mapLoaded = false,
+      // Store Location objects in the event the same string location is used multiple times
```
humphd: Grammar is odd here.

templates/assets/plugins/googlemap/popcorn.googlemap.js
```diff
@@ -169,26 +172,44 @@ var googleCallback;
       target.appendChild( outerdiv );
     }
 
+    function geoCodeCallback( results, status ) {
+      // second check for innerdiv since it could have disappeared before
+      // this callback is actual run
+      if ( innerdiv ) {
```
humphd: Just return early if it's not there instead of doing a deeply nested if block underneath.

templates/assets/plugins/googlemap/popcorn.googlemap.js
```diff
@@ -169,26 +172,44 @@ var googleCallback;
...
+        if ( status === google.maps.GeocoderStatus.OK ) {
+          options.lat = results[ 0 ].geometry.location.lat();
+          options.lng = results[ 0 ].geometry.location.lng();
+          _cachedGeoCode[ options.location ] = location = new google.maps.LatLng( options.lat, options.lng );
+
+          map = buildMap( options, innerdiv, that );
+        } else if ( status === google.maps.GeocoderStatus.OVER_QUERY_LIMIT ) {
```
humphd: Question: is there any other status code we might get?  Maybe we should add an else that console.warns it?

templates/assets/plugins/googlemap/popcorn.googlemap.js
```diff
@@ -169,26 +172,49 @@ var googleCallback;
       target.appendChild( outerdiv );
     }
 
+    function geoCodeCallback( results, status ) {
+      // second check for innerdiv since it could have disappeared before
+      // this callback is actual run
```
secretrobotron: "actual" -> "actually"

PR #1359, opened 2012-11-06
[#2836] Project Title should be in a disabled state while media ia chang...

mjschranz:
...ing

https://webmademovies.lighthouseapp.com/projects/65733/tickets/2836-header-fields-are-active-while-media-is-loading

### DIFF
--- a/src/ui/header.js
+++ b/src/ui/header.js
@@ -11,7 +11,7 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
         _userData = new UserData( butter, options ),
         _rootElement = Lang.domFragment( HEADER_TEMPLATE, ".butter-header" ),
         _webmakerNavBar = _rootElement.querySelector( "#webmaker-nav" ),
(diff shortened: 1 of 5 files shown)


src/ui/header.js
```diff
@@ -125,6 +140,19 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
         _previewBtn.style.display = "none";
         _projectTitle.style.display = "none";
         _saveButton.innerHTML = "Sign in to save";
+      },
+      mediaChange: function( on ) {
```
secretrobotron: For consistency with kate's original design here, can you just split this into `mediaChanging` and `mediaReady`?

css/header.less
```diff
@@ -133,6 +133,12 @@
   z-index: @MODAL_Z_INDEX;
 }
 
+.butter-header-disabled {
```
k88hudson: Is this actually necessary? Why not just have selectable and cursor stuff in the general `.butter-disabled` class?

PR #1361, opened 2012-11-06
[#2833] Change first run detection to use localStorage rather than cooki...

mjschranz:
...es

https://webmademovies.lighthouseapp.com/projects/65733/tickets/2833-first-run-dialog-appears-every-time-i-close-my-browser


src/core/project.js
```diff
@@ -2,16 +2,8 @@
...
-
-  // Local storage provider for backups.  All the browsers we support also
-  // support localStorage, but fail in such a way that we don't crash.
-  var __butterStorage = "localStorage" in window ?
-    window.localStorage :
-    { getItem: function(){},
-      setItem: function(){},
-      removeItem: function(){} };
+define( [ "core/eventmanager", "core/media", "util/shims" ],
+        function( EventManager, Media, Shims ){
```
ScottDowne: I do not see Shims being used in here. Can we remove it?

humphd: We don't need this at all, since main.js will load the shims for us, and they'll be there for the life of the app.  Yeah, remove all this shim stuff.

src/first-run.js
```diff
@@ -7,7 +7,7 @@
  *
  * Determines whether or not a user should be shown a first-run dialog
  */
-define( [ "dialog/dialog", "util/cookie", "ui/widget/tooltip" ], function( Dialog, Cookie, ToolTip ) {
+define( [ "dialog/dialog", "util/shims", "ui/widget/tooltip" ], function( Dialog, Shims, ToolTip ) {
```
humphd: Same deal.

src/util/shims.js
```diff
@@ -18,6 +18,17 @@ define([], function(){
     };
   }
 
+  // Shim our access to localStorage so it's not done in multiple places
+  if ( !window.__butterStorage ) {
+    // Local storage provider for backups.  All the browsers we support also
+    // support localStorage, but fail in such a way that we don't crash.
+    window.__butterStorage = "localStorage" in window ?
```
humphd: Move this stuff to a module (not the part from 26-30, which can stay here).  Basically, we want to have something like src/util/storage.js and then require Storage, which should be the same as your __butterStorage var above.

src/first-run.js
```diff
@@ -81,13 +81,17 @@ define( [ "dialog/dialog", "util/cookie", "ui/widget/tooltip" ], function( Dialo
...
-        setupFirstRun();
-        dialog = Dialog.spawn( "first-run" );
-        dialog.open( false );
-        dialog.listen( "close", onDialogClose );
-      }
+      try {
+        var data = __butterStorage.getItem( "butter-first-run" );
+
+        if ( !data ) {
+          __butterStorage.setItem( "butter-first-run", true );
```
humphd, line 91: One down side to this is that we'll always show first-run in browsers that don't support localStorage.  How many browsers is that, btw?

ScottDowne, line 91: Yeah, we should confirm this.

Confirm that the browsers we support all support local storage.

I believe we should be good for this, though.

mjschranz, line 91: http://caniuse.com/#feat=namevalue-storage

src/first-run.js
```diff
@@ -7,7 +7,10 @@
  *
  * Determines whether or not a user should be shown a first-run dialog
  */
-define( [ "dialog/dialog", "util/cookie", "ui/widget/tooltip" ], function( Dialog, Cookie, ToolTip ) {
+define( [ "dialog/dialog", "ui/widget/tooltip" ], function( Dialog, ToolTip ) {
```
secretrobotron: To illustrate the dependency tree a bit better, it's not a bad idea to include `util/shims` here. If, for some reason, this file needs to be used elsewhere, or `util/shims` isn't guaranteed to be included already, placing it in the `define` will ensure that it's loaded before continuing. Require will make sure it's only loaded once, so it's really just good practice as a software engineer.

secretrobotron: You should probably do the same to `src/core/project` as well.

src/first-run.js
ScottDowne, line 91: Yeah, that's a solid list.

PR #1363, opened 2012-11-06
[#2840] Googlemaps wasn't caching values in all situations. Caught some ...

mjschranz:
...missing circumstances

https://webmademovies.lighthouseapp.com/projects/65733/tickets/2840-map-bug

### DIFF
--- a/templates/assets/editors/googlemap/googlemap-editor.js
+++ b/templates/assets/editors/googlemap/googlemap-editor.js
@@ -161,6 +161,22 @@
           }, false );
         }
 
+        function updateLocation( te, prop ) {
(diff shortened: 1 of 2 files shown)


templates/assets/editors/googlemap/googlemap-editor.js
```diff
@@ -45,9 +47,8 @@
 
       var pluginOptions = {},
           ignoreKeys = [
-            "target",
-            "start",
-            "end"
+            "location",
+            "zoom"
```
secretrobotron: Why not just leave this as it was and extend the if/else block here https://github.com/mozilla/butter/pull/1363/files#L0L168 ?

PR #1364, opened 2012-11-06
T2819

secretrobotron:


cornfield/views/embed-shell.jade
```diff
@@ -23,7 +23,7 @@ html
   body
     iframe(id='embed', src='#{embedSrc}', width='1280', height='745', mozallowfullscreen='mozallowfullscreen', webkitallowfullscreen='webkitallowfullscreen', allowfullscreen='allowfullscreen')
     script
-      if ( location.search ) {
+      if ( location.search && location.search.search(/(\?|&)preview=true(\?|&|$)/) > -1 ) {
         var embed = document.querySelector('#embed');
-        embed.src = embed.src + location.search;
+        embed.src = embed.src + "?previewTime=" + Date.now();
```
humphd: The only danger I see here is if we ever add a query string to our template genearation (e.g., ?autoplay=1) then this will double-add the ? character.  Probably not a big deal, but something to consider.

cornfield/views/embed-shell.jade
```diff
@@ -23,7 +23,7 @@ html
   body
     iframe(id='embed', src='#{embedSrc}', width='1280', height='745', mozallowfullscreen='mozallowfullscreen', webkitallowfullscreen='webkitallowfullscreen', allowfullscreen='allowfullscreen')
     script
-      if ( location.search ) {
+      if ( location.search && location.search.search(/(\?|&)preview=true(\?|&|$)/) > -1 ) {
```
jbuck: I think you can remove the check for `?` at the end of the search string

PR #1370, opened 2012-11-07
[2850] Popcorn wrapper defensive checks on ready media, and turn off cus...

ScottDowne:
...tom input keys while media is not ready.

### DIFF
--- a/src/core/popcorn-wrapper.js
+++ b/src/core/popcorn-wrapper.js
@@ -464,14 +464,14 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
 
     // Passthrough to the Popcorn instances play method
     this.play = function(){
-      if ( _popcorn.paused() && _mediaReady ) {
+      if ( _mediaReady && _popcorn.paused() ) {
         _popcorn.play();
       }
     };
 
     // Passthrough to the Popcorn instances pause method
     this.pause = function(){
-      if ( !_popcorn.paused() && _mediaReady ) {
+      if ( _mediaReady && !_popcorn.paused() ) {
         _popcorn.pause();
       }
     };
@@ -575,7 +575,7 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
       currentTime: {
         enumerable: true,
         set: function( val ){
-          if( _popcorn && _mediaReady ){
+          if( _mediaReady && _popcorn ){
             _popcorn.currentTime( val );
           } //if
         },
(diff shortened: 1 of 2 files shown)


src/core/popcorn-wrapper.js
```diff
@@ -461,14 +464,14 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
 
     // Passthrough to the Popcorn instances play method
     this.play = function(){
-      if ( _popcorn.paused() ) {
+      if ( _popcorn.paused() && _mediaReady ) {
```
humphd: Flip the order of these calls, since we want to short-circuit if _mediaReady is false.

src/core/popcorn-wrapper.js
```diff
@@ -461,14 +464,14 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
...
         _popcorn.play();
       }
     };
 
     // Passthrough to the Popcorn instances pause method
     this.pause = function(){
-      if ( !_popcorn.paused() ) {
+      if ( !_popcorn.paused() && _mediaReady ) {
```
humphd: Same here.

src/core/popcorn-wrapper.js
```diff
@@ -572,7 +575,7 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
       currentTime: {
         enumerable: true,
         set: function( val ){
-          if( _popcorn ){
+          if( _popcorn && _mediaReady ){
```
humphd: and here

src/core/popcorn-wrapper.js
ScottDowne: I did this in this order, because _popcorn.paused is more likely to be false than the media is to be not ready, thus we'll break sooner, more often.

That was my thinking, anyway.

I was also under the impression checking media.paused is safe when the media is not loaded, no?

humphd: Since _popcorn can be null, and you're doing a property lookup to get at paused(), it's safer to do the check on the boolean first.  I'd flip them, yes.

src/core/popcorn-wrapper.js
```diff
@@ -609,10 +612,10 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
         set: function( val ){
           if( _popcorn ){
             if( val ){
-              _popcorn.pause();
```
humphd, line 612: Why this change?

src/core/popcorn-wrapper.js
```diff
@@ -609,10 +612,10 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
...
+              _this.pause();
             }
             else {
-              _popcorn.play();
```
humphd, line 615: Same question here..

PR #1371, opened 2012-11-07
[#2851] Make save button disabled until UI is ready. Share button should...

mjschranz:
... also be disabled in certain project states

https://webmademovies.lighthouseapp.com/projects/65733/tickets/2851-save-button-in-header-can-still-be-clicked-during-initial-loading


src/ui/header.js
```diff
@@ -301,6 +300,9 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
       if ( butter.project.name ) {
         _projectName.textContent = butter.project.name;
       }
+
+      // We don't want the save button to be enabled until the UI is ready.
+      toggleSaveButton( true );
```
ScottDowne: I don't think we need this.

PR #1376, opened 2012-11-14
[#2856] Re-add logic for browser detection using UAParser lib

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733/tickets/2856-ua-detection-regression

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -10,7 +10,7 @@
   var ACCEPTED_UA_LIST = {
     "Chrome": 17,
     "Firefox": 10,
-    "MSIE": 9,
+    "IE": 9,
     "Safari": 6,
     "Opera": 9
   };
@@ -22,7 +22,8 @@
             "util/xhr", "util/lang", "util/tutorial",
             "text!default-config.json", "text!layouts/ua-warning.html",
             "ui/widget/tooltip", "crashreporter", "core/project",
-            "util/shims"                  // keep this at the end so it doesn't need a spot in the function signature
+            // keep these at the end so they don't need a spot in the function signature
+            "UAParser/ua-parser", "util/shims"
           ],
           function(
             EventManager, Logger, Config, Target, Media, Page,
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -22,7 +22,7 @@
             "util/xhr", "util/lang", "util/tutorial",
             "text!default-config.json", "text!layouts/ua-warning.html",
             "ui/widget/tooltip", "crashreporter", "core/project",
-            "util/shims"                  // keep this at the end so it doesn't need a spot in the function signature
+            "UAParser/ua-parser", "util/shims"    // keep these at the end so they don't need a spot in the function signature
```
dseif: move `"util/shims"` down a line and put the comment above it as it's pretty long

src/main.js
```diff
@@ -22,7 +22,8 @@
             "util/xhr", "util/lang", "util/tutorial",
             "text!default-config.json", "text!layouts/ua-warning.html",
             "ui/widget/tooltip", "crashreporter", "core/project",
-            "util/shims"                  // keep this at the end so it doesn't need a spot in the function signature
+            "UAParser/ua-parser",
```
humphd: Put this on the same line as util/shims so we don't remove them or break things later.  I suggest:

``` javascript
"ui/widget/tooltip", "crashreporter", "core/project",

// keep these at the end so they don't need a spot in the function signature
"UAParser/ua-parser", "util/shims"
```

src/main.js
```diff
@@ -46,21 +49,23 @@
...
 
-      var ua = navigator.userAgent,
-          acceptedUA;
+      // ua-parser uses the current browsers UA by default
+      var ua = new UAParser().getResult(),
+          name = ua.browser.name,
+          major = ua.browser.major,
+          acceptedUA = false;
+
       for ( var uaName in ACCEPTED_UA_LIST ) {
```
humphd, line 65: for..in loops should wrap their body in a hasOwnProperty check.

src/main.js
dseif: Ah that is my fault. I told him to split up that line, sorry.

PR #1377, opened 2012-11-15
[#2881] Projects Table needed to explicitly define charset types for pro...

mjschranz:
...duction

https://webmademovies.lighthouseapp.com/projects/65733/tickets/2881-arabic-string-storageretrieval

### DIFF
--- a/cornfield/config/default.json
+++ b/cornfield/config/default.json
@@ -82,7 +82,11 @@
     "options": {
       "logging": false,
       "dialect": "sqlite",
-      "storage": "popcorn.sqlite"
+      "storage": "popcorn.sqlite",
+      "define": {
+        "charset": "utf8",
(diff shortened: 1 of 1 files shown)


cornfield/lib/models/project.js
```diff
@@ -51,5 +51,9 @@ module.exports = function(sequelize, DataTypes) {
       type: DataTypes.STRING,
       allowNull: false
     }
+  },
+  {
+    charset: "utf8",
+    collate: "utf8_general_ci"
```
jbuck: Instead of putting this in each model (project and image), you can put it in the db options globally: `define: { charset: 'utf8' }`

Check http://www.sequelizejs.com/#instances-configuration if it's not clear how, but basically move this bit of code into default.json.

PR #1378, opened 2012-11-15
[#2858] Video no longer reports a duration or currentTime of 1hr when it...

humphd:
... shouldn't. Videos with a duration of over 24 hours also work fine now

### DIFF
--- a/src/util/lang.js
+++ b/src/util/lang.js
@@ -70,22 +70,35 @@ define( [ "./shims" ], function(){
 
     secondsToSMPTE: function( time ) {
       var timeStamp = new Date( 0, 0, 0 ),
-          timeArr,
           days,
-          seconds;
+          hours,
+          minutes,
+          seconds,
+          timeString = "";
 
-      timeStamp.setSeconds( time );
+      timeStamp.setSeconds( +time );
+
+      // Because we can't rely on .toLocaleTimeString() on the Date object, getting the individual values
+      // for each portion of the time stamp and constructing it ourself should suffice.
       days = timeStamp.getDay();
+      hours = timeStamp.getHours();
+      minutes = timeStamp.getMinutes();
+      seconds = timeStamp.getSeconds();
 
-      seconds = timeStamp.toLocaleTimeString();
+      // If the video is for some reason longer than 24 hours, we need to make sure we handle it accordingly.
       if ( days > 0 ) {
-        timeArr = seconds.split( ":" );
-        timeArr[ 0 ] = +timeArr[ 0 ] + ( days * 24 );
-        seconds = timeArr.join( ":" );
+        hours = +hours + ( days * 24 );
       }
 
-      return seconds;
-    }, //secondsToSMPTE
+      [ hours, minutes, seconds ].forEach(function( element, index, array ) {
+        var trailingChar = ( index === array.length - 1 ) ? "" : ":";
+
+        timeString += element >= 10 ? element : "0" + element;
+        timeString += trailingChar;
+      });
+
+      return timeString;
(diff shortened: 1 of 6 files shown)


src/util/lang.js
```diff
@@ -68,14 +68,22 @@ define( [ "./shims" ], function(){
...
+          days,
           seconds;
+
       timeStamp.setSeconds( time );
-      seconds = timeStamp.toTimeString().substr( 0, 8 );
-      if( seconds > 86399 ){
-        seconds = Math.floor( (timeStamp - Date.parse("1/1/70") ) / 3600000) + seconds.substr(2);
+      days = timeStamp.getDay();
+
+      seconds = timeStamp.toLocaleTimeString();
```
humphd: I think this is still liable to break.  From MDN:

"The toLocaleTimeString method relies on the underlying operating system in formatting dates. It converts the date to a string using the formatting convention of the operating system where the script is running. For example, in the United States, the month appears before the date (04/15/98), whereas in Germany the date appears before the month (15.04.98).

Methods such as getHours, getMinutes, and getSeconds give more consistent results than toLocaleTimeString. Use toLocaleTimeString when the intent is to display to the user a string formatted using the regional format chosen by the user. Be aware that this method, due to its nature, behaves differently depending on the operating system and on the user's settings.
"

I wonder if we should getHours, getMinutes, getSeconds and then concat with ':' ourselves.

src/util/lang.js
```diff
@@ -68,14 +68,22 @@ define( [ "./shims" ], function(){
       }
     }, //smpteToSeconds
 
-    secondsToSMPTE: function( time ){
-      var timeStamp = new Date( 1970, 0, 1 ),
+    secondsToSMPTE: function( time ) {
+      var timeStamp = new Date( 0, 0, 0 ),
```
humphd: `var timeStamp = new Date( 0 );`

dseif: I don't think we can do that as it doesn't return what we want.

```
var a = new Date( 0 );
// returns 19 for me
a.getHours();

var b = new Date( 0, 0, 0 );
// returns 0 for me
b.getHours();
```

src/util/lang.js
```diff
@@ -68,16 +68,26 @@ define( [ "./shims" ], function(){
...
+      timeStamp.setSeconds( +time );
+      days = timeStamp.getDay();
+      timeString = timeStamp.toTimeString().split( " " )[ 0 ];
+
+      // If the video is for some reason longer than 24 hours, we need to make sure we handle it accordingly.
+      if ( days > 0 ) {
+        (function( timeArr ) {
+          timeArr[ 0 ] = +timeArr[ 0 ] + ( days * 24 );
+          timeString = timeArr.join( ":" );
+        })( timeString.split( ":" ) );
```
secretrobotron: There's a lint error here: `src/util/lang.js: line 86, col 11, Move the invocation into the parens that contain the function.`

While I do _love_ the creativity you employed here to introduce and utilize the `timeArr` variable, this approach is strictly more expensive than just storing `timeArr` explicitly with a  `var` statement.

This is interesting though: what happens if you have something that's longer than a day? Do the hours/minutes/seconds make sense still?

dseif: Well considering the amount of times a user will actually have a video that is longer than 24 hours is probably very, very low, would creating a variable for `timeArr` be worth it? Seems like creating a variable for something that will happen 1 in every 1000 vidoes or something would be more of a waste than just taking this approach.

Yes, the the hours:minutes:seconds make sense when the video is over 24 hours ( shows up like 48:12:44 ).

PR #1381, opened 2012-11-19
[#2888] Class list shims were not added yet, causing the app to crash in...

mjschranz:
... IE9 before even loading

https://webmademovies.lighthouseapp.com/projects/65733/tickets/2888-ie9-unable-to-get-value-of-the-property-add-object-is-null-or-undefined

### DIFF
--- a/templates/basic/template.js
+++ b/templates/basic/template.js
@@ -4,8 +4,6 @@
 
 ( function( Butter, EditorHelper ) {
   document.addEventListener( "DOMContentLoaded", function() {
-    document.querySelector( "#embed-wrapper" ).classList.add( "faded" );
-
     Butter.init({
       config: "config.json",
       ready: function( butter ) {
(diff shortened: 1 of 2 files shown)


templates/basic/template.js
```diff
@@ -4,7 +4,9 @@
 
 ( function( Butter, EditorHelper ) {
   document.addEventListener( "DOMContentLoaded", function() {
-    document.querySelector( "#embed-wrapper" ).classList.add( "faded" );
+    var embedWrapper = document.querySelector( "#embed-wrapper" );
+
+    embedWrapper.className += " faded";
```
secretrobotron: Can you use `classList` instead?

secretrobotron: If the shim isn't ready yet, add a comment here to explain. Perhaps this line can be moved down inside the `init` function instead, when we know the shim exists.

PR #1383, opened 2012-11-19
[#2896] destroy and hide first run popups using the tooltip api to avoid...

cadecairos:
... errors if the callback somehow executes twice


src/first-run.js
```diff
@@ -37,9 +37,12 @@ define( [ "dialog/dialog", "ui/widget/tooltip", "util/shims" ], function( Dialog
           hidden: false
         });
 
+        popupTooltip = ToolTip.get( "tooltip-popup" );
```
secretrobotron, line 40: Is it not reliable enough to just accept the object returned by the contructor above?

cadecairos, line 40: that's a flaw in the tooltip system. 

IIRC, I kept it similar to the old tooltip setup which returned the tooltip element and not the object created. probably so It wouldn't break code elsewhere.

perhaps we should just change that?

PR #1385, opened 2012-11-20
[#2902] Copy and paste functionality with ctrlc and ctrlv

ScottDowne:


src/main.js
```diff
@@ -245,6 +246,57 @@
...
+        var newTrackEvent,
+            start,
+            end,
+            popcornOptions,
+            offset = 0,
+            track;
+        // get the first events start time to compare with the current time,
+        // to find the paste offset.
+        if ( _copiedEvents[ 0 ] ) {
+          _this.deselectAllTrackEvents();
```
k88hudson: This will deselect everything even if the track event can't be pasted... could that ever happen? What if you try to paste it at the very end of the video

src/main.js
```diff
@@ -245,6 +246,57 @@
...
+        if ( _sortedSelectedEvents.length ) {
+          _copiedEvents = [];
+          for ( var i = 0; i < _sortedSelectedEvents.length; i++ ) {
+            _copiedEvents.unshift( _sortedSelectedEvents[ i ].copy() );
+          }
+        }
+      };
+
+      _this.pasteTrackEvents = function() {
+        var newTrackEvent,
```
k88hudson: You're not using this anywhere

src/main.js
ScottDowne: Just pasting at the end of the video is an issue here, and only if you're right at the end, otherwise, you are going to paste something.

src/core/trackevent.js
```diff
@@ -207,6 +207,32 @@ define( [ "./logger", "./eventmanager", "./observer",
       _track = null;
     };
 
+    /**
+     * Member: copy
+     *
+     * Returns a copy of the data needed to create a track event just like this one.
+     * Does not copy state or unique data.
+     */
+    this.copy = function() {
```
jbuck, line 216: I don't know if this is a problem or not, but this will only do a shallow copy, not a deep copy.

ScottDowne, line 216: It is only copying the track event's data. On a paste, it makes new track events out of that data.

Track events don't really have a use for references in their data. They consider data to just be data. They never use a reference to an object in said data.

We're good here.

src/main.js
```diff
@@ -245,6 +246,49 @@
...
+          _this.deselectAllTrackEvents();
+          offset = _currentMedia.currentTime - _copiedEvents[ 0 ].popcornOptions.start;
+          for ( var i = 0; i < _copiedEvents.length; i++ ) {
+            popcornOptions = {};
+            for ( var prop in _copiedEvents[ i ].popcornOptions ) {
+              if ( _copiedEvents[ i ].popcornOptions.hasOwnProperty( prop ) ) {
+                popcornOptions[ prop ] = _copiedEvents[ i ].popcornOptions[ prop ];
+              }
+            }
+            start = popcornOptions.start = popcornOptions.start + offset;
```
jbuck: Could you just use `popcornOptions.start/end` instead of separate `start/end` variables? Seems like duplication.

src/main.js
```diff
@@ -532,6 +576,21 @@
           },
           enumerable: true
         },
+        sortedCopiedEvents: {
+          get: function() {
+            // _copiedEvents is already sorted,
+            // but we define both a sorted and regular
+            // for parity with the rest of the API.
```
jbuck: That's just silly...

src/ui/ui.js
```diff
@@ -526,7 +525,19 @@ define( [ "core/eventmanager", "./toggler",
           butter.deselectAllTrackEvents();
           orderedTrackEvents[ index ].selected = true;
         } // if
-      } // tab key
+      }, // tab key
+
+      67: function( e ) { // c key
+        if ( e.ctrlKey || e.metaKey ) {
```
jbuck, line 531: I wish there was a better way of implementing the "use the command key on OS X, and the control key everywhere else" behaviour

PR #1390, opened 2012-11-20
[#2894] Googlemaps Update Function

cadecairos:
Lighthouse Ticket: https://webmademovies.lighthouseapp.com/projects/65733-popcorn-maker/tickets/2894

### DIFF
--- a/public/src/editor/trackevent-editor.js
+++ b/public/src/editor/trackevent-editor.js
@@ -468,7 +468,7 @@ define([ "util/lang", "util/keys", "util/time", "./base-editor", "ui/widget/tool
         if ( e.keyCode === KeysUtils.ENTER ) {
           if ( !e.shiftKey ) {
             e.preventDefault();
-            
+
             if ( isNumber ) {
               val = validateNumber( val );
             }
@@ -674,6 +674,10 @@ define([ "util/lang", "util/keys", "util/time", "./base-editor", "ui/widget/tool
         }
         try {
           editorElement.type = manifestEntry.type;
+          // step="any" will stop the :invalid pseudo class in Chrome from being applied if the value is a not a "whole" number. i.e. 1.234
+          if ( editorElement.type === "number" ) {
+            editorElement.step = manifestEntry.step || "any";
+          }
         }
         catch ( e ) {
           // Suppress IE9 errors
--- a/public/templates/assets/editors/googlemap.js
+++ b/public/templates/assets/editors/googlemap.js
@@ -1,11 +1,9 @@
-/*global EditorHelper,google,clearTimeout,setTimeout*/
+/*global EditorHelper,google*/
 
-EditorHelper.addPlugin( "googlemap", function( trackEvent, popcornInstance ) {
-  var UPDATE_TIMEOUT = 250;
+EditorHelper.addPlugin( "googlemap", function( trackEvent, popcorn ) {
 
   var container,
       target,
-      popcorn = popcornInstance,
       popcornEventMapReference;
 
   function setup() {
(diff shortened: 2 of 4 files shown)


src/core/popcorn-wrapper.js
```diff
@@ -66,6 +66,7 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
     }; //interrupt
 
     // Update Popcorn events with data from a butter trackevent
+
```
mjschranz: Remove this.

src/default-config.json
```diff
@@ -9,7 +9,7 @@
   "title": "Popcorn Maker",
   "ui": {
     "enabled": true,
-    "onLeaveDialog": true,
+    "onLeaveDialog": false,
```
mjschranz: Why was this changed?

templates/assets/editors/googlemap.js
```diff
@@ -26,76 +25,68 @@ EditorHelper.addPlugin( "googlemap", function( trackEvent, popcornInstance ) {
...
+
+          updateOptions.heading = pov.heading;
+          updateOptions.pitch = pov.pitch;
+          updateOptions.zoom = pov.zoom;
+          if ( latlng ) {
+            updateOptions.lat = latlng.lat();
+            updateOptions.lng = latlng.lng();
+          }
+
+          trackEvent.update( updateOptions);
```
mjschranz: Space.

src/editor/trackevent-editor.js
```diff
@@ -664,6 +664,9 @@ define([ "util/lang", "util/keys", "util/time", "./base-editor", "ui/widget/tool
         }
         try {
           editorElement.type = manifestEntry.type;
+          if ( editorElement.type === "number" ) {
+            editorElement.step = "any";
+          }
```
mjschranz: It doesn't look like you use this code anywhere. Can you explain it's purpose?

templates/assets/plugins/googlemap/popcorn.googlemap.js
```diff
@@ -446,6 +448,121 @@ var googleCallback;
...
+              if ( status === google.maps.GeocoderStatus.OK ) {
+                options.lat = results[ 0 ].geometry.location.lat();
+                options.lng = results[ 0 ].geometry.location.lng();
+                _cachedGeoCode[ options.location ] = location = new google.maps.LatLng( options.lat, options.lng );
+                map.panTo( location );
+              }
+            });
+          }
+        }
+
```
mjschranz: Extra line.

templates/assets/plugins/googlemap/popcorn.googlemap.js
```diff
@@ -446,6 +448,121 @@ var googleCallback;
...
+          trackEvent.top = options.top;
+          trackEvent._container.style.top = trackEvent.top + "%";
+        }
+
+        if ( options.height && options.height !== trackEvent.height ) {
+          trackEvent.height = options.height;
+          trackEvent._container.style.height = trackEvent.height + "%";
+        }
+
+        if ( options.width && options.width !== trackEvent.width ) {
```
mjschranz: You do checks to ensure top/left aren't under zero. Why not the same for width/height?

src/default-config.json
cadecairos: woops

src/editor/trackevent-editor.js
cadecairos: That fixes inputs with number types in the editor. without it any value not divisible by one is shown as invalid.

templates/assets/plugins/googlemap/popcorn.googlemap.js
cadecairos: I don't see why anyone would want height or width to be zero... but I can check anyways?

templates/assets/editors/googlemap.js
```diff
@@ -26,76 +25,100 @@ EditorHelper.addPlugin( "googlemap", function( trackEvent, popcornInstance ) {
...
+            streetViewTimeout,
+            zoomTimeout;
+
+        function locationChanged() {
+
+          if ( locationTimeout ) {
+            clearTimeout( locationTimeout );
+          }
+
+          setTimeout(function() {
```
mjschranz: This should be setting the timeouts id to `locationTimeout`.

templates/assets/editors/googlemap.js
```diff
@@ -26,76 +25,108 @@ EditorHelper.addPlugin( "googlemap", function( trackEvent, popcornInstance ) {
...
-    if ( !trackEvent.popcornTrackEvent._map ) {
-      trackEvent.popcornTrackEvent.onmaploaded = function( options, map ) {
-        popcornEventMapReference = map;
+      if ( !trackEvent.popcornTrackEvent._map ) {
+        trackEvent.popcornTrackEvent.onmaploaded = function( options, map ) {
+          popcornEventMapReference = map;
+          setupMapListeners();
+        };
+      }
+      else {
```
mjschranz: Small nit. I know that it was like this before but ideally we could bump this onto the line above so we gradually keep our code style in check.

src/editor/trackevent-editor.js
jbuck: So this basically handles the `0` case? If so, please add a comment to that effect because it's not obvious at all

cadecairos: The specific reason for this case was heading and pitch in Google Maps. they can be set to numbers like `23.812648112311` and without the `step="any"` chrome defaults to `step="1"`, which only allows for "whole" numbers like 1, 2, 3..

cadecairos: Might be worth noting that the step attribute isn't implemented in any other browsers. Maybe in Opera, though I'm not entirely sure.

templates/assets/editors/googlemap.js
```diff
@@ -1,7 +1,6 @@
-/*global EditorHelper,google,clearTimeout,setTimeout*/
+/*global EditorHelper,google*/
 
 EditorHelper.addPlugin( "googlemap", function( trackEvent, popcornInstance ) {
```
jbuck: Instead of doing `var popcorn = popcornInstance`, change this variable name to `popcorn`

templates/assets/editors/googlemap.js
jbuck: Yep, it'd be nice to fix this

templates/assets/editors/googlemap.js
```diff
@@ -26,76 +25,108 @@ EditorHelper.addPlugin( "googlemap", function( trackEvent, popcornInstance ) {
...
-      setup();
-    }
+        setup();
+      }
 
-    if ( !trackEvent.popcornTrackEvent._map ) {
-      trackEvent.popcornTrackEvent.onmaploaded = function( options, map ) {
-        popcornEventMapReference = map;
+      if ( !trackEvent.popcornTrackEvent._map ) {
+        trackEvent.popcornTrackEvent.onmaploaded = function( options, map ) {
```
jbuck: Maybe I'm reading this wrong, but aren't you already listening for this with `popcorn.on( "googlemaps-loaded")`, so this is actually a dead code path?

templates/assets/plugins/googlemap/popcorn.googlemap.js
```diff
@@ -217,19 +184,63 @@ var googleCallback;
...
+
+        options.toString = function() {
+          if ( options.type === "STREETVIEW" ) {
+            return "Streeview";
+          }
+          return options.location || ( ( options.lat && options.lng ) ? options.lat + ", " + options.lng : options._natives.manifest.options.location[ "default" ] );
+        };
+      },
       /**
        * @member webpage
```
jbuck: what is this i don't even

PR #1391, opened 2012-11-21
T2875

ScottDowne:


src/core/popcorn-wrapper.js
```diff
@@ -574,9 +574,17 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
...
-          if( _mediaReady && _popcorn ){
-            _popcorn.currentTime( val );
+        set: function( val ) {
+          if ( _mediaReady && _popcorn ) {
+            try {
+              _popcorn.currentTime( val );
+            } catch ( e ) {
+              // IE9 throws this error when trying to seek outside of buffered range.
+              // Happens with mp4.
+              if ( e.message !== "DOM Exception: INDEX_SIZE_ERR (1)" ) {
```
secretrobotron, line 584: I'm not sure if this error is persistent across all browsers. However, `e.name` should be `INDEX_SIZE_ERR`, `e.code` should be `11`, and `e instanceof DOMException` should be `true`.

PR #1395, opened 2012-11-26
[#2189] Functionally complete WIP for undo redo.

ScottDowne:


src/ui/ui.js
```diff
@@ -281,8 +281,8 @@ define( [ "core/eventmanager", "./toggler",
         popcornOptions.end = ( overlappingTrackEventElement.offsetLeft - overlappingTrackEventElement.clientLeft * 2 ) /
           butter.timeline.getCurrentTrackWidth() * butter.currentMedia.duration;
       }
-
-      trackEvent.update( popcornOptions );
+// do something about this false!!!! object?
```
humphd, line 289: Fix this, file a bug to fix this, but don't leave it like this.

ScottDowne, line 289: Totally forgot!

I knew I left a reminder there for a reason. This is for me to sort out, here an now, I think.

I usually drop comments on the far left if I don't want them to stay. They stick out more.

PR #1396, opened 2012-11-29
no bug - clean up track event interaction.

ScottDowne:


src/timeline/media.js
```diff
@@ -99,35 +100,53 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
       }
     }
 
-    function onTrackEventMouseOut( e ){
+    function onTrackEventMouseOut( e ) {
```
jbuck: Why create an empty function?

src/timeline/media.js
```diff
@@ -99,35 +100,53 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
...
     }
 
-    function onTrackEventDragStarted( e ){
+    function onTrackEventDragStarted( e ) {
       _currentMouseDownTrackEvent = null;
+      _currentMouseDownTrackEventWasSelected = false;
     }
 
-    function onTrackEventMouseDown( e ){
+    function onTrackEventMouseDown( e  ) {
```
jbuck: Spacing after the e

src/timeline/media.js
ScottDowne: No idea, actually. It was always there, I just fixed up the whitespace because I did for the other events that go hand in hand with it.

I cleaned up the other events whitespace because I touched them.

Personally, I don't mind if it serves no purpose other than completeness.

There was an empty onTrackEventMouseUp before I did this patch, so I just used it. I appreciated it, actually.

The real reason it is there is a mystery!

src/timeline/media.js
```diff
@@ -99,35 +100,50 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
...
+    function onTrackEventMouseDown( e ) {
       var trackEvent = e.data.trackEvent,
-          originalEvent = e.data.originalEvent,
-          isHandle = originalEvent.target.classList.contains( "handle" );
+          tracks,
+          originalEvent = e.data.originalEvent;
 
+      // track the mousedown event and previous selected state for mouseup.
       _currentMouseDownTrackEvent = trackEvent;
+      _currentMouseDownTrackEventWasSelected = trackEvent.selected;
```
secretrobotron: Do you need to variables, or can you just use `_currentMouseDownTrackEvent`? When `mouseup` occurs, `trackEvent.selected` should still be what it was on `mousedown`, right?

src/timeline/media.js
```diff
@@ -99,35 +100,50 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
...
-        trackEvent.selected = true;
-      }
-
-      if( !originalEvent.shiftKey && !isHandle ){
-        var tracks = _media.tracks;
-        for( var t in tracks ){
-          if( tracks.hasOwnProperty( t ) ){
+      if ( !originalEvent.shiftKey && !trackEvent.selected ) {
+        tracks = _media.tracks;
+        for ( var t in tracks ) {
```
secretrobotron: Not sure why this is a `for in` loop, but it doesn't need to be. Can you make this an index-based hot-loop (`for i = 0 ... length`)?

src/timeline/media.js
```diff
@@ -99,35 +100,50 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
...
-    function onTrackEventMouseUp( e ){
+    function onTrackEventMouseUp( e ) {
+      var trackEvent = e.data.trackEvent,
+          tracks,
+          originalEvent = e.data.originalEvent;
+
+      if ( trackEvent === _currentMouseDownTrackEvent ) {
+        if ( !originalEvent.shiftKey ) {
+          tracks = _media.tracks;
+          for ( var t in tracks ) {
```
secretrobotron: Same as https://github.com/mozilla/butter/pull/1396/files#L1R140

src/timeline/media.js
```diff
@@ -156,7 +172,6 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
         trackEvent.view.unlisten( "trackeventmousedown", onTrackEventMouseDown );
         if( _trackEventHighlight === "hover" ){
           trackEvent.view.unlisten( "trackeventmouseover", onTrackEventMouseOver );
-          trackEvent.view.unlisten( "trackeventmouseout", onTrackEventMouseOut );
```
secretrobotron, line 159: I'm fairly certain nothing uses `trackeventmouseout` anywhere in the codebase. Perhaps its whole presence should be eradicated.

src/timeline/media.js
```diff
@@ -99,35 +100,50 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
...
+      if ( !originalEvent.shiftKey && !trackEvent.selected ) {
+        tracks = _media.tracks;
+        for ( var t in tracks ) {
+          if ( tracks.hasOwnProperty( t ) ) {
             tracks[ t ].deselectEvents( trackEvent );
           }
         }
       }
+      trackEvent.selected = true;
     }
```
secretrobotron, line 138: Might it be easier to just listen to `window.onmouseup` here instead of using `trackevent.listen('trackeventmouseup')`? If you mouse down on a trackevent, assigning the listener here would just let you capture the trackevent in a closure instead of using any member variables, right?

src/timeline/media.js
ScottDowne: I set selected to true at the end of onTrackEventMouseDown.

I want to know, on the up, if it was selected before we went in here, if it was not, I apply different logic.

src/timeline/media.js
ScottDowne: Can do.

src/timeline/media.js
ScottDowne, line 159: We should probably just file it, but I have no problems doing it.

src/timeline/media.js
```diff
@@ -99,35 +98,47 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
...
+            tracks[ i ].deselectEvents( trackEvent );
           }
+        } else if ( trackEvent.selected && wasSelected ) {
+          trackEvent.selected = false;
         }
-      }
+      };
+
+      onTrackEventDragStarted = function() {
+
```
secretrobotron, line 130: We don't need whitespace at the beginning of any scope, right?

src/timeline/media.js
```diff
@@ -99,35 +98,47 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
...
+          originalEvent = e.data.originalEvent;
+
+      wasSelected = trackEvent.selected;
+
+      if ( !originalEvent.shiftKey && !trackEvent.selected ) {
+        tracks = _media.tracks;
+          for ( i = 0, length = tracks.length; i < length; i++ ) {
+          tracks[ i ].deselectEvents( trackEvent );
+        }
+      }
```
secretrobotron, line 113: Add a "\n" here.

src/timeline/media.js
```diff
@@ -99,35 +98,47 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
...
-      _currentMouseDownTrackEvent = null;
-    }
-
-    function onTrackEventMouseDown( e ){
+    function onTrackEventMouseDown( e ) {
       var trackEvent = e.data.trackEvent,
-          originalEvent = e.data.originalEvent,
-          isHandle = originalEvent.target.classList.contains( "handle" );
+          tracks, i, length,
+          wasSelected = false,
```
secretrobotron: No need to initialize `wasSelected` since it's set immediately here https://github.com/mozilla/butter/pull/1396/files#L1R109

src/timeline/media.js
```diff
@@ -99,35 +98,47 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
...
+      if ( !originalEvent.shiftKey && !trackEvent.selected ) {
+        tracks = _media.tracks;
+          for ( i = 0, length = tracks.length; i < length; i++ ) {
+          tracks[ i ].deselectEvents( trackEvent );
+        }
+      }
+      trackEvent.selected = true;
 
-      _currentMouseDownTrackEvent = trackEvent;
+      onTrackEventMouseUp = function() {
```
secretrobotron: Is there a reason these aren't just declared like `function onTrackEventMouseUp()`?

ScottDowne: Just habit, really. I'll change up.

PR #1398, opened 2012-11-30
[#2937] Hover works again for track event highlighting.

mjschranz:
Old event listener code was expecting data in the wrong format

https://webmademovies.lighthouseapp.com/projects/65733-popcorn-maker/tickets/2937-hover-option-for-trackevent-highlighting-breaks-the-app


src/timeline/media.js
```diff
@@ -89,16 +89,17 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
       }
     }
 
-    function onTrackEventMouseOver( e ){
-      var trackEvent = e.trackEvent,
-          corn = trackEvent.popcornOptions;
+    function onTrackEventMouseOut( e ){
```
secretrobotron: This probably don't need to be here.

src/timeline/media.js
```diff
@@ -89,16 +89,17 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
...
+    }
 
-      if( corn.target ){
-        blinkTarget( corn.target );
-      }
+    function onTrackEventMouseUp( e ){
```
secretrobotron: Neither does this.

PR #1401, opened 2012-12-06
[#1925] Add delete-track-events dialog.

cadecairos:
Lighthouse Ticket: https://webmademovies.lighthouseapp.com/projects/65733-popcorn-maker/tickets/1925

### DIFF
--- a/src/dialog/dialogs/delete-track-events.html
+++ b/src/dialog/dialogs/delete-track-events.html
@@ -1,0 +1,12 @@
+<!--  This Source Code Form is subject to the terms of the MIT license
+      If a copy of the MIT license was not distributed with this file, you can
+      obtain one at https://raw.github.com/mozilla/butter/master/LICENSE -->
+
+<div class="butter-dialog butter-form small">
+  <p><strong>Deleting all track events cannot be undone. Do you wish to continue?</strong></p>
+  <div class="buttons vbox center">
+    <button class="butter-btn btn-light butter-dialog-button yes">Yes</button>
+    <button class="butter-btn btn-light butter-dialog-button no">No</button>
+  </div>
+  <a class="close-button"><span class="icon icon-only icon-x"></span></a>
+</div>
--- a/src/editor/media-editor.js
+++ b/src/editor/media-editor.js
@@ -2,8 +2,8 @@
  * If a copy of the MIT license was not distributed with this file, you can
  * obtain one at https://raw.github.com/mozilla/butter/master/LICENSE */
 
-define( [ "util/lang", "util/uri", "util/keys", "editor/editor", "text!layouts/media-editor.html" ],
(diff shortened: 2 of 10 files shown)


src/dialog/dialogs/delete-track-events.html
```diff
@@ -0,0 +1,12 @@
+<!--  This Source Code Form is subject to the terms of the MIT license
+      If a copy of the MIT license was not distributed with this file, you can
+      obtain one at https://raw.github.com/mozilla/butter/master/LICENSE -->
+
+<div class="butter-dialog butter-form small">
+  <p>Would you like to delete all track events?</p>
```
mjschranz: I think a more appropriate message might be along the lines of `Would you like all track events to be removed with the media change as well?`

That might not be that great but just something to distinguish this from our normal trackevent deletion dialog.

src/editor/media-editor.js
```diff
@@ -194,11 +194,27 @@ define( [ "util/lang", "util/uri", "util/keys", "editor/editor", "text!layouts/m
...
-  function onMediaReady() {
+  function finishMediaLoad() {
     showError( false );
     setLoadSpinner( false );
   }
 
+  function onMediaReady() {
+    var dialog;
+
+    if ( Butter.app.currentMedia.hasTrackEvents() ) {
```
mjschranz: You could just store the butter instance you receive on line 218.

src/editor/media-editor.js
```diff
@@ -194,7 +201,20 @@ define( [ "util/lang", "util/uri", "util/keys", "editor/editor", "text!layouts/m
...
+    var dialog = Dialog.spawn( "delete-track-events", {
+      data: _butter
+    });
+
+    dialog.open();
+  }
+
   function onMediaReady() {
+    if ( _clearEventsOnReady ) {
+      spawnClearDialog();
```
ScottDowne: Why make this a function? Why not just put the code in this block?

src/editor/media-editor.js
```diff
@@ -194,7 +201,20 @@ define( [ "util/lang", "util/uri", "util/keys", "editor/editor", "text!layouts/m
     showError( true );
   }
 
+  function spawnClearDialog() {
+    _clearEventsOnReady = false;
```
ScottDowne: Why set this to false?

Is the checkbox deselected? If not, is there a reason for this?

You can make it so this value only gets changed when the checkbox is clicked.

src/editor/media-editor.js
```diff
@@ -194,7 +201,20 @@ define( [ "util/lang", "util/uri", "util/keys", "editor/editor", "text!layouts/m
...
+
+    var dialog = Dialog.spawn( "delete-track-events", {
+      data: _butter
+    });
+
+    dialog.open();
+  }
+
   function onMediaReady() {
+    if ( _clearEventsOnReady ) {
```
ScottDowne: Also check that we still have track events here.

PR #1406, opened 2012-12-10
[#2950] Editors close when track event is deselected. Also, editors open...

ScottDowne:
... when selected, and not just when clicked. Example, tab changes the active editor/closes it now too.

### DIFF
--- a/src/timeline/media.js
+++ b/src/timeline/media.js
@@ -151,8 +151,8 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
       function onTrackEventAdded( e ){
         var trackEvent = e.data;
         trackEvent.view.listen( "trackeventmousedown", onTrackEventMouseDown );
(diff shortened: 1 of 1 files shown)


src/timeline/media.js
```diff
@@ -136,14 +144,15 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
...
       });
 
       function onTrackEventAdded( e ){
         var trackEvent = e.data;
         trackEvent.view.listen( "trackeventmousedown", onTrackEventMouseDown );
-        trackEvent.view.element.addEventListener( "click", function( e ) {
-          butter.editor.editTrackEvent( trackEvent );
-        });
+        trackEvent.listen( "trackeventselected", onTrackEventSelected);
+        trackEvent.listen( "trackeventdeselected", onTrackEventDeselected);
```
mjschranz: Spacing on these two.

PR #1408, opened 2012-12-10
Bug 2914 - Upgrade to RequireJS v2.1.2 and Almond v0.2.2

jbuck:


jbuck:
This looks like a large patch, but it's actually pretty small. It just looks like a large change in `src/butter.js` because I combined `src/butter.js`, `src/config.js`, and `src/main.js` into one file. The changes are located at the top, where I create a `window.Butter` shim, and around line 50 where the requirejs config now lives.

The new config is pretty cool too; instead of having to list paths and dependencies in the entry point and the optimization config, it's all in the entry point now. And it's _finally_ a single entry point for butter.

Other changes include:
- `external/require/`, `external/ua-parser/`, and `tools/almond.js` are just version upgrades.
- Renamed optimization files to be more consistent.

npm-shrinkwrap.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "butter",
-  "version": "1.0.8",
+  "version": "1.0.15",
```
humphd: Obviously we're going to be out of sync now with the length of time this review has taken.  When you rebase, this will get fixed.

src/butter.js
```diff
@@ -1,42 +1,1027 @@
-/*! This Source Code Form is subject to the terms of the MIT license
```
humphd: You're sure we don't need this any more?  That syntax forces Require to not rip out the comment block.  As I recall, we had moved away from injecting a header, and relied on this happening.  I'll test later, but maybe you know for sure.

src/butter.js
```diff
@@ -1,42 +1,1027 @@
...
+/* This Source Code Form is subject to the terms of the MIT license
  * If a copy of the MIT license was not distributed with this file, you can
  * obtain one at https://raw.github.com/mozilla/butter/master/LICENSE */
 
-/*jshint evil: true */
+// Since the template.js will be ready before Butter, prepare a shim
```
humphd: s/Since the/Since /

src/butter.js
```diff
@@ -1,42 +1,1027 @@
...
+      init();
+    };
+    rscript.src = "/external/require/require.js";
+    document.head.appendChild(rscript);
+    return;
+  }
+
+  var WARNING_WAIT_TIME = 500;
+
+  var ACCEPTED_UA_LIST = {
```
humphd: We should really move this into the config file...(another bug obviously).

src/butter.js
```diff
@@ -1,42 +1,1027 @@
...
+    "Chrome": 17,
+    "Firefox": 10,
+    "IE": 9,
+    "Safari": 6,
+    "Opera": 12
+  },
+
+  MOBILE_OS_BLACKLIST = [
```
humphd, line 38: Same there.

src/butter.js
```diff
@@ -1,42 +1,1027 @@
...
+    deps: [ "util/shims", "jsSHA" ],
+    paths: {
+      'text': '../external/require/text',
+      'jsSHA': '../external/jsSHA/sha1',
+      'UAParser': '../external/ua-parser/ua-parser'
+    }
+  });
+
+  require( [
+            "core/eventmanager", "core/logger", "core/config", "core/track",
```
humphd, line 60: My only fear with this style of multiple-items-per-line is that it's easy to miss seeing one, and double add it, remove it, or otherwise mess up this order.  The file is already massive, why not split these 1-per-line, and same with function signature below, to match?

src/config.js
```diff
@@ -1,21 +0,0 @@
-/* This Source Code Form is subject to the terms of the MIT license
- * If a copy of the MIT license was not distributed with this file, you can
- * obtain one at https://raw.github.com/mozilla/butter/master/LICENSE */
-
-define(function( require, exports, module ){
-
-  //Use the global requirejs to create a new context.
-  var ctx = requirejs.config({
-    context: "butter",
-    baseUrl: module.uri.substring(0, module.uri.lastIndexOf('/')),
```
humphd, line 10: So good to get rid of this!

src/butter.js
jbuck: I'd love to move away from "accepted UAs" to "your browser doesn't support these features, so it won't work" too

cadecairos: +1 We really should move to feature detection over UA sniffing.

PR #1411, opened 2012-12-12
[#2957] Upgrade to jshint 0.9.1 and fixed all lint errors from doing so

mjschranz:


package.json
```diff
@@ -83,7 +83,7 @@
   },
   "dependencies": {
     "shelljs": "~0.0.9",
-    "jshint": "0.6.2",
+    "jshint": "0.9.1",
```
jbuck: Use the "~0.9.1" syntax here. You'll also need to regenerate the shrinkwrap file using `npm shrinkwrap` once you've installed v0.9.1.

src/core/project.js
```diff
@@ -2,8 +2,8 @@
  * If a copy of the MIT license was not distributed with this file, you can
  * obtain one at https://raw.github.com/mozilla/butter/master/LICENSE */
 
-define( [ "core/eventmanager", "core/media", "util/shims" ],
```
jbuck, line 5: I think you'll need to leave "util/shims" in the list here; just loading it adds the shims properly. You don't need the Shims var though.

src/cornfield/module.js
```diff
@@ -2,8 +2,8 @@
  * If a copy of the MIT license was not distributed with this file, you can
  * obtain one at https://raw.github.com/mozilla/butter/master/LICENSE */
 
-define([ "util/xhr", "jsSHA/sha1", "util/shims" ],
```
jbuck, line 5: jsSHA/sha1 needs to be left there, PM crashes otherwise

src/first-run.js
```diff
@@ -7,7 +7,7 @@
  *
  * Determines whether or not a user should be shown a first-run dialog
  */
-define( [ "dialog/dialog", "ui/widget/tooltip", "util/shims" ], function( Dialog, ToolTip, Shims ) {
```
jbuck, line 10: Restore "util/shims"

templates/assets/editors/googlemap/googlemap-editor.js
```diff
@@ -1,4 +1,4 @@
-/*global google*/
+/*global */
```
jbuck: You can just delete this comment entirely if the "google" global isn't being used

src/cornfield/module.js
secretrobotron, line 5: Add a space after the first round brace here.

PR #1418, opened 2012-12-17
Initial work for our 0.1 of the webmaker dashboard

mjschranz:


cornfield/app.js
```diff
@@ -14,7 +14,7 @@ var express = require('express'),
     clientSessions = require('client-sessions'),
     lessMiddleware = require('less-middleware'),
     CONFIG = require('config'),
-    User = require( './lib/user' )( CONFIG.database ),
+    User = require( './lib/user' )( CONFIG.database, null, app.settings.env ),
```
jbuck, line 17: You can use process.env anywhere, so you don't need to pass this

Edit `process.env.NODE_ENV`

cornfield/lib/user.js
```diff
@@ -9,7 +9,7 @@ function defaultDBReadyFunction( err ) {
   }
 }
 
-module.exports = function( config, dbReadyFn ) {
+module.exports = function( config, dbReadyFn, environment ) {
```
jbuck, line 12: All of the stuff in this file could be redistributed to lib/webmaker_metrics or app.js. As humph pointed out, we're going to get bad data if the daily collection metrics were run on two webheads like our current setup.

cornfield/lib/webmaker_metrics.js
```diff
@@ -0,0 +1,38 @@
+"use strict";
+
+var metrics,
+    API_KEY = "ntQg43WhJFAAhxgm4dmZj1gFl56cLCg6SMmxNW5a9YUVW1Ca3j",
```
jbuck, line 4: this isn't a good place for a private api key

PR #1420, opened 2012-12-28
T2373b

mjschranz:
https://webmademovies.lighthouseapp.com/projects/65733-popcorn-maker/tickets/2373-land-plugin-updating-popcorn-maker-side


templates/assets/plugins/text/popcorn.text.js
```diff
@@ -13,13 +13,81 @@
 
   var DEFAULT_FONT_COLOR = "#000000",
       DEFAULT_SHADOW_COLOR = "#444444",
-      DEFAULT_BACKGROUND_COLOR = "#888888";
+      DEFAULT_BACKGROUND_COLOR = "#888888",
+      TEXT_SHADOW = "0 1px 5px {{shadowColor}}, 0 1px 10px {{shadowColor}}";
```
secretrobotron: Since you're using `TEXT_SHADOW.replace` in a few distinct places, perhaps it should be a function like `getTextShadowStyle(shadowColor)`.

secretrobotron:
1. Bug: Move the top trackevent to the 2nd track. The event's content drops to the bottom of the canvas. Selecting a different option for the "Text Position" property fixes the problem temporarily.
2. Similar bug: The same problem seems to occur with new trackevents as well. However, any "Text Position" property will have the same results -- content is stuck at the bottom.
3. Some in-line comments.

src/core/popcorn-wrapper.js
```diff
@@ -67,42 +67,37 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
...
-          // create the trackevent
-          _popcorn[ trackEvent.type ]( options );
-          // store a local reference to the newly created trackevent
-          _butterEventMap[ butterId ] = _popcorn.getLastTrackEventId();
+        if ( _popcorn[ trackEvent.type ] ) {
+
+          if ( popcornId ) {
+            _popcorn[ trackEvent.type ]( popcornId, newOptions );
+          } else {
+            _butterEventMap[ butterId ] = popcornId = butterId;
```
secretrobotron: If you're using the TrackEvent's `id` to dictate the id of the popcorn event (which is what's happening here, correct?), then `butterEventMap` is useless, since you can just look at the `id` of the connected TrackEvent.

src/core/trackevent.js
```diff
@@ -178,7 +178,7 @@ define( [ "./logger", "./eventmanager", "./observer",
       // if PopcornWrapper exists, it means we're connected properly to a Popcorn instance,
       // and can update the corresponding Popcorn trackevent for this object
       if ( _popcornWrapper ) {
-        _popcornWrapper.updateEvent( _this );
+        _popcornWrapper.synchronizeEvent( _this, updateOptions );
```
secretrobotron, line 181: Passing in `updateOptions` without updating it with some specific options might cause some trouble. For instance, z-index. If you place two trackevents, one on the first track and one on the second, and reorder the tracks, their zindex doesn't change properly. Curiously, their zindex does change when you move the trackevents themselves.

src/core/popcorn-wrapper.js
mjschranz: I was using it as a cleaner means of knowing if this was the track events first time through or if it was already "in our system".

I suppose I could change it, but I figure this is the least messy way.

secretrobotron: Does `_popcorn.getTrackEventById(_trackEvent.id)` not work?

templates/assets/editors/editorhelper.js
```diff
@@ -367,6 +369,31 @@
...
+
+      ln = container.childNodes.length;
+      for ( var k = 0; k < ln; ) {
+        element = container.childNodes[ k ];
+        if ( element.classList.contains( "ui-draggable-handle" ) ) {
+          container.removeChild( element );
+        } else {
+          k++;
+        }
+        ln = container.childNodes.length;
```
secretrobotron: You can probably make this loop simpler by iterating over the child nodes backwards. `for(var k = container.childNodes.length; k >= 0; --k)`. Then, `removeChild` won't affect the rest of the array.

src/core/popcorn-wrapper.js
mjschranz: Of course it would work, but wouldn't it be less efficient to loop through popcorn to find the correct track event vs checking a cached id?

secretrobotron: Marginally more speed-efficient, probably. To the same degree, more memory intensive, since you're storing twice the amount of mapping data trying to maintain `butterEventMap`. However, here I'm concerned with making the code less confusing. `_butterEventMap` is a pointless construct since we're letting Popcorn do the book-keeping for us. Our `Media->Track->TrackEvent` structure keeps track of our own `TrackEvents`, and since they're linked by name to  Popcorn's trackevent system, the only difference is one or two function calls (inside of Popcorn).

test/popcorn-wrapper/index.html
```diff
@@ -50,63 +50,49 @@
...
+              id: id,
+              type: "text",
+              popcornOptions: options
+            });
 
-          // Check new data vs stored values
-          equal( trackEvent.start, _trackEventData.start, "Successfully updated start time to " + _trackEventData.start );
-          equal( trackEvent.end, _trackEventData.end, "Successfully updated end time to " + _trackEventData.end );
-          equal( trackEvent.text, _trackEventData.text, "Successfully updated text to " + _trackEventData.text );
+            _pUpdate.synchronizeEvent( { id: id, type: "text" }, updateOptions );
```
secretrobotron: This doesn't look quite right. Did you mean to use the object above (https://github.com/mozilla/butter/pull/1420/files#L6R84)?

mjschranz: Nope. The first case is when we would basically be passing just the track event (although yes, in reality we always pass along updating options now). The difference is because it's the first time it's being adding to the popcorn instance they will be ignored.

mjschranz: Rather, I think I see what you mean. I'll give them better names for less confusion.

src/core/trackevent.js
ScottDowne, line 181: I'm not hitting this.

Either it was fixed, or I need STRs.

Pretty sure line 169 fixed it though.

PR #1426, opened 2013-01-08
[#2992] Prevent editor from being closed by dragging/resizing/droppable/...

mjschranz:
...contenteditable if already opened for current track event

### DIFF
--- a/src/editor/module.js
+++ b/src/editor/module.js
@@ -160,13 +160,19 @@ define( [ "core/eventmanager", "core/trackevent", "./editor",
      * @param {TrackEvent} trackEvent: TrackEvent to edit
      */
     _this.editTrackEvent = function( trackEvent ) {
+      var editorType = Editor.isRegistered( trackEvent.type ) ? trackEvent.type : "default";
+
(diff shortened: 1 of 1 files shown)


src/editor/module.js
```diff
@@ -160,13 +160,19 @@ define( [ "core/eventmanager", "core/trackevent", "./editor",
...
       if ( !trackEvent || !( trackEvent instanceof TrackEvent ) ) {
         throw new Error( "trackEvent must be valid to start an editor." );
       }
-      var editorType = Editor.isRegistered( trackEvent.type ) ? trackEvent.type : "default";
-      return _this.openEditor( editorType, {
-        openData: trackEvent
-      });
+
+      if ( _currentEditor && _currentEditor.getTrackEvent ) {
+        if ( trackEvent.id !== _currentEditor.getTrackEvent().id ) {
```
ScottDowne: I am still pretty sure this needs to be a ===.

We want to return the existing open editor if the editing track event is the same as the current open editor's track event, otherwise open a new one.

PR #1429, opened 2013-01-10
updated instructions for running butter on windows

Pomax:
This gives the specific visual studio and SDK things that need to be installed on windows before "npm install" will succeed. The process is slightly more elaborate than initially documented =)


jbuck:
Cherry-picked by f44656f0ce89904669b67db1ba1ac3876927dab5

PR #1430, opened 2013-01-15
[#2988] Add preview buttons for each project on the dashboard

mjschranz:


jbuck:
Travis build failed due to `==` instead of `===`, which is weird because your patch doesn't do either...

PR #1433, opened 2013-01-16
[#3003] Fallback to null wrapper with same duration on loading fails.

ScottDowne:
This patch only handles it on embed.

### DIFF
--- a/cornfield/app.js
+++ b/cornfield/app.js
@@ -170,17 +170,11 @@ app.post( '/api/publish/:id',
       var headEndTagIndex,
           bodyEndTagIndex,
           externalAssetsString = '',
-          popcornString = '',
-          currentMedia,
-          currentTrack,
-          currentTrackEvent,
-          mediaPopcornOptions,
           templateURL,
           baseString,
           headStartTagIndex,
           templateScripts,
           startString,
-          numSources,
           j, k, len;
 
       templateURL = templateFile.substring( templateFile.indexOf( '/templates' ), templateFile.lastIndexOf( '/' ) );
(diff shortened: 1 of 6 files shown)


src/core/media.js
```diff
@@ -281,6 +282,15 @@
         }
       }
 
+      this.listen( "mediatimeout", function() {
```
secretrobotron: You don't really need to listen for this event to execute this code, since the event is being dispatched from this object anyway. Try placing it https://github.com/ScottDowne/butter/blob/98cde0dec37946d656418d39114fa229885e54c7/src/core/media.js#L89 instead.

src/core/media.js
```diff
@@ -281,6 +282,15 @@
...
+        // If we have a duration from saved data, we can load up the media with that instead of spinning forever.
+        if ( _fallback && _target ) {
```
secretrobotron: It's not clear why these are the preconditions for interrupting a load and using a `#t=` as a fallback. Can you flesh out that comment a bit?

Also, what's the point of `_fallback`? Why not just always use `_duration`? Maybe you can always _try_ to store a good duration by recording what used to be there as `_oldDuration` or something. What happens if a proper duration isn't stored and the media fails to load? Some explanation here in general would be good for sanity.

ScottDowne: Regarding using duration. If a media existed, and was then updated inside the editor, with an invalid url, it'll detect an old duration. I could clear the duration, but that worried me because of potential unexpected results. This seemed much more explicit, and would only ever happen via saved data, and could not hurt anything else.

Basically, duration can be set from other areas.

I would be willing to sort out duration to make it work though, if you think it is worth it.

cornfield/app.js
```diff
@@ -212,52 +206,17 @@ app.post( '/api/publish/:id',
...
              data.substring( headEndTagIndex, bodyEndTagIndex ) +
-             popcornString + data.substring( bodyEndTagIndex );
+             data.substring( bodyEndTagIndex );
 
       // Convert 1234567890 => "kf12oi"
       var idBase36 = utils.generateIdString( id ),
           publishUrl = utils.generatePublishUrl( id ),
           iframeUrl = utils.generateIframeUrl( id );
 
+      fs.writeFile( "view/v/" + idBase36 + ".json", JSON.stringify( projectData, null, 2 ) );
```
jbuck, line 218: This will work locally, but not on prod where we use S3 to store published files. This line should probably go above in the publishEmbed function, or the like

css/embed.less
```diff
@@ -7,6 +7,14 @@
   transition: @transition1;
 }
 
+.transition(@transition1, @transition2) {
```
jbuck, line 10: This function already exists in `css/globals.less`. If you need to use it, just import it

css/embed.less
```diff
@@ -666,6 +674,27 @@ body {
...
+
+  height: 3em;
+  background: url( "../resources/icons/big-loader.gif" ) center no-repeat;
+
+  width: 100%;
+  line-height: 1;
+  top: 50%;
+
+  margin-top: -1.5em;
+  z-index: 9999;
```
jbuck, line 695: I wonder if there's a z-index variable you can ride on, rather than hardcoding one?

src/embed.js
```diff
@@ -265,9 +267,11 @@ function init( window, document ) {
       var Butter = {
             version: "Butter-Embed-@VERSION@"
           },
-          popcorn = Popcorn.byId( "Butter-Generated" ),
+          popcorn,
           config,
-          qs = URI.parse( window.location.href ).queryKey,
+          uri = URI.parse( window.location.href ),
+          qs = uri.queryKey,
+          id = uri.file.replace( "_.html", "" ),
```
jbuck, line 274: This is problematic. Hmmmmmm, I wonder how we're going to resolve the filename of the JSON file based on whether it's a local file, or a file hosted on S3

PR #1436, opened 2013-01-21
[t3019] better args for url.replace in BaseLoader

secretrobotron:


src/loaders/base-loader.js
```diff
@@ -13,12 +13,17 @@ define( [], function() {
     var match,
         replacement;
 
+    // Replace {variables}
     while ( VAR_REGEX.test( url ) ) {
       match = VAR_REGEX.exec( url );
       replacement = configDirs[ match[ 1 ] ] || "";
       url = url.replace( match[0], replacement );
     }
-    return url.replace( "//", "/" );
```
cadecairos, line 21: Why did we even need this in the first place? I can't think of a use case where this replace would be useful

PR #1442, opened 2013-01-24
[t3026] Reduce min sizes for resizable images. Fix handle zindexing

mjschranz:


templates/assets/css/jquery-ui/jquery.ui.butter.less
```diff
@@ -49,7 +49,7 @@
 .ui-resizable-handle {
   position: absolute;
   display: block;
-  z-index: @handleZIndex + 1;
+  z-index: @handleZIndex + 1 !important;
```
cadecairos: Why did you add `!important`?

PR #1443, opened 2013-01-24
[#3028] Sequencer living as spawner for now. Need UI to polish it off.

ScottDowne:


templates/assets/editors/sequencer/sequencer.js
```diff
@@ -0,0 +1,20 @@
...
+  _popcornOptions = trackEvent.popcornTrackEvent;
+  _container = _popcornOptions._container;
+  _target = _popcornOptions._target;
+
+  if ( window.jQuery ) {
+
+    window.EditorHelper.resizable( trackEvent, _container, _target, {
+      minWidth: 40,
+      minHeight: 40
+    });
```
mjschranz: Read the following https://github.com/mozilla/butter/blob/master/templates/assets/editors/editorhelper.js#L128-L130

Without specifying `handlePositions` it's exposing a bug with jQueryUI Resizable and your resize handles are appearing below the drag handles.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,465 @@
...
+      // Higher this is, the longer out of sync we'll accept a clip.
+      // 0 forces seeks to wait.
+      // When one clip changes over to another, there is a moment where the video needs to play.
+      // We can assume this is going to be fast enough in most cases.
+      // If the seek doesn't finish in this time, we pause the video to continue to wait.
+      // This means in best cases our sequence is smooth, but slightly out of sync,
+      // In worst cases it pauses.
+      WAIT_FOR_SEEK_BEFORE_PAUSE = 350;
+ 
+  Popcorn.plugin( 'sequencer', {
```
mjschranz: Quotes.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,465 @@
...
+          options.p.off( "seeked", seekedEvent );
+          options.p.on( "play", playedEvent );
+          options.p.play();
+        };
+        options.p.on( "seeked", seekedEvent);
+        // assume the is fast, but if not, puse the main video and wait.
+        seekTimeout = setTimeout( function() {
+          _this.pause();
+        }, WAIT_FOR_SEEK_BEFORE_PAUSE );
+        options.p.currentTime( _this.currentTime() - options.start + (+options.from) );
```
mjschranz: space.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,465 @@
...
+          }
+        }
+      };
+
+      options._pauseEvent = function() {
+        options.p.pause();
+      };
+
+      options._seekedEvent = function() {
+        options.p.currentTime( _this.currentTime() - options.start + (+options.from) );
```
mjschranz: Space.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,465 @@
...
+          this.on( "play", options._surpressPlayEvent );
+          if ( !this.paused() ) {
+            options.playWhenReady = true;
+            this.pause();
+            options.p.pause();
+          }
+          options.addSource();
+        }
+      }
+      if ( updates.mute != null ) {
```
mjschranz: Why the `!=` for all of these? Why not just do a check for `hasOwnProperty`? That way you know they provided the options explicitly, especially in the case of options that are strings and you might consider and empty string valid.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,465 @@
...
+          hidden: true
+        },
+        denied: {
+          hidden: true
+        }
+      }
+    }
+  });
+
+}(Popcorn));
```
mjschranz: Space.

src/main.js
```diff
@@ -170,7 +170,9 @@
           start = 0;
         }
 
-        end = start + _defaultTrackeventDuration;
+        if ( end == null ) {
+          end = start + _defaultTrackeventDuration;
+        }
```
mjschranz: Rather than tacking on more to the signature definition we could change this to accept a single popcornOptions object. That way we can defined start/end easily on their own and anything else we might want to in the future.

src/core/popcorn-wrapper.js
```diff
@@ -71,26 +71,45 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
...
-            }
+        // make sure we have a reference to the trackevent before calling toString
+        if ( trackEvent.popcornTrackEvent ) {
+          trackEvent.view.elementText = trackEvent.popcornTrackEvent.toString();
+          // we should only get here if no exceptions happened
+          trackEvent.dispatch( "trackeventupdated", trackEvent );
+        }
+      }
+
+      if( _popcorn ){
```
mjschranz: Add back the spacing.

src/core/popcorn-wrapper.js
```diff
@@ -71,26 +71,45 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
...
+        // make sure the plugin is still included
+        if( _popcorn[ trackEvent.type ] ){
```
mjschranz: Spacing.

src/core/popcorn-wrapper.js
```diff
@@ -71,26 +71,45 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
...
+        if ( _popcorn.getTrackEvent( popcornId ) ) {
+          _popcorn[ trackEvent.type ]( popcornId, newOptions );
+        } else {
+          _popcorn[ trackEvent.type ]( popcornId, options );
+        }
+
+        popcornEvent = _popcorn.getTrackEvent( popcornId );
+        trackEvent.popcornTrackEvent = popcornEvent;
+
+        if( trackEvent.view ){
```
mjschranz: spacing.

src/core/popcorn-wrapper.js
```diff
@@ -71,26 +71,45 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
...
+          if( popcornEvent.toString ){
```
mjschranz: Spacing.

src/core/popcorn-wrapper.js
```diff
@@ -71,26 +71,45 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
...
+            trackEvent.view.setToolTip( popcornEvent.toString() );
           }
+          else{
```
mjschranz: up to the line above.

css/media-editor.less
```diff
@@ -22,77 +49,198 @@
...
-      float: left;
+  .media-gallery-scroll-container {
+    position: absolute;
+    top: 45px;
+    left: 0;
+    right: 0;
+    bottom: 0;
+    overflow-y: scroll;
+  }
+  .media-gallery-list{
```
mjschranz: space bracket

src/main.js
ScottDowne: I'm torn on having an options object here.

If we have one, we open the flood gates to anything and everything. It makes it easy to pass in and check against props that have nothing to do with generating a safe track event. End actually makes a lot of sense for generating a safe track event. We could probably pull this change out an do it in another ticket seperate from sequencer.

mjschranz: I was mainly thinking of an easy way to deal with your double equals here. I can see why you wouldn't want a generic options object but you will have to put checks in here then to ensure end is what you think it is.

src/core/popcorn-wrapper.js
```diff
@@ -71,26 +71,57 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
...
+              } else {
+                trackEvent.view.element.classList.add( "sequencer-audio" );
+                trackEvent.view.element.classList.remove( "sequencer-video" );
+              }
+            }
+          }
+          else{
+            trackEvent.view.setToolTip( JSON.stringify( options ) );
           }
+        }
```
mjschranz, line 103: Fix the spacing of this if structure.

src/core/popcorn-wrapper.js
```diff
@@ -440,11 +477,11 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
     /* Wait for Popcorn to be set up and to have the required players load (uses
      * checkTimeoutLoop).
      */
-    function waitForPopcorn( readyCallback, timeoutCallback ){
-      if( _mediaType !== "object" ){
-        _onPlayerTypeRequired( _mediaType );
+    function waitForPopcorn( readyCallback, timeoutCallback, mediaType ){
+      if( mediaType !== "object" ){
```
mjschranz: spacing.

src/core/popcorn-wrapper.js
```diff
@@ -440,11 +477,11 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
     /* Wait for Popcorn to be set up and to have the required players load (uses
      * checkTimeoutLoop).
      */
-    function waitForPopcorn( readyCallback, timeoutCallback ){
-      if( _mediaType !== "object" ){
-        _onPlayerTypeRequired( _mediaType );
+    function waitForPopcorn( readyCallback, timeoutCallback, mediaType ){
```
mjschranz: Spacing.

src/core/trackevent.js
```diff
@@ -174,23 +174,13 @@ define( [ "./logger", "./eventmanager", "./observer",
 
       _popcornOptions.start = newStart;
       _popcornOptions.end = newEnd;
+      _this.popcornOptions = _popcornOptions;
```
mjschranz, line 177: Why does this need to happen here? The popcorn options attribute of the TrackEvent should always be pointing to `_popcornOptions`

src/editor/media-gallery-editor.js
```diff
@@ -0,0 +1,216 @@
+/* This Source Code Form is subject to the terms of the MIT license
+ * If a copy of the MIT license was not distributed with this file, you can
+ * obtain one at https://raw.github.com/mozilla/butter/master/LICENSE */
+
+define( [ "util/lang", "util/xhr", "util/keys", "util/mediatypes", "editor/editor", "util/time", "util/dragndrop", "text!layouts/media-editor.html" ],
```
mjschranz: Maybe consider adding this onto more than one line so it doesn't stretch so far to the right?

src/editor/media-gallery-editor.js
```diff
@@ -0,0 +1,216 @@
...
+      return;
+    }
+    if ( /[0-9]{1,9}/.test( duration ) ) {
+      _media.url = "#t=," + duration;
+    } else {
+      _media.url = duration;
+    }
+  }
+
+
```
mjschranz: Extra new line.

src/editor/media-gallery-editor.js
```diff
@@ -0,0 +1,216 @@
...
+      thumbnailImg.src = data.thumbnail;
+    }
+    thumbnailBtn.appendChild( thumbnailImg );
+    thumbnailBtn.src = data.thumbnail;
+
+    el.classList.add( "mg-" + data.type );
+    el.classList.add( "new" );
+
+    if ( data.denied ) {
+     el.querySelector( ".mg-error" ).innerHTML = "Embedding disabled by request";
```
mjschranz: indent 1 space.

src/editor/media-gallery-editor.js
```diff
@@ -0,0 +1,216 @@
...
+      _addBtn.classList.remove( "hidden" );
+      _cancelBtn.classList.remove( "hidden" );
+    } else {
+      _addBtn.classList.add( "hidden" );
+      _cancelBtn.classList.add( "hidden" );
+    }
+  }
+
+  function onEnter( e ) {
+    if (  e.keyCode === KeysUtils.ENTER ) {
```
mjschranz: extra space

src/editor/media-gallery-editor.js
```diff
@@ -0,0 +1,216 @@
...
+    _butter = butter;
+
+    setup();
+
+    Editor.BaseEditor.extend( _this, butter, rootElement, {
+      open: function() {
+        _media = butter.currentMedia;
+
+        setBaseDuration( _media.duration );
+
```
mjschranz: Might as well remove the extra new line.

src/editor/sequencer-editor.js
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        for ( key in _pluginOptions ) {
+          if ( _pluginOptions.hasOwnProperty( key ) ) {
+            option = _pluginOptions[ key ];
+
+            if ( option.elementType === "select" ) {
+              _this.attachSelectChangeHandler( option.element, option.trackEvent, key, _this.updateTrackEventSafe );
+            }
+            else if ( option.elementType === "input" ) {
```
mjschranz: bump up a line.

src/editor/sequencer-editor.js
```diff
@@ -0,0 +1,152 @@
...
+                _this.attachInputChangeHandler( option.element, option.trackEvent, key, sourceCallback );
+              } else if ( key === "from" ) {
+                _this.attachInputChangeHandler( option.element, option.trackEvent, key, fromCallback );
+              } else if ( option.element.type === "checkbox" ) {
+                _this.attachCheckboxChangeHandler( option.element, option.trackEvent, key, checkboxCallback );
+              } else {
+                _this.attachInputChangeHandler( option.element, option.trackEvent, key, _this.updateTrackEventSafe );
+              }
+            }
+            else if ( option.elementType === "textarea" ) {
```
mjschranz: bump up a line

src/editor/sequencer-editor.js
```diff
@@ -0,0 +1,152 @@
...
+        trackEvent: trackEvent,
+        callback: callback,
+        basicContainer: basicContainer,
+        ignoreManifestKeys: [ "start", "end" ]
+      });
+
+      attachHandlers();
+      _this.updatePropertiesFromManifest( trackEvent );
+      if ( trackEvent.popcornOptions.source ) {
+        if( !Array.isArray( trackEvent.popcornOptions.source ) ) {
```
mjschranz: space.

src/editor/sequencer-editor.js
```diff
@@ -0,0 +1,152 @@
...
+      }
+      _this.setTrackEventUpdateErrorCallback( _this.setErrorState );
+    }
+
+    function anchorClickPrevention( anchorContainer ) {
+      if ( anchorContainer ) {
+        
+        anchorContainer.onclick = _falseClick;
+      }
+    }
```
mjschranz: You don't have any anchors involved. Remove this copy pasta code.

src/editor/sequencer-editor.js
```diff
@@ -0,0 +1,152 @@
...
+
+    function onTrackEventUpdated( e ) {
+      _trackEvent = e.target;
+
+      var anchorContainer = _trackEvent.popcornTrackEvent._container.querySelector( "a" );
+      anchorClickPrevention( anchorContainer );
```
mjschranz: Copy pasta. Remove.

src/editor/sequencer-editor.js
```diff
@@ -0,0 +1,152 @@
...
+
+      _this.updatePropertiesFromManifest( _trackEvent );
+      if ( _trackEvent.popcornOptions.source ) {
+        if( !Array.isArray( _trackEvent.popcornOptions.source ) ) {
```
mjschranz: Space.

src/editor/sequencer-editor.js
```diff
@@ -0,0 +1,152 @@
...
+      }
+      _this.setErrorState( false );
+    }
+
+    // Extend this object to become a TrackEventEditor
+    Butter.Editor.TrackEventEditor.extend( _this, butter, rootElement, {
+      open: function( parentElement, trackEvent ) {
+        var anchorContainer = trackEvent.popcornTrackEvent._container.querySelector( "a" );
+
+        anchorClickPrevention( anchorContainer );
```
mjschranz: Copy pasta. Remove.

src/main.js
mjschranz: This still applies.

src/timeline/status.js
```diff
@@ -44,7 +44,8 @@ define( [ "util/lang" ], function( util ){
       if( typeof( time ) === "string" || !isNaN( time ) ){
         if( setCurrentTime ){
           try {
-            _media.currentTime = Popcorn.util.toSeconds( time );
+            time = Popcorn.util.toSeconds( time );
+            _media.currentTime = time;
```
mjschranz, line 48: Not really sure why this changes anything.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,517 @@
...
+          _this.play();
+        }
+      };
+
+      options.tearDown = function() {
+        _this.off( "volumechange", options._volumeEvent );
+        if ( options.p ) {
+          // XXX: pull the SoundCloud iframe element out of our video div, and quarantine
+          // so we don't delete it, and block loading future SoundCloud instances. See above.
+          // this is also fixing an issue in youtube, so we do it for all medias with iframes now.
```
mjschranz: If all of this is fixing issues in general for iframes, why not name it something more appropriate?

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,517 @@
...
+          clearTimeout( options.loadTimeout );
+        }
+        // if the video is denied for any reason, most cases youtube embedding disabled,
+        // don't bother waiting and display fail case.
+        if ( options.denied ) {
+          options.fail();
+        } else {
+          options.loadTimeout = setTimeout( options.fail, MEDIA_LOAD_TIMEOUT );
+        }
+        options.p = Popcorn.smart( options._container, options.source, {frameAnimation: true} );
```
mjschranz: Spacing on teh object.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,517 @@
...
+          options.readyEvent();
+        } else {
+          options.p.on( "loadedmetadata", options.readyEvent );
+        }
+      };
+      options.setupContainer();
+      if ( options.source ) {
+        options.sourceToArray();
+        if ( options.fallback ) {
+          if( !Array.isArray( options.fallback ) ) {
```
mjschranz: spacing.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,517 @@
...
+          options._container.style.zIndex = 0;
+        }
+      }
+      if ( updates.fallback ) {
+        options.fallback = updates.fallback;
+      }
+      if ( updates.source ) {
+        options.sourceToArray();
+        if ( options.fallback ) {
+          if( !Array.isArray( options.fallback ) ) {
```
mjschranz: Space.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,517 @@
...
+            options.fallback = [ options.fallback ];
+          }
+          if( !Array.isArray( updates.source ) ) {
```
mjschranz: Space.

src/core/popcorn-wrapper.js
```diff
@@ -71,26 +71,57 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
...
+            if ( trackEvent.type === "sequencer" ) {
+              if ( !trackEvent.popcornOptions.hidden ) {
+                trackEvent.view.element.classList.add( "sequencer-video" );
+                trackEvent.view.element.classList.remove( "sequencer-audio" );
+              } else {
+                trackEvent.view.element.classList.add( "sequencer-audio" );
+                trackEvent.view.element.classList.remove( "sequencer-video" );
+              }
+            }
+          }
```
mjschranz: I forgot what all this was doing. You explained it to me before but please do so again.

src/core/popcorn-wrapper.js
```diff
@@ -71,26 +71,57 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
...
+          _popcorn[ trackEvent.type ]( popcornId, newOptions );
+        } else {
+          _popcorn[ trackEvent.type ]( popcornId, options );
+        }
+
+        popcornEvent = _popcorn.getTrackEvent( popcornId );
+        trackEvent.popcornTrackEvent = popcornEvent;
+
+        trackEvent.popcornOptions.start = +popcornEvent.start;
+        trackEvent.popcornOptions.end = +popcornEvent.end;
```
mjschranz, line 86: Wait, why? Updating adds this. I don't understand the need behind this.

src/layouts/media-editor.html
```diff
@@ -1,52 +1,53 @@
...
+    <fieldset class="media-gallery-base-media butter-form-inline form-single top">
+      <label>Duration</label>
+      <input type="text" class="media-base-duration" value="">
+    </fieldset>
+    <fieldset class="add-media-panel">
+      <div class="media-editor-inner-wrapper primary-media-wrapper">
+        <p>
+          <span class="fl-right">
+            <span class="media-icon youtube-icon"></span>
+            <!--vimeo is temporarily not supported<span class="media-icon vimeo-icon"></span>-->
```
humphd: Why are we not supporting Vimeo?  If this is staying, we need a bug filed to re-enable it, and the bug # here.

src/util/mediatypes.js
```diff
@@ -0,0 +1,117 @@
...
+  return {
+    checkUrl: function( url ) {
+      for ( var type in REGEX_MAP ) {
+        if ( REGEX_MAP.hasOwnProperty( type ) ) {
+          if ( REGEX_MAP[ type ].test( url ) ) {
+            return type;
+          }
+        }
+      }
+      return "html5";
```
humphd, line 25: This is sane, but I wonder if we should follow-up on this later so that we give some indication to the user that we're making a guess, and trying HTML5 media.  If they give us a URL run through a shortener, or we otherwise fail to catch a valid URL of another type in the regexes above, we'll fail, and fail in ways that are hard to debug.

src/util/mediatypes.js
```diff
@@ -0,0 +1,117 @@
...
+    },
+    getMetaData: function( baseUrl, callback, type ) {
+      var data = {},
+          id,
+          parsedUri,
+          splitUriDirectory,
+          xhrURL,
+          testEl;
+
+      if ( baseUrl === "test" ) {
```
humphd: You're going to pull this out, right?  Let's not do this in production.

src/util/mediatypes.js
```diff
@@ -0,0 +1,117 @@
...
+        baseUrl = EXAMPLES.youtube;
+      }
+
+      type = data.type = type || this.checkUrl( baseUrl );
```
humphd: Simplify this line, split into 2:

```
type = type || this.checkUrl( baseUrl );
data.type = type;
```

src/util/mediatypes.js
```diff
@@ -0,0 +1,117 @@
...
+      callback = callback || function(){};
+
+      if ( type === "youtube" ) {
+        parsedUri = URI.parse( baseUrl );
+        id = parsedUri.queryKey.v || parsedUri.directory.replace( "/", "" );
```
humphd: Add a note here on the two formats you're expecting (i.e., give examples of each, so debugging is easier later).

src/util/mediatypes.js
```diff
@@ -0,0 +1,117 @@
...
+        if ( !id ) {
+          return;
```
humphd: So the type was "youtube" but you failed to get an id and you're just going to silently return?  This is going to hide a bug if we get the parsing of their URLs wrong, no?  Should we not do something in the editor that says, "This YouTube URL doesn't seem to be valid, please check..."

src/util/mediatypes.js
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        xhrURL = "https://gdata.youtube.com/feeds/api/videos/" + id + "?v=2&alt=jsonc&callback=?";
+        Popcorn.getJSONP( xhrURL, function( resp ) {
+          var raw = resp.data;
+          if ( !raw ) {
+            return;
```
humphd: Under which conditions will we hit this?  Can we do better than hide the failure?  "Unable to retrieve data from YouTube for this clip..."

src/util/mediatypes.js
```diff
@@ -0,0 +1,117 @@
...
+          callback( data );
+        });
+      } else if ( type === "soundcloud" ) {
+        parsedUri = URI.parse( baseUrl );
+        splitUriDirectory = parsedUri.directory.split( "/" );
+        id = splitUriDirectory[ splitUriDirectory.length - 1 ];
+        xhrURL = "http://api.soundcloud.com/tracks/" + id + ".json?callback=?&client_id=PRaNFlda6Bhf5utPjUsptg";
+        Popcorn.getJSONP( xhrURL, function( raw ) {
+          if ( !raw ) {
+            return;
```
humphd: Same deal here.

src/util/mediatypes.js
```diff
@@ -0,0 +1,117 @@
...
+          }
+
+          if ( raw.sharing === "private" || raw.embeddable_by === "none" ) {
+            data.denied = true;
```
humphd: Good to drop a note in here about why you are doing this; helps later with debugging.

src/util/mediatypes.js
```diff
@@ -0,0 +1,117 @@
...
+        if ( REGEX_MAP.hasOwnProperty( type ) ) {
+          if ( REGEX_MAP[ type ].test( url ) ) {
+            return type;
+          }
+        }
+      }
+      return "html5";
+    },
+    getMetaData: function( baseUrl, callback, type ) {
+      var data = {},
```
humphd: Instead of create a variable for this, just create an object literal for each callback below:

```
callback({
  source: "http://www.youtube.com/watch?v=" + id,
  title: raw.title,
   ...
});
```

The fewer variables you have, the fewer state bugs you'll have.

src/util/mediatypes.js
```diff
@@ -0,0 +1,117 @@
...
+          data.duration = raw.duration / 1000;
+          data.title = raw.title;
+          data.hidden = true;
+          callback( data );
+        });
+      } else if ( type === "vimeo" ) {
+        parsedUri = URI.parse( baseUrl );
+        splitUriDirectory = parsedUri.directory.split( "/" );
+        id = splitUriDirectory[ splitUriDirectory.length - 1 ];
+        xhrURL = "http://vimeo.com/api/v2/video/" + id + ".json?callback=?";
```
humphd, line 96: Same deal, don't hide error cases.

src/util/mediatypes.js
```diff
@@ -0,0 +1,117 @@
...
+        Popcorn.getJSONP( xhrURL, function( raw ) {
+          raw = raw && raw[ 0 ];
+          if ( !raw ) {
+            return;
+          }
+          data.source = baseUrl;
+          data.thumbnail = raw.thumbnail_small;
+          data.duration = raw.duration;
+          data.title = raw.title;
```
humphd: This needs a better name, `mediaElem` maybe?

templates/assets/editors/sequencer/sequencer.js
```diff
@@ -0,0 +1,21 @@
...
+  var _container,
+      _popcornOptions,
+      _target;
+
+  _popcornOptions = trackEvent.popcornTrackEvent;
+  _container = _popcornOptions._container;
+  _target = _popcornOptions._target;
+
+  if ( window.jQuery ) {
+
```
humphd: Remove blank line.

css/editors.less
```diff
@@ -269,6 +269,11 @@
     .clearfix();
   }
 
+  .editor-inner {
+    padding-left: 20px;
+    padding-right: 20px;
+  }
+
```
mjschranz: This class doesn't appear to do anything and can be removed, along with wherever it's used.

css/media-editor.less
```diff
@@ -4,11 +4,38 @@
 
 .media-editor {
 
+  @titleBarHeight: 60px;
+  @addClipHeight: 190px;
+
+  .butter-editor-body {
+    top: @titleBarHeight;
+  }
+
```
mjschranz, line 14: This class also doesn't appear to add much. Removing it changed nothing.

css/media-editor.less
```diff
@@ -22,77 +49,198 @@
...
-      color: #000;
+  .media-gallery {
+    position: absolute;
+    bottom: 0;
+    top: @addClipHeight + 20px;
+    width: 100%;
+    margin-top: -1px;
+  }
+
+  .media-gallery-heading {
```
mjschranz: You could change this so the class is inside `media-gallery`. It's only used in the scenario where `media-gallery` is a parent of it anyway.

``` less
.media-gallery {
  position: absolute;
  bottom: 0;
  top: @addClipHeight + 20px;
  width: 100%;
  margin-top: -1px;

 .media-gallery-heading {
    background: #FFF;
    padding: 15px 20px;
    border: 0 solid @baseLightOutline;
    border-bottom-width: 1px;
    border-top-width: 1px;
    font-size: 11px;
    font-weight: 700;
    text-transform: uppercase;
    position: absolute;
    top: 0;
    left: 0;
    right: 0;
  }
}
```

css/media-editor.less
```diff
@@ -22,77 +49,198 @@
...
+    background: #FFF;
+    padding: 15px 20px;
+    border: 0 solid @baseLightOutline;
+    border-bottom-width: 1px;
+    border-top-width: 1px;
+    font-size: 11px;
+    font-weight: 700;
+    text-transform: uppercase;
+
```
mjschranz, line 78: Lose the new line.

src/core/media.js
```diff
@@ -548,7 +548,7 @@
           },
           set: function( time ){
             if( time ){
-              _duration = time;
+              _duration = +time;
               _logger.log( "duration changed to " + _duration );
```
mjschranz, line 552: Was there somehow a chance we were passing in strings here?

src/timeline/media.js
```diff
@@ -199,10 +199,30 @@ define( [ "core/trackevent", "core/track", "core/eventmanager",
...
+        if ( popcornOptions && popcornOptions.end ) {
+          end = popcornOptions.end + start;
+        }
+        trackEvent = butter.generateSafeTrackEvent( type, start, end, track );
+        if ( popcornOptions ) {
+          if ( popcornOptions.end ) {
+            popcornOptions.end = trackEvent.popcornOptions.end;
+          }
+          trackEvent.update( popcornOptions );
+        }
```
mjschranz, line 225: This is sort of why I was thinking earlier you could simply have generateSafeTrackEvent take a Popcorn Options Object. Why not let the creation of a TrackEvent object handle this for you?

I don't disagree that options other than start and end aren't important for the idea of figuring out where a trackevent can go but at the same time this function is used as a utility for adding new trackevents at run time in general. I really don't see the downside of doing this, nor why it's considered opening a flood gate.

On top of that you are forcing an extra update call to be thrown around in the event system.

src/util/mediatypes.js
mjschranz: I think data is appropriate. This isn't the element itself but just options/data being used to build the element itself in the new media editor.

src/util/mediatypes.js
```diff
@@ -0,0 +1,117 @@
...
+            data.denied = true;
+          }
+          data.source = baseUrl;
+          data.thumbnail = raw.artwork_url;
+          data.duration = raw.duration / 1000;
+          data.title = raw.title;
+          data.hidden = true;
+          callback( data );
+        });
+      } else if ( type === "vimeo" ) {
```
mjschranz, line 92: Sort of along the line of Humph's comment about commenting the Vimeo line out in the editor HTML, should we not consider bailing early here if we aren't supporting Vimeo right now, along with some sort of info to the user?

templates/basic/style.less
```diff
@@ -55,43 +55,51 @@ body.editor-minimized .body-wrapper {
...
+  &.show-media {
+    opacity: 1;
+    // this zindex puts it above all track events
+    z-index: 1000;
+    visibility: visible;
+  }
+  &.show-waiting {
+    opacity: 1;
+    // this zindex puts it above all track events
+    background: rgba(0,0,0,1);
```
mjschranz: spacing

templates/basic/style.less
```diff
@@ -55,43 +55,51 @@ body.editor-minimized .body-wrapper {
...
+  &.show-global,
+  &.show-media {
+    opacity: 1;
+    // this zindex puts it above all track events
+    z-index: 1000;
+    visibility: visible;
+  }
+  &.show-waiting {
+    opacity: 1;
+    // this zindex puts it above all track events
```
mjschranz, line 83: This comment should be nugged down to line 85 if you are keeping it.

src/editor/media-gallery-editor.js
```diff
@@ -0,0 +1,216 @@
...
+      duration = Time.toSeconds( duration );
+    }
+    if ( duration === _media.duration ) {
+      return;
+    }
+    if ( /[0-9]{1,9}/.test( duration ) ) {
+      _media.url = "#t=," + duration;
+    } else {
+      _media.url = duration;
+    }
```
mjschranz, line 64: As mentioned in ticket, you don't deal with duration being an empty string here.

src/editor/sequencer-editor.js
```diff
@@ -0,0 +1,152 @@
...
+
+      function attachHandlers() {
+        var key,
+            option;
+
+        function sourceCallback( trackEvent, updateOptions, prop ) {
+          MediaUtils.getMetaData( updateOptions.source, function( data ) {
+            updateOptions.duration = data.duration;
+            updateOptions.denied = data.denied;
+            trackEvent.update( updateOptions );
```
mjschranz, line 45: Could a new video not also perhaps now be considered hidden? I'm wondering if we care about that data returned from the MediaUtils and perhaps anything else as well.

src/util/mediatypes.js
mjschranz, line 96: This comments looks like it was for line 96.

src/editor/sequencer-editor.js
ScottDowne, line 45: We only load in what cannot be set by the user.

If, for example, a title was set by the user, and then the source is updated. Should we blow away the title? Or leave it? Right now we only do changes like that if it is a new, empty media. Once it exists, we change only what the user changes.

src/timeline/media.js
ScottDowne, line 225: I agree there is work we can do here to make this better.

I don't think we need to worry about it now.

Can we file this, a ticket to make generateSafeTrackEvent's signature better?

src/core/media.js
ScottDowne, line 552: Yes, actually. From my testing it was always coming in as a string.

This at the very worst is just defensive.

css/media-editor.less
ScottDowne, line 14: It adds a few pixels. Hard to notice, but they are there.

Trying changing it inside a debugger, with the editor open. You'll see it change.

mjschranz, line 14: I did this and noticed nothing happening.

src/util/mediatypes.js
ScottDowne: I actually forget why this happened. I'll get back to you on this...

src/util/mediatypes.js
ScottDowne: Going to add an error dialog we can use.

src/util/mediatypes.js
ScottDowne, line 25: Looks like more error dialog, maybe also a warning dialog?

src/layouts/media-editor.html
ScottDowne: We're going to vimeo after all. I'll add this back in.

src/core/popcorn-wrapper.js
ScottDowne, line 86: This gives track event's the ability to override defaults we impose that are not always valid.

So, if a track event's setup changes these start or end, they are correct in popcorn, and the event fires them at the updated time, but our UI is off. In our case we needed this for sequencer, to enforce the time to be whatever was in the media's duration, but I abstracted it to allow any plaugin that may need to do this the ability to do it.

src/core/popcorn-wrapper.js
ScottDowne: This toggles the track event view's display from audio to video if the media is hidden. Hidden is the inverse of audio only, and if so, don't call it a video but call it audio.

This is pretty much icing, and a nice to have.

src/timeline/status.js
ScottDowne, line 48: At some point we had a way to set the time right in the controls to allow time changes.

It seems we never actually set it, and ended up displaying NaN or something random when the player tried to do anything.

src/core/trackevent.js
ScottDowne, line 177: Track events can now be updated async, because they may be waiting for media.

I needed to change the order of things so views and super scrollbar views are updated after the async is done.

See: https://github.com/ScottDowne/butter/commit/cf803fa0384dc1739b35e0eceda472ef6c2041b1

css/media-editor.less
ScottDowne, line 14: chrome?

ScottDowne, line 14: I tried in chrome and the debugger simply did not let me change it.

ScottDowne, line 14: Got it to change, and it's changing on me when I change the css.

src/editor/media-gallery-editor.js
ScottDowne, line 64: Done.

src/main.js
ScottDowne: I want this in another ticket.

If == is that bad, I'll do all the proper checks instead. Easier than changing a function signature more than I need to.

src/timeline/media.js
mjschranz, line 225: Sure. File away.

src/editor/media-gallery-editor.js
ScottDowne: This was done. Not sure why this is still open.

src/editor/media-gallery-editor.js
ScottDowne: Also done.

templates/assets/editors/sequencer/sequencer.js
ScottDowne: This was done. Seems the comment is left open.

src/core/media.js
```diff
@@ -573,24 +573,59 @@
...
+                if ( importData.url && _duration >= 0 ) {
+                  url = importData.url;
+                  if ( !Array.isArray( url ) ) {
+                    url = [ url ];
+                  }
+                  // If source is a single array and of type null player, don't bother making a sequence.
+                  if ( url.length > 1 || !( /#t=\d*,?\d+?/ ).test( url[ 0 ] ) ) {
+                    // grab first source as main source.
+                    source.push( URI.makeUnique( url.shift() ).toString() );
+                    for ( var i = 0; i < url.length; i++ ) {
```
cadecairos: `i` is already defined in this scope, above on line 593

src/editor/media-gallery-editor.js
```diff
@@ -0,0 +1,233 @@
...
+    thumbnailBtn.src = data.thumbnail;
+
+    el.classList.add( "mg-" + data.type );
+    el.classList.add( "new" );
+
+    if ( data.denied ) {
+      el.querySelector( ".mg-error" ).innerHTML = "Embedding disabled by request";
+    }
+
+    setTimeout( function() {
```
cadecairos: `setTimeout(function() {`

src/editor/media-gallery-editor.js
```diff
@@ -0,0 +1,233 @@
...
+      } else {
+        addTrackEvent();
+      }
+    }
+
+    thumbnailBtn.addEventListener( "click", addEvent, false );
+
+    deleteBtn.addEventListener( "click", function() {
+      thumbnailBtn.removeEventListener( "click", addEvent, false );
+      _galleryList.removeChild( el );
```
cadecairos, line 149: Have we considered removing instances of sequencer with the same source as the clip being deleted?

src/editor/media-gallery-editor.js
```diff
@@ -0,0 +1,233 @@
...
+    }, false );
+
+    if ( _galleryList.firstChild ) {
+      _galleryList.insertBefore( el, _galleryList.firstChild );
+    } else {
+      _galleryList.appendChild( el );
```
cadecairos: You don't need this if-else, if the second argument to `insertBefore` is null it just add the element to the target node. (thus making it the first child)

src/editor/media-gallery-editor.js
```diff
@@ -0,0 +1,233 @@
...
+
+    _urlInput.addEventListener( "focus", onFocus, false );
+    _urlInput.addEventListener( "input", onInput, false );
+    _urlInput.addEventListener( "keydown", onEnter, false );
+
+    _addBtn.addEventListener( "click", addMediaToGallery, false );
+    _cancelBtn.addEventListener( "click", resetInput, false );
+
+    _durationInput.addEventListener( "keydown", onDurationChange, false );
+    _durationInput.addEventListener( "blur", function( e ) {
```
cadecairos: Why are all the other listeners named functions and this is anon? Might as well be consistent and define it above.

cadecairos: or below, under `onDurationChange`

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+
+        container.style.zIndex = 0;
+        container.className = "popcorn-sequencer";
+        container.style.position = "absolute";
+        container.style.width = "100%";
+        container.style.height = "100%";
+        container.style.top = 0;
+        container.style.left = 0;
+
+        if ( target ) {
```
humphd: This is unnecesary, as lines 37-8 above establish target if it's missing.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+
+      options.readyEvent = function() {
+        options.failed = false;
+        options.p.off( "loadedmetadata", options.readyEvent );
+        options.ready = true;
+        options._container.style.width = ( options.width || "100" ) + "%";
+        options._container.style.height = ( options.height || "100" ) + "%";
+        options._container.style.top = ( options.top || "0" ) + "%";
+        options._container.style.left = ( options.left || "0" ) + "%";
+        _this.on( "volumechange", options._volumeEvent );
```
humphd, line 93: Could `options` ever lack an `_volumeEvent` function?  Should you guard here?  I guess at worst it will silently fail, but that kind of sucks, and we seem to have various volume related bugs.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
+// PLUGIN: sequencer
```
humphd, line 1: Let's add an overall big comment explaining the entire way this thing works.  Give a 10,000 foot view of your approach, how a Sequencer instance is really a Popcorn instance, how you forward and use events, etc.  Anyone having to understand this code (like me), needs it.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+      options.hideLoading = function() {
+        document.querySelector( ".loading-message" ).classList.remove( "show-media" );
+      };
+      if ( !options.from || options.from > options.duration ) {
+        options.from = 0;
+      }
+
+      options.readyEvent = function() {
+        options.failed = false;
+        options.p.off( "loadedmetadata", options.readyEvent );
```
humphd: I think something more descriptive and helpful than `options.p` is needed here.  Try to explain what this is via the name.  It's the Sequencer's Popcorn instance, so convey that somehow.

templates/assets/plugins/sequencer/popcorn.sequencer.js
humphd, line 1: Per our discussion on irc, let's rename this plugin from `sequencer` to `clip`, which is short for Media Clip.

templates/assets/plugins/sequencer/popcorn.sequencer.js
humphd, line 93: I see the issue, it's declared below.  I think you should move `options.readyEvent` down so the `options._volumeEvent` code gets declared first, making readability much easier.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+      };
+
+      options.sourceToArray = function() {
+        // If our src is not an array, create an array of one.
+        options.source = typeof options.source === "string" ? [ options.source ] : options.source;
+      };
+
+      options.fail = function() {
+        if ( options.ready && !options.denied ) {
+          return;
```
humphd: Silent fails scare me.  Can we do better?

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+          _this.play();
+        }
+      };
+
+      options.tearDown = function() {
+        _this.off( "volumechange", options._volumeEvent );
+        if ( options.p ) {
+          // XXX: pull the SoundCloud iframe element out of our video div, and quarantine
+          // so we don't delete it, and block loading future SoundCloud instances. See above.
+
```
humphd: Remove this line, capitalize the 't' in the next sentence.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+          // this is also fixing an issue in youtube, so we do it for all medias with iframes now.
```
humphd: "fixing an issue" -- give more detail.  What issue does it fix?

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+          options._container.style.zIndex = +options.zindex;
+        }
+        if ( options.playWhenReady ) {
+          _this.play();
+        }
+      };
+
+      options.tearDown = function() {
+        _this.off( "volumechange", options._volumeEvent );
+        if ( options.p ) {
```
humphd: Do you remove p from options?  When can this happen?

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+          // XXX: pull the SoundCloud iframe element out of our video div, and quarantine
+          // so we don't delete it, and block loading future SoundCloud instances. See above.
+
+          // this is also fixing an issue in youtube, so we do it for all medias with iframes now.
+          var iframeParent = options.p.media.parentNode,
+              iframe = iframeParent.querySelector( "iframe" ) || iframeParent.querySelector( "video" ) || iframeParent.querySelector( "audio" );
```
humphd: Better name for this, since it might also be a media element.  Perhaps `mediaElem`?

src/editor/media-gallery-editor.js
ScottDowne: Hm, doesn't seem to work with it removed.

I looked at the spec, and it says you are right, but something seems to be not right with it on chrome and ff.

ScottDowne: Duh, never mind. I didn't remove the if, just the else.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
+// PLUGIN: sequencer
+
+(function ( Popcorn ) {
+
+  // XXX: SoundCloud has a bug (reported by us, but as yet unfixed) which blocks
```
humphd, line 5: Let's rework this comment, since you're using this for multiple video types now vs. just SoundCloud.  Also, please join the two comments into one.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+    quarantine.style.height = "0px";
+    quarantine.style.overflow = "hidden";
+    quarantine.style.visibility = "hidden";
+    document.body.appendChild( quarantine );
+
+    getIframeQuarantine.instance = quarantine;
+    return quarantine;
+  }
+
+  var MEDIA_LOAD_TIMEOUT = 10000;
```
humphd: How did you arrive at this value?

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+          // this is also fixing an issue in youtube, so we do it for all medias with iframes now.
+          var iframeParent = options.p.media.parentNode,
+              iframe = iframeParent.querySelector( "iframe" ) || iframeParent.querySelector( "video" ) || iframeParent.querySelector( "audio" );
+          if ( iframe ) {
+            getIframeQuarantine().appendChild( iframe );
+          }
+          options.p.destroy();
+        }
+
+        // Tear-down old instances, special-casing SoundCloud removal, see above.
```
humphd: Looks like you no longer special-case SoundCloud, so update this comment.

templates/assets/plugins/sequencer/popcorn.sequencer.js
ScottDowne, line 93: I do not see how. It is created in _setup, and never changed.

ScottDowne, line 93: The above comment was for your first comment.

Looking into the second now.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+          clearTimeout( options.loadTimeout );
+        }
+        // if the video is denied for any reason, most cases youtube embedding disabled,
+        // don't bother waiting and display fail case.
+        if ( options.denied ) {
+          options.fail();
+        } else {
+          options.loadTimeout = setTimeout( options.fail, MEDIA_LOAD_TIMEOUT );
+        }
+        options.p = Popcorn.smart( options._container, options.source, { frameAnimation: true } );
```
humphd: So `options.p` depends on `options.addSource` being called before other methods that need it.  Another good place to do some comments, especially at the top.  Methods that have to happen in order or bug ridden; but I understand why you're doing it here.  Just be clear about it, and maybe guard against it in other spots vs. assuming `options.p` exists?

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+        options._container.style.height = "100%";
+        if ( options.p.media.readyState >= 1 ) {
+          options.readyEvent();
+        } else {
+          options.p.on( "loadedmetadata", options.readyEvent );
+        }
+      };
+      options.setupContainer();
+      if ( options.source ) {
+        options.sourceToArray();
```
humphd: This seems fragile, again you have an order required to using methods.  Why not make `options.source` a JS getter which does this automatically?

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+        if ( options.fallback ) {
+          if ( !Array.isArray( options.fallback ) ) {
```
humphd: You've got a couple of ways you're checking for arrays, and you probably want to favour this one.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+
+      options._startEvent = function() {
+        // wait for this seek to finish before displaying it
+        // we then wait for a play as well, because youtube has no seek event,
+        // but it does have a play, and won't play until after the seek.
+        // so we know if the play has finished, the seek is also finished.
+        var seekedEvent = function () {
+          var playedEvent = function() {
+            // We've managed to seek, clear any pause fallbacks.
+            //clearTimeout( seekTimeout );
```
humphd: Don't leave this commented and in the code.

templates/assets/plugins/sequencer/popcorn.sequencer.js
ScottDowne: Ah, this is a cancel of the fail. Might be more obvious if I clearTimeout on ready. Making that return not even needed.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+            if ( options.active ) {
+              options._volumeEvent();
+            }
+          };
+          options.p.off( "seeked", seekedEvent );
+          options.p.on( "play", playedEvent );
+          options.p.play();
+        };
+        options.p.on( "seeked", seekedEvent);
+        options.p.currentTime( _this.currentTime() - options.start + (+options.from) );
```
humphd: Might be good to check the range on this value before you drop it into currentTime, so it doesn't under/over-flow the video duration.

src/util/mediatypes.js
```diff
@@ -0,0 +1,125 @@
...
+        // http://www.youtube.com/watch?v=p_7Qi3mprKQ
+        // Or at the end of the url like this:
+        // http://youtu.be/p_7Qi3mprKQ
+        id = parsedUri.queryKey.v || parsedUri.directory.replace( "/", "" );
+        if ( !id ) {
+          return;
+        }
+
+        xhrURL = "https://gdata.youtube.com/feeds/api/videos/" + id + "?v=2&alt=jsonc&callback=?";
+        Popcorn.getJSONP( xhrURL, function( resp ) {
```
cadecairos, line 49: we shouldn't rely on Popcorn utility methods.. weren't we going to add this to xhr.js?

src/util/mediatypes.js
```diff
@@ -0,0 +1,125 @@
...
+          });
+        });
+      } else if ( type === "vimeo" ) {
+        parsedUri = URI.parse( baseUrl );
+        splitUriDirectory = parsedUri.directory.split( "/" );
+        id = splitUriDirectory[ splitUriDirectory.length - 1 ];
+        xhrURL = "http://vimeo.com/api/v2/video/" + id + ".json?callback=?";
+        Popcorn.getJSONP( xhrURL, function( respData ) {
+          respData = respData && respData[ 0 ];
+          if ( !respData ) {
```
cadecairos, line 99: Does this protect against private Vimeo videos?

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+          if ( !Array.isArray( options.fallback ) ) {
+            options.fallback = [ options.fallback ];
+          }
+          options.source = options.source.concat( options.fallback );
+        }
+        options.addSource();
+      }
+
+      options._startEvent = function() {
+        // wait for this seek to finish before displaying it
```
humphd, line 212: This seeked/play/play-event trick is YT specific, and needs to go in the wrapper.  Try to only do things here that are wrapper-agnostic.  Let's move all such hacks into the wrappers, where we can properly test and isolate special cases.  If you find yourself doing wrapper specific things here, assume you're doing it wrong so we don't grow hairs in this part of the code.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+        };
+        options.p.on( "seeked", seekedEvent);
+        options.p.currentTime( _this.currentTime() - options.start + (+options.from) );
+      };
+
+      options._playWhenReadyEvent = function() {
+        options.playWhenReady = true;
+      };
+
+      options._seqPlayEvent = function() {
```
humphd: Make sure you rename all the sequencer, seq\* names when you change to clip.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+        if ( _this.paused() ) {
+          setTimeout( function() {
```
humphd: Why are you using timeouts?  Let's just use events, and if the events aren't working up here, fix the bugs in the wrappers.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+            options.p.off( "play", options._seqPlayEvent );
+            _this.play();
+            options.p.on( "pause", options._seqPauseEvent );
+          }, 0 );
+        }
+      };
+
+      options._playEvent = function() {
```
humphd: It would be helpful to have better naming (and a big comment explaining the names) as to why you have multiple versions of each playEvent, pauseEvent...

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+        if ( options.p.paused() ) {
+          options.p.play();
+        }
+      };
+
+      options._seqPauseEvent = function() {
+        if ( !_this.paused() ) {
+          setTimeout( function() {
```
humphd: Same question.  Don't do this, it hides bugs, causes errors to get eaten in background threads, etc.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+        }
+      };
+
+      options._pauseEvent = function() {
+        if ( !options.p.paused() ) {
+          options.p.pause();
+        }
+      };
+
+      options._volumeEvent = function() {
```
humphd: Another clarity issue.  You have things like `options.mute` and `options.p.mute` and it's easy to get lost here.  What if you namespace things such that you have `options.onething` and `options.otherthing` and you put all your properties and methods into the appropriate object namespace?

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+            options.p.unmute();
+            options.p.volume( ( options.volume / 100 ) * _this.volume() );
+          } else {
+            options.p.mute();
+          }
+        }
+      };
+
+      options._seekedEvent = function() {
+        options.p.currentTime( _this.currentTime() - options.start + (+options.from) );
```
humphd: Validate this time so it's always safely in range.

humphd: Another thought: you seem to do this same operation a few times, should it go in a function you reuse?

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+    quarantine.style.overflow = "hidden";
+    quarantine.style.visibility = "hidden";
+    document.body.appendChild( quarantine );
+
+    getIframeQuarantine.instance = quarantine;
+    return quarantine;
+  }
+
+  var MEDIA_LOAD_TIMEOUT = 10000;
+
```
cadecairos: there's an extra space here :D

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+      options._seekedEvent = function() {
+        options.p.currentTime( _this.currentTime() - options.start + (+options.from) );
+      };
+
+      options.toString = function() {
+        return options.title || options.source || "";
+      };
+
+      if ( options.duration > 0 && options.end - ( options.start - ( +options.from ) ) > options.duration ) {
```
humphd: This could use parens or a newline to make it more readable:

```
if ( options.duration > 0 &&
     options.end - ( options.start - ( +options.from ) ) > options.duration )
```

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+        options.end = options.duration + ( options.start - ( +options.from ) );
+      }
+    },
+    _update: function( options, updates ) {
+      if ( updates.hasOwnProperty( "duration" ) ) {
```
humphd, line 331: This function does a lot of state changes, add comments to explain what's up.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+      if ( updates.source ) {
+        options.sourceToArray();
+        if ( options.fallback ) {
+          if ( !Array.isArray( options.fallback ) ) {
+            options.fallback = [ options.fallback ];
+          }
+          if ( !Array.isArray( updates.source ) ) {
+            updates.source = [ updates.source ];
+          }
+          options.source = options.source.concat( options.fallback );
```
humphd: This whole block seems like code I've read elsewhere.  Try to not repeat.

templates/assets/plugins/sequencer/popcorn.sequencer.js
humphd, line 331: I appreciate the use of `hasOwnProperty` but I think it's overkill in here and makes the code longer and harder to read.  Just check for the property in each case, unless you know you need to do this.

templates/assets/plugins/sequencer/popcorn.sequencer.js
```diff
@@ -0,0 +1,506 @@
...
+          options.setupContainer();
+          this.on( "play", options._playWhenReadyEvent );
+          if ( !this.paused() ) {
+            options.playWhenReady = true;
+            this.pause();
+            if ( options.p && !options.p.paused() ) {
+              options.p.pause();
+            }
+          }
+          options.addSource();
```
humphd: Again, I feel like I've read this code once already.  Can this get moved to somewhere common?

src/util/mediatypes.js
```diff
@@ -0,0 +1,125 @@
...
+          callback({
+            source: baseUrl,
+            type: type,
+            thumbnail: respData.thumbnail_small,
+            duration: respData.duration,
+            title: respData.title
+          });
+        });
+      } else if ( type === "html5" ) {
+        testEl = document.createElement( "video" );
```
humphd: Same naming issue here, testEl is no good.  This is a `video` or `videoElem`.

src/util/mediatypes.js
```diff
@@ -0,0 +1,125 @@
...
+      var id,
+          parsedUri,
+          splitUriDirectory,
+          xhrURL,
+          testEl;
+
+      type = type || this.checkUrl( baseUrl );
+      callback = callback || function(){};
+
+      if ( type === "youtube" ) {
```
humphd, line 37: Let's try to move away from this.  I don't want to have knowledge of the wrappers/media types up here.  We should put this code into functions on the wrapper, exposed on it's `prototype._utils` object.  We can land it like this, but let's clean things up so that you get the wrapper to parse the URL, extract the ID or whatever, and give you back an object you can use.

src/util/mediatypes.js
```diff
@@ -0,0 +1,125 @@
+/* This Source Code Form is subject to the terms of the MIT license
+ * If a copy of the MIT license was not distributed with this file, you can
+ * obtain one at https://raw.github.com/mozilla/butter/master/LICENSE */
+
+"use strict";
+
+define( [ "util/xhr", "util/uri" ],
+  function( XHR, URI ) {
+
+  var REGEX_MAP = {
```
humphd, line 10: This stuff should disappear into the wrappers, delegate the parsing to them.

templates/assets/plugins/sequencer/popcorn.sequencer.js
ScottDowne: If a track event is created without a source, we don't create anything that needs to be torn down.

Also, if a teardown happens before the p is ready, I should probably tear it down once ready.

templates/assets/plugins/sequencer/popcorn.sequencer.js
ScottDowne: No real reason. Seemed like a nice round number.

I have no data to go by, so I picked something to start with.

ScottDowne: I'm starting high here, and we can iterate it down. Assuming the worst.

templates/assets/plugins/sequencer/popcorn.sequencer.js
ScottDowne: I'm doing some more work here.

templates/assets/plugins/sequencer/popcorn.sequencer.js
ScottDowne: This is a hard one.

This is a fix I found after a coupld days of debugging, and I'm still having some issues fully understanding it.

Without it, you can add two youtube clips, and if you click on the youtube element itself fast enough, you can crash flash at the worst, at the best you hang the browser. I could do it in Windows too, so it wasn't Linux.

The key is two youtube elements cause it to happen, and clicking the element faster than it can be processed.

I planned on going back to this, but I found something that worked and needed to move on for now.

I can try to filter this out inside the wrapper play and pause events coming from youtube, would probably make more sense.

css/media-editor.less
```diff
@@ -22,77 +50,200 @@
...
+      margin-left: -7px;
+      left: 50%;
+      background: url( "../resources/icons/glyphiconshalflings-white.png" ) -408px -96px no-repeat;
     }
   }
+  .mg-html5 .mg-thumbnail {
+    > video {
+      height: auto;
+    }
+    background-image: url( "../resources/icons/media.png" ) center no-repeat;
```
cadecairos: lint doesn't like this for some reason, complains about "center"

templates/basic/style.less
```diff
@@ -55,43 +55,51 @@ body.editor-minimized .body-wrapper {
...
+  .transition( opacity 0.35s, visibility 0.35s );
 
-    opacity: 0;
-    visibility: hidden;
+  height: 3em;
+  background: rgba(255,255,255,1) url( "/resources/icons/big-loader.gif" ) center no-repeat;
 
-    .transition( opacity 0.35s, visibility 0.35s );
+  width: 100%;
+  height: 100%;
```
cadecairos, line 69: duplicate declaration of height - see line 66

src/core/media.js
```diff
@@ -573,24 +573,60 @@
...
+            if ( importData.duration >= 0 ) {
+              _duration = importData.duration;
+              _this.url = "#t=," + _duration;
             }
             if( importData.tracks ){
               var importTracks = importData.tracks;
               if( Array.isArray( importTracks ) ) {
-                for ( var i = 0, l = importTracks.length; i < l; ++i ) {
-                  var newTrack = new Track();
+                for ( i = 0, l = importTracks.length; i < l; ++i ) {
```
cadecairos, line 594: duplicate declaration of `l`

cadecairos, line 594: my bad, it's not declared at all.

src/editor/sequencer-editor.js
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        function checkboxCallback( trackEvent, updateOptions ) {
+          trackEvent.update( updateOptions );
+        }
+
+        function fromCallback( trackEvent, updateOptions ) {
+          updateOptions.from = Time.toSeconds( updateOptions.from );
+          if ( updateOptions.from >= trackEvent.popcornOptions.duration ) {
+            updateOptions.from = trackEvent.popcornOptions.from
```
cadecairos: semicolon

src/editor/sequencer-editor.js
```diff
@@ -0,0 +1,137 @@
...
+        _pluginOptions.source.element.value = URI.stripUnique( _trackEvent.popcornOptions.source[ 0 ] ).toString();
+      }
+      if ( _trackEvent.popcornOptions.fallback ) {
+        _pluginOptions.fallback.element.value = URI.stripUnique( _trackEvent.popcornOptions.fallback[ 0 ] ).toString();
+      }
+      _this.setErrorState( false );
+    }
+
+    // Extend this object to become a TrackEventEditor
+    Butter.Editor.TrackEventEditor.extend( _this, butter, rootElement, {
```
cadecairos: Lint complains about the reference to Butter here.

src/editor/sequencer-editor.js
```diff
@@ -0,0 +1,137 @@
...
+
+      function callback( elementType, element, trackEvent, name ) {
+        _pluginOptions[ name ] = { element: element, trackEvent: trackEvent, elementType: elementType };
+      }
+
+      function attachHandlers() {
+        var key,
+            option;
+
+        function sourceCallback( trackEvent, updateOptions, prop ) {
```
cadecairos: prop is not used

src/layouts/media-editor.html
```diff
@@ -1,52 +1,53 @@
...
-    <a class="butter-media-apply butter-btn btn-green butter-disabled">Apply</a>
-    <label class="butter-clear-track-events butter-disabled" >
-      <input type="checkbox"/>
-      Clear track events
-    </label>
-  </span>
 </div>
+
+<!-- This is added to media-gallery for each new media item -->
+<li class="media-gallery-item">
```
cadecairos, line 44: html5 lint doesn't like how this li is defined outside of a body tag... have we worked around this elsewhere?

templates/basic/style.less
secretrobotron, line 69: Did css-lint not pick this up?

cadecairos, line 69: It did, I'm just making note of the issues in the pull request.

secretrobotron, line 69: That's nice of you :). Just making sure it's still working.

PR #1466, opened 2013-02-01
[t3068] Prevent super scrollbar from resetting position and zoom level o...

mjschranz:
...n media change


mjschranz:
Use #1471 instead.

PR #1469, opened 2013-02-04
[t3119] Prevent new events from generateSafeTrackEvent increasing curren...

mjschranz:
...tTime

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -4,8 +4,7 @@
 
 (function () {
 
-  var DEFAULT_TRACKEVENT_OFFSET = 0.01,
-      WARNING_WAIT_TIME = 500;
+  var WARNING_WAIT_TIME = 500;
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -195,9 +194,7 @@
           type: type
         });
 
-        if( _currentMedia.currentTime < _currentMedia.duration - DEFAULT_TRACKEVENT_OFFSET ){
-          _currentMedia.currentTime += DEFAULT_TRACKEVENT_OFFSET;
-        }
+        _currentMedia.currentTime = popcornOptions.start;
```
ScottDowne: Pretty sure you do not need this line.

This is happening elsewhere. Going to see if I can find the line, as maybe it is better to do it here, but we shouldn't be doing it twice.

PR #1472, opened 2013-02-04
[t3118] Prevent certain key presses from firing multiple times on keydow...

mjschranz:
...n

### DIFF
--- a/src/core/media.js
+++ b/src/core/media.js
@@ -110,8 +110,8 @@
       };
 
       this.clear = function(){
-        while( _tracks.length > 0 ){
-          _this.removeTrack( _tracks[ 0 ] );
+        for ( var i = _tracks.length - 1; i >= 0; i-- ) {
+          _this.removeTrack( _tracks[ i ] );
         }
       };
 
@@ -152,9 +152,6 @@
           throw "Track already belongs to a Media object. Use `media.removeTrack` prior to this function.";
         }
 
-        // Sort tracks first, so we can guarantee their ordering
-        _this.sortTracks( true );
-
         // Give new track last order since it's newest
         track.order = _tracks.length;
 
@@ -178,27 +175,21 @@
           throw "Track already belongs to a Media object. Use `media.removeTrack` prior to this function.";
         }
 
-        // Sort tracks first, so we can guarantee their ordering
-        _this.sortTracks( true );
-
         var idx = _orderedTracks.indexOf( otherTrack );
 
         if ( idx > -1 ) {
           // Give new track last order since it's newest
-          newTrack.order = _tracks.length;
+          newTrack.order = idx;
 
           // Insert new track
           _orderedTracks.splice( idx, 0, newTrack );
 
-          // Fix all the order properties on subsequent tracks
-          for ( var i = idx, l = _orderedTracks.length; i < l; ++i ) {
-            _orderedTracks[ i ].order = i;
-          }
-
           setupNewTrack( newTrack );
 
           _this.dispatch( "trackadded", newTrack );
-          _this.dispatch( "trackorderchanged", _orderedTracks );
+
+          // Sort tracks after added one to update their order.
+          _this.sortTracks( idx + 1 );
 
           addNewTrackTrackEvents( newTrack );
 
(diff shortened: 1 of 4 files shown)


src/ui/ui.js
```diff
@@ -17,11 +17,15 @@ define( [ "core/eventmanager", "./toggler",
...
-    "INPUT",
-    "VIDEO",
-    "AUDIO"
-  ];
+        "TEXTAREA",
+        "INPUT",
+        "VIDEO",
+        "AUDIO"
+      ],
+      __unwantedHoldKeys = [
```
ScottDowne: Change this variable name to disabledKeyRepeats and add space key to it, maybe?

src/ui/ui.js
```diff
@@ -546,24 +550,35 @@ define( [ "core/eventmanager", "./toggler",
...
+
+    function unbindKeyListeners() {
       window.removeEventListener( "keydown", onKeyDown, false );
+      window.removeEventListener( "keyup", onKeyUp, false );
     }
 
-    function bindKeyDownListener() {
+    function bindKeyListeners() {
       window.addEventListener( "keydown", onKeyDown, false );
+      window.addEventListener( "keyup", onKeyUp, false );
```
ScottDowne: You should only add this event inside the onKeyDown. This way, you have less events being held. Less prone to bugs too.

src/ui/ui.js
```diff
@@ -546,24 +550,35 @@ define( [ "core/eventmanager", "./toggler",
...
     }
 
-    function unbindKeyDownListener() {
+    function onKeyUp() {
+      window.addEventListener( "keydown", onKeyDown, false );
+    }
+
+    function unbindKeyListeners() {
       window.removeEventListener( "keydown", onKeyDown, false );
+      window.removeEventListener( "keyup", onKeyUp, false );
```
ScottDowne: onKeyUp should unbind itself after usage, and not here.

src/ui/ui.js
```diff
@@ -546,24 +550,35 @@ define( [ "core/eventmanager", "./toggler",
       // this allows backspace and del to do the same thing on windows and mac keyboards
       key = key === 46 ? 8 : key;
       if( processKey[ key ] && !eTarget.isContentEditable && __unwantedKeyPressElements.indexOf( eTarget.nodeName ) === -1 ){
+
+        if ( __unwantedHoldKeys.indexOf( key ) > -1 ) {
+          window.removeEventListener( "keydown", onKeyDown, false );
```
ScottDowne: Add the keyUp listener here.

src/ui/ui.js
```diff
@@ -546,24 +550,35 @@ define( [ "core/eventmanager", "./toggler",
...
+        }
+
         processKey[ key ]( e );
       } // if
     }
 
-    function unbindKeyDownListener() {
+    function onKeyUp() {
+      window.addEventListener( "keydown", onKeyDown, false );
```
ScottDowne: Like said under this comment, should remove keyup event here, and not in unbind.

src/ui/ui.js
ScottDowne: Actually, consider adding esc and del to this list too.

src/core/media.js
```diff
@@ -110,8 +110,8 @@
       };
 
       this.clear = function(){
-        while( _tracks.length > 0 ){
-          _this.removeTrack( _tracks[ 0 ] );
+        for( var i = _tracks.length - 1; i >= 0; i-- ) {
```
jbuck: `for (`

jbuck: Also, can't you just set `_tracks.length = 0;` to clear an array? or does `removeTrack()` do more work than that?

src/core/media.js
```diff
@@ -238,9 +231,10 @@
             "trackeventdeselected"
           ]);
           track.setPopcornWrapper( null );
-          _this.sortTracks();
           track._media = null;
           _this.dispatch( "trackremoved", track );
+          _orderedTracks.splice( orderedIndex, 1 );
```
jbuck: Wait, we remove a track after we send an event for it? That order seems wrong...

ScottDowne: Agreed, this seems wrong.

Did you guys discuss this and come to an agreement?

src/core/media.js
```diff
@@ -238,9 +231,10 @@
...
+          _this.sortTracks( orderedIndex );
```
ScottDowne: Do we need to sort after a splice? Should be fine, no?

ScottDowne: Ah, I see. It is adding a order to the track event's data. Makes sense.

PR #1473, opened 2013-02-04
[#3121] Handle event selection and deselection before firing related events

cadecairos:
https://webmademovies.lighthouseapp.com/projects/65733-popcorn-maker/tickets/3123

### DIFF
--- a/src/core/trackevent.js
+++ b/src/core/trackevent.js
@@ -356,9 +356,11 @@ define( [ "./logger", "./eventmanager", "./observer",
           _selected = val;
           _view.selected = _selected;
           if ( _selected ){
+            _this.notify( "selected" );
             _this.dispatch( "trackeventselected" );
           }
           else {
+            _this.notify( "deselected" );
             _this.dispatch( "trackeventdeselected" );
           } //if
         }
--- a/src/main.js
+++ b/src/main.js
@@ -248,13 +248,19 @@
         _sortedSelectedEvents = _selectedEvents.slice().sort( trackEventTimeSortingFunction );
       }
 
-      function onTrackEventSelected( e ) {
-        _selectedEvents.push( e.target );
+      function onTrackEventSelected( notification ) {
+        var trackEvent = notification.origin;
+        for ( var i = _selectedEvents.length - 1; i >= 0; i-- ) {
+          if ( _selectedEvents[ i ] === trackEvent ) {
+            return;
+          }
+        }
+        _selectedEvents.push( trackEvent );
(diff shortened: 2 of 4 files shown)


src/core/trackevent.js
```diff
@@ -42,6 +42,7 @@ define( [ "./logger", "./eventmanager", "./observer",
     options = options || {};
 
     var _this = this,
+        _butter = window.Butter.app,
```
secretrobotron: Please don't do this; it's not sustainable. Ideally, `butter` shouldn't even be referenced inside this file.

Instead, try employing the observer object to do your bidding. See https://github.com/mozilla/butter/blob/master/src/core/trackevent.js#L142. `notify` works like `dispatch`, except all handlers are called immediately instead of relying on the browser's event loop. So, `src/core/main.js`, you can subscribe to some notification you emit here, keeping `TrackEvent` decoupled from the rest of the code (leading to profit, happiness, and word peace).

src/main.js
```diff
@@ -264,6 +270,10 @@
 
       function onTrackEventAdded( e ) {
         var trackEvent = e.data;
+
+        trackEvent.subscribe( "selected", onTrackEventSelected );
+        trackEvent.subscribe( "deselected", onTrackEventDeSelected);
```
mjschranz: Missed a space.

src/main.js
```diff
@@ -272,6 +282,10 @@
       function onTrackEventRemoved( e ) {
         var trackEvent = e.data,
             idx = _selectedEvents.indexOf( trackEvent );
+
+        trackEvent.unsubscribe( "selected", onTrackEventSelected );
+        trackEvent.unsubscribe( "deselected", onTrackEventDeSelected);
```
mjschranz: Space!

src/core/trackevent.js
```diff
@@ -344,8 +344,8 @@ define( [ "./logger", "./eventmanager", "./observer",
        * Specifies the state of selection. When true, this TrackEvent is selected.
        *
        * @malleable: Yes.
-       * @event trackeventselected: Dispatched when selected state changes to true.
-       * @event trackeventdeselected: Dispatched when selected state changes to false.
+       * @event trackeventselection: Dispatched when selected state changes to true.
+       * @event trackeventdeselection: Dispatched when selected state changes to false.
```
mjschranz: Why are the event names changed when they aren't what is actually dispatched?

cadecairos: just kruft from an old commit. ill change em

PR #1476, opened 2013-02-05
added CSS transition for the post-roll container - 1s delay, 1s fadein

Pomax:
also modified embed.js to not use show() and hide() for the post roll container, because they flip display:block and display:none, which clobbers CSS transitions completely. I only refactored this for the post roll container, but I'll file an issue to refactor them out of existence.

### DIFF
--- a/css/embed.less
+++ b/css/embed.less
@@ -341,7 +341,17 @@ body {
   left: 0;
   width: 100%;
   height: 100%;
-  display: none;
+  display: block;
+  opacity: 0;
+
+  // No transition from "visible" to "invisible"
+  .transition( none );
+
+  &.embed-dialog-open {
+    opacity: 1;
+    // Fade-in transition from "invisible" to visible"
+    .transition( opacity 1s ease 1s );
+  }
 }
 
 #post-roll,
@@ -471,6 +481,7 @@ body {
   }
 
 } // post-roll
+
 
 
 /*********************************************************
--- a/src/embed.js
+++ b/src/embed.js
@@ -79,9 +79,9 @@ function init() {
       popcorn.pause();
     }
 
-    addStateClass( "embed-dialog-open" );
+    setStateClass( "embed-dialog-open" );
     hide( "#controls-big-play-button" );
-    hide( "#post-roll-container" );
+    clearStateClass();
     show( "#share-container" );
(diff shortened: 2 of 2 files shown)


Pomax:
I have some commented code in this, if this gets the approval from Kate I shall remove the comments and finalise the pull request

src/embed.js
```diff
@@ -146,6 +147,15 @@ function init( window, document ) {
     el.classList.add( state );
   }
 
+  function removeStateClass( state) {
+    var el = $( "#post-roll-container" );
+
+    while ( el.classList.contains( state ) ) {
```
k88hudson: Don't think this is necessary, all you need here is `el.classList.remove( state );` -- the `remove` method will remove the class if it exists and simply do nothing if it doesn't.

css/embed.less
```diff
@@ -341,7 +341,14 @@ body {
   left: 0;
   width: 100%;
   height: 100%;
-  display: none;
+  display: block;
+  opacity: 0.0;
```
k88hudson: Our style is typically `0` here

css/embed.less
```diff
@@ -341,7 +341,14 @@ body {
...
+
+  // FADE-IN WHEN SHOWING POST ROLL
+  .transition(opacity 1s ease 1s);
+  &.embed-dialog-open {
+    opacity: 1.0;
+  }
```
k88hudson: `1`, not `1.0`

css/embed.less
```diff
@@ -341,7 +341,14 @@ body {
   left: 0;
   width: 100%;
   height: 100%;
-  display: none;
+  display: block;
+  opacity: 0.0;
+
+  // FADE-IN WHEN SHOWING POST ROLL
+  .transition(opacity 1s ease 1s);
```
k88hudson: Spacing: `( opacity 1s ease 1s )`

`1s` is actually fine here, I tried it and it feels right.

src/embed.js
```diff
@@ -146,6 +147,15 @@ function init( window, document ) {
     el.classList.add( state );
   }
 
+  function removeStateClass( state) {
```
k88hudson: Spacing on `( state )`

PR #1478, opened 2013-02-06
[t3132] Bail earlier if we can't use the droppable item data

mjschranz:


templates/assets/editors/editorhelper.js
```diff
@@ -359,6 +352,14 @@
               window.webkitURL.revokeObjectURL( imgSrc );
             }
           };
+
+          // If we can't reasonably convert whatever they're dropping on us then bail
+          var exportAs = mimeTypeMapping[ file.type ];
+          if ( !exportAs ) {
+            butter.dispatch( "droppable-unsupported" );
```
jbuck: Is this the right message we want to send to the user? Maybe we need a different event for "media type unsupported"?

mjschranz: It's a general event we dispatch and really only the image editor uses it.

I suppose we could/should change it up.

templates/assets/editors/editorhelper.js
```diff
@@ -359,6 +352,14 @@
               window.webkitURL.revokeObjectURL( imgSrc );
             }
           };
+
+          // If we can't reasonably convert whatever they're dropping on us then bail
```
jbuck: Nit: move this up to line 280-290

PR #1479, opened 2013-02-06
Bug 2954 - adding in FaceBook Open Graph tags

rossbruniges:
Not understanding exactly the desire for the bug I tried to see if by inserting in unique values for each of the project pages (loaded from the JSON file) would actually work, and be something that would be useful.

We can test things with http://developers.facebook.com/tools/debug

### DIFF
--- a/templates/basic/index.html
+++ b/templates/basic/index.html
@@ -1,13 +1,10 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
+  <title>Popcorn Maker</title>
   <meta charset="utf-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
-
-  <title>Popcorn Maker</title>
-
   <meta name="viewport" content="width=device-width">
-
   <link rel="stylesheet" href="style.css">
   <link rel="stylesheet" href="../assets/css/jquery-ui/jquery.ui.butter.css" data-butter-exclude>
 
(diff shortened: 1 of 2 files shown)


templates/basic/index.html
```diff
@@ -43,6 +39,15 @@
       var s = document.getElementsByTagName('script')[0]; s.parentNode.insertBefore(ga, s);
     })();
   </script>
+  <meta id="og-title" property="og:title" content="Popcorn Maker">
```
jbuck: Any reason for the id tags here, and not having them on the `public/index.html` file?

templates/basic/index.html
```diff
@@ -43,6 +39,15 @@
...
+  <meta id="og-description" property="og:description" content="Video beyond the box. Popcorn Maker makes it easy to enhance, remix and share web video. ">
+  <meta id="og-url" property="og:url" content="https://popcorn.webmaker.org">
+  <meta property="og:site_name" content="Mozilla Webmaker">
+  <meta property="og:type" content="video.other">
```
jbuck: I think we want to leave this as `content="website"`, since none of the video attributes are being defined.

templates/basic/index.html
rossbruniges: Copy and paste fail from the 'bad pull request'

templates/basic/index.html
rossbruniges: K

rossbruniges:
Fixed up nits

PR #1483, opened 2013-02-07
Fix #3064 - quick hack of the README.md file to have the contributor info more up-front 

rossbruniges:
Have moved a bunch of the other stuff to the WIKI - take a look at https://github.com/rossbruniges/butter/tree/readme-extra to see how it ends up looking.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Butter
 ======
 
-An SDK for authoring Popcorn projects.
+An online application for authoring Popcorn projects.
 
 Supported Platforms
 -------------------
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,215 +16,69 @@ In the current version we are supporting:
...
------------------
+#### Mac OS X
+Xcode and Command Line Tools package (warning, this is a large download - so grab some tea while you're doing it)
 
-1. `git clone --recursive https://github.com/mozilla/butter.git`
-2. `cd butter`
-3. `npm install`
+#### Windows
+Python 2.5+ and Visual Studio 2010; specifically:
+* uninstall *any and all* "Microsoft Visual C++ 2010 x86/64 Redistributable" copies
```
mjschranz, line 32: I'm wondering if `any and all` should be bolded rather than italicized. Seems like it might stand out more.

rossbruniges, line 32: Probably also a good idea to have a windows user confirm that these actually work if we're going to fine tooth over it.

README.md
```diff
@@ -16,215 +16,69 @@ In the current version we are supporting:
...
+### Installation FAQs
 
-There are two types of storage Cornfield needs to run:
+We've tried to include answers to the commen questions, if you have any more problems please let us know.
 
-* A database: To store user project data, a database is required. Cornfield uses the `Sequelize` middleware, so multiple database options are available. You are responsible for setting up and maintaining a PostgreSQL or MySQL database enrivonment if you choose to use either, so make sure a user has access to a database called `popcorn` for cornfield to run correctly.
+#### When I try and run npm install I see an error containing 'Error: not found: make'
 
-* A data-blob store: To store published projects, Cornfield can use the filesystem, or Amazon's S3. See the configuration options below for setting up this feature for your environment.
+Make sure that you've installed the Xcode command line tools, they are not bundled with Xcode 4.3 and you need to download/install them separately - (http://developer.apple.com/library/ios/#documentation/DeveloperTools/Conceptual/WhatsNewXcode/Articles/xcode_4_3.html)
```
mjschranz, line 65: I'm pretty sure this sort of error could come around for any dev environment people try to get this running with. I'm wondering if it should be changed to either something more general to ensure that people set up their environment correctly or maybe something for each.

rossbruniges, line 65: If we can have some Windows/Linux users check this out that would be awesome. I've only had fellow mac users as my guinea pigs - off which these changes were made.

PR #1485, opened 2013-02-07
[#3076] Projects loaded with 'deleteDefaultEventsOnMediaChange' set to t...

cadecairos:
...rue will remove events on the first media change.

https://webmademovies.lighthouseapp.com/projects/65733/tickets/3076


src/core/project.js
```diff
@@ -12,6 +12,7 @@ define( [ "core/eventmanager", "core/media" ],
     var _this = this,
         _id, _name, _template, _author, _dataObject,
         _publishUrl, _iframeUrl, _remixedFrom,
+        _deleteDefaultEventsOnMediaChange;
```
jbuck: Don't put a `;` here, you need a `,`

PR #1497, opened 2013-02-13
[t3065] Implement missing search type parameter for twitter searches

mjschranz:


templates/assets/plugins/twitter/popcorn.twitter.js
```diff
@@ -142,7 +142,7 @@
       }
       options._container.appendChild( titleText );
 
-      query = ( options.search || options.username ) + numberOfTweets;
+      query = ( options.search || options.username ) + numberOfTweets + options.searchType;
```
cadecairos: Why doesn't the plugin implement the "popular" option for result_type?

https://dev.twitter.com/docs/api/1/get/search

PR #1499, opened 2013-02-14
[t3182] Prevent twitter requests failing if unable to retrieve more twee...

mjschranz:
...ts


src/editor/trackevent-editor.js
```diff
@@ -482,13 +484,20 @@ define([ "util/lang", "util/keys", "util/time", "./base-editor", "ui/widget/tool
...
       if ( element.type === "number" || isNumber ) {
         element.addEventListener( "change", function() {
-          var updateOptions = {},
-              val = element.value;
 
-          val = validateNumber( val );
+          if ( ignoreChange ) {
+            ignoreChange = false;
+          } else {
+            var updateOptions = {},
```
cadecairos: don't declare the variables in this else block

PR #1501, opened 2013-02-15
[t3053] Wikipedia _update function. Fixed resize handle z-indexing

mjschranz:


templates/assets/editors/wikipedia.js
```diff
@@ -10,12 +10,12 @@ EditorHelper.addPlugin( "wikipedia", function( trackEvent ) {
  target = _popcornOptions._target;
 
   if ( window.jQuery ) {
-
+    window.EditorHelper.draggable( trackEvent, _container, target );
     window.EditorHelper.resizable( trackEvent, _container, target, {
       minWidth: 40,
-      minHeight: 40
+      minHeight: 40,
+      handlePositions: "e,se,s"
```
jbuck: Why not have handles on all sides?

PR #1510, opened 2013-02-21
[t3198] Added html5slider.js library to allow for input type range eleme...

mjschranz:
...nt to be used


public/src/editor/default.js
```diff
@@ -54,7 +56,9 @@ define( [ "text!./default.html", "editor/editor", "util/lang" ],
             if ( element.type === "checkbox" ) {
               _this.attachCheckboxChangeHandler( element, trackEvent, name, _this.updateTrackEventSafe );
             }
-            else {
+            else if ( manifestOptions[ name ].type === "slider" ) {
```
cadecairos: To better follow the HTML5 spec, this should be "range" not "slider".

PR #1513, opened 2013-02-22
[3214] Fix statsd prefixing

humphd:


cornfield/lib/metrics.js
```diff
@@ -29,9 +29,10 @@ module.exports = {
       options = {};
       StatsD = NullStatsDClient;
     } else {
-      StatsD = require('node-statsd').StatsD;
+      StatsD = require( 'node-statsd' ).StatsD;
       // If a prefix is given, use it. Otherwise use `<env>.butter.'
-      options.prefix = options.prefix || env + ".butter.";
```
jbuck, line 34: instead of passing in `env` to this class, just use process.env.NODE_ENV directly

cornfield/lib/metrics.js
```diff
@@ -29,9 +29,10 @@ module.exports = {
...
+      options.prefix = options.prefix ||
+                       ( env === 'production' ? 'production' : 'development' ) + ".butter.";
```
jbuck, line 36: ( process.env.NODE_ENV ? process.env.NODE_ENV : 'development' ) + ".butter.";

PR #1515, opened 2013-02-26
[t3174] Save media clips used as data on each project. When projects are...

mjschranz:
... loaded from dashboard/remix clips that were loaded will appear in gallery

### DIFF
--- a/src/editor/media-gallery-editor.js
+++ b/src/editor/media-gallery-editor.js
@@ -25,7 +25,8 @@ define( [ "util/lang", "util/xhr", "util/keys", "util/mediatypes", "editor/edito
 
       _butter,
       _media,
-      _this;
+      _this,
+      _guid = 0;
 
   function toggleAddNewMediaPanel() {
     _parentElement.classList.toggle( "add-media-collapsed" );
@@ -68,7 +69,8 @@ define( [ "util/lang", "util/xhr", "util/keys", "util/mediatypes", "editor/edito
     var el = _GALLERYITEM.cloneNode( true ),
         deleteBtn = el.querySelector( ".mg-delete-btn" ),
         thumbnailBtn = el.querySelector( ".mg-thumbnail" ),
-        thumbnailImg;
+        thumbnailImg,
+        idx = _guid++;
 
     DragNDrop.helper( thumbnailBtn, {
       pluginOptions: {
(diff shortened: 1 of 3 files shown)


src/editor/media-gallery-editor.js
```diff
@@ -81,8 +84,13 @@ define( [ "util/lang", "util/xhr", "util/keys", "util/mediatypes", "editor/edito
       }
     });
 
+    if ( !ignoreClipAdd ) {
+      _media.clipData[ id ] = data;
```
ScottDowne: Why do this as an object with an id? Why not push to an array?

mjschranz: Because then it's easier to know which to remove when they click on the delete buttons.

src/editor/media-gallery-editor.js
```diff
@@ -64,11 +66,12 @@ define( [ "util/lang", "util/xhr", "util/keys", "util/mediatypes", "editor/edito
     }
   }
 
-  function onSuccess( data ) {
+  function onSuccess( data, ignoreClipAdd ) {
```
ScottDowne: What is the purpose of ignoreClipAdd?

Is it so clips don't get added again on a load? If so, why not blow away what's in there and re add it? Or ensure it is not loaded into clipData until it hits here.

src/editor/media-gallery-editor.js
```diff
@@ -202,6 +216,18 @@ define( [ "util/lang", "util/xhr", "util/keys", "util/mediatypes", "editor/edito
 
     _durationInput.addEventListener( "keydown", onDurationChange, false );
     _durationInput.addEventListener( "blur", onBlur, false );
+
+    if ( _initialLoad ) {
```
ScottDowne: No way to make this work inside a mediaready event? That would be ideal.

src/editor/media-gallery-editor.js
ScottDowne: Also, consider that this is no longer only a success callback, it might make sense to dry it out, or make another function, or dry out addMediaToGallery.

src/editor/media-gallery-editor.js
```diff
@@ -81,8 +83,20 @@ define( [ "util/lang", "util/xhr", "util/keys", "util/mediatypes", "editor/edito
...
+      el.classList.add( "new" );
+
+      setTimeout(function() {
+        el.classList.remove( "new" );
+      }, 2000 );
+    }
+
     thumbnailBtn.setAttribute( "data-popcorn-plugin-type", "sequencer" );
     thumbnailBtn.setAttribute( "data-butter-draggable-type", "plugin" );
+    deleteBtn.setAttribute( "data-butter-media-clip", idx );
```
k88hudson: This seems pretty weird to me, why not just attach the event listener here instead of storing it in the DOM?

PR #1521, opened 2013-02-28
[#3200] Display warnings and errors on media gallery for denied and medi...

ScottDowne:
...a is taking too long to load.


src/editor/media-gallery-editor.js
```diff
@@ -169,19 +185,19 @@ define( [ "util/lang", "util/xhr", "util/keys", "util/mediatypes", "editor/edito
...
     }
+    _loadingSpinner.classList.add( "hidden" );
+    _errorMessage.classList.add( "hidden" );
   }
 
   function onEnter( e ) {
     if ( e.keyCode === KeysUtils.ENTER ) {
       e.preventDefault();
-      addMediaToGallery();
+      onAddMediaClick()
```
cadecairos: missing semicolon

src/editor/media-gallery-editor.js
```diff
@@ -65,6 +68,17 @@ define( [ "util/lang", "util/xhr", "util/keys", "util/mediatypes", "editor/edito
     }
   }
 
+  function onDenied( error ) {
+    clearTimeout( _cancelSpinner );
+    clearTimeout( _mediaLoadTimeout );
+    _errorMessage.innerHTML = error;
+    _loadingSpinner.classList.add( "hidden" );
+    _addMediaPanel.classList.add( "invalid-field" );
+    setTimeout( function() {
```
k88hudson, line 78: spacing

src/editor/media-gallery-editor.js
```diff
@@ -174,11 +189,14 @@ define( [ "util/lang", "util/xhr", "util/keys", "util/mediatypes", "editor/edito
       return;
     }
 
-    data.source = val;
+    data.source = url;
     data.type = "sequencer";
-
-    _loadingSpinner.classList.remove( "hidden" );
-    MediaUtils.getMetaData( data.source, onSuccess );
+    _mediaLoadTimeout = setTimeout( function() {
```
k88hudson, line 195: spacing

css/media-editor.less
```diff
@@ -244,20 +244,19 @@
...
       display: none;
     }
   }
 
   .media-loading-spinner {
-    left: 10px;
-    position: relative;
-    top: 3px;
+    float: left;
+    margin-right: 10px;
```
mjschranz, line 258: Why the location move for the loading spinner? I don't know if where it was before was good but I definitely think this spot is poor.

css/ui-states.less
```diff
@@ -61,6 +61,9 @@
   &.hidden {
     display: none;
   }
+  .invalid-field & {
```
mjschranz: Why not just

``` css
&.invalid-field {
.....
}
```

and just apply this to `_loadingSpinner`? Maybe it's just because I don't know enough about css but seeing this here was quite odd looking.

src/editor/media-gallery-editor.js
```diff
@@ -174,11 +189,14 @@ define( [ "util/lang", "util/xhr", "util/keys", "util/mediatypes", "editor/edito
...
+      _errorMessage.innerHTML = "Your media source is taking too long to load";
```
mjschranz: Add a period here at least. I kind of wish you added more here like the old error messages before sequencer but I don't think there's enough room.

src/util/mediatypes.js
```diff
@@ -52,6 +53,12 @@ define( [ "util/xhr", "util/uri" ],
           if ( !respData ) {
             return;
           }
+
+          if ( respData.accessControl.embed === "denied" ) {
+            errorCallback( "Embedding of this Youtube video is disabled" );
```
mjschranz: Period.

src/util/mediatypes.js
```diff
@@ -80,19 +85,16 @@ define( [ "util/xhr", "util/uri" ],
...
-          var denied;
           if ( !respData ) {
             return;
           }
 
           if ( respData.sharing === "private" || respData.embeddable_by === "none" ) {
-            // This informs the plugin that embed is not allowed.
-            // The plugin's action of what to do with this is up to the plugin.
-            denied = true;
+            errorCallback( "Embedding of this Soundcloud media is disabled" );
```
mjschranz: Period.

src/util/mediatypes.js
```diff
@@ -110,7 +112,7 @@ define( [ "util/xhr", "util/uri" ],
           if ( !respData ) {
```
mjschranz, line 55: Maybe vimeo doesn't have a good error code to let us know if embed is disabled, but couldn't we at least add something here making use of the error callback?

src/editor/media-gallery-editor.js
ScottDowne: Turns out in UI land a period is only used if you have more than once sentence. If you have only one, it's not actually a sentence but a block of text. Like a title, and you don't put a period at the end of a title.

src/util/mediatypes.js
ScottDowne, line 55: Vimeo should get its own ticket to use this.

mjschranz, line 55: To use an errorcallback? I really don't understand why we can't do something simple here that simply returns an error message saying Vimeo failed for some unknown reason.

css/ui-states.less
ScottDowne: Yeah, pretty confusing. Basically means .invalid-field is a parent and not child. To be honest this ticket is so low priority now, that I don't know when I'll get back to it, so probably not worth failing it on this.

ScottDowne: Screw it, I'll do it.

css/media-editor.less
ScottDowne, line 258: Visual fields.

It needs to be closer to where the user is looking.

Also needs to be next to the error message, and there isn't enough room for the error message where it was.

What abot it is poor?

PR #1527, opened 2013-03-04
[t1929] sanitization fix

Pomax:
slightly less sanitization, which we can do due to post-DOMContentLoaded buildup.

main analysis and justifications: https://webmademovies.lighthouseapp.com/projects/65733-popcorn-maker/tickets/1929


cornfield/lib/sanitizer.js
```diff
@@ -14,7 +14,20 @@ var sanitizer = {
     }
 
     return value;
+  },
+  reconstituteHTML: function reconstituteHTML( s ) {
+    s = s.replace( /&#34;/g, '"').replace( /&#39;/g, "'");
+    s = s.replace( /&quot;/g, '"').replace( /&apos;/g, "'");
+    s = s.replace( /&ls;/g, '<').replace( /&gt;/g, '>');
+    return s;
```
humphd: Please change this to not store so many temporary variables (it's faster, too):

``` javascript
function reconstituteHTML( s ) {
  return s.replace( /&#34;/g, '"')
          .replace( /&#39;/g, "'")
          .replace( /&quot;/g, '"')
          .replace( /&apos;/g, "'")
          .replace( /&ls;/g, '<')
          .replace( /&gt;/g, '>');
}
```

src/core/project.js
```diff
@@ -182,6 +182,13 @@ define( [ "core/eventmanager", "core/media" ],
       var oldTarget, targets, targetData,
           mediaData, media, m, i, l;
 
+      // convert text with HTML entities into plain text.
+      function reconstituteHTML( htmlString ) {
+        var _ = document.createElement("div");
```
humphd: Let's not use `_` as a variable name here, since it's going to confuse people familiar with underscore.js.  Just call it `div` or something.

src/core/project.js
```diff
@@ -197,7 +204,9 @@ define( [ "core/eventmanager", "core/media" ],
       }
 
       if ( json.name ) {
-        _name = json.name;
+        // undo HTML entities possibly introduced by templating
```
humphd: What does "undo HTML entites" mean?  Can you make this comment more clear?  Also, you have an extra space on the following line between "being" and "applied".

cornfield/routes/index.js
```diff
@@ -191,9 +191,9 @@ module.exports = function routesCtor( app, User, filter, sanitizer, stores, util
         return;
       }
 
-      var projectJSON = JSON.parse( project.data, sanitizer.escapeHTMLinJSON );
-      projectJSON.name = "Remix of " + sanitizer.escapeHTML( project.name );
-      projectJSON.template = sanitizer.escapeHTML( project.template );
+      var projectJSON = JSON.parse( project.data, sanitizer.reconstituteHTMLinJSON );
+      projectJSON.name = "Remix of " + project.name;
```
humphd, line 195: So it's safe to use `project.name` and `project.template` raw like this?  Why was it not safe before but is now?

cornfield/lib/sanitizer.js
Pomax: I can do that, but this is also extremely ugly. Is there a replace(search array, replacement array) that we can make use of? (I know JS doesn't come with that by default, but if we already have that somewhere else we can recycle)

src/core/project.js
Pomax: I never used underscore, I tend to use $ and _ and throwaways, but good point, I will change this.

cornfield/routes/index.js
Pomax, line 195: they technically already were, because of how and when they're unpacked. The project name is introduced well after DOMContentLoaded, so was technically already safe. The template name is even safer, there was no good reason for it to be sanitized at all since we already do a "does the template, by name, exist in the filesystem? if not, throw error". I think it accidentally got caught in a "better make all metadata safe" step.

src/core/project.js
Pomax: HTML entities for ampersand, less-than, etc. I'll adjust the text to be more indicative.

cornfield/lib/sanitizer.js
humphd: We don't, and for the amount we'll do this, I'd just make my suggested change here and move on.  It's not worth adding code for abstraction in this case.

cornfield/routes/index.js
humphd, line 195: OK, this is what I thought, but wanted to check for sure.  Flow diagrams FTW!

cornfield/lib/sanitizer.js
```diff
@@ -14,7 +14,22 @@ var sanitizer = {
...
+    if ( typeof value === "string") {
+      return sanitizer.reconstituteHTML( value );
+    }
+
+    return value;
   }
 };
 
-module.exports = sanitizer;
+module.exports = sanitizer;
```
jbuck: Leave the newline at the end of file

cornfield/lib/sanitizer.js
```diff
@@ -14,6 +14,21 @@ var sanitizer = {
     }
 
     return value;
+  },
+  reconstituteHTML: function reconstituteHTML( s ) {
+    return s.replace( /&#34;/g, '"' )
+            .replace( /&#39;/g, "'" )
+            .replace( /&quot;/g, '"' )
+            .replace( /&apos;/g, "'" )
+            .replace( /&ls;/g, '<' )
```
secretrobotron: This looks wrong. The encoding for a left bracket is `&lt`, no?

Pomax: it totally is. I shall befixolate.

cornfield/lib/sanitizer.js
```diff
@@ -14,6 +14,21 @@ var sanitizer = {
...
 
     return value;
+  },
+  reconstituteHTML: function reconstituteHTML( s ) {
+    return s.replace( /&#34;/g, '"' )
+            .replace( /&#39;/g, "'" )
+            .replace( /&quot;/g, '"' )
+            .replace( /&apos;/g, "'" )
+            .replace( /&lt;/g, '<' )
+            .replace( /&gt;/g, '>' );
```
jbuck: are we missing replacing &amp; ?

src/core/project.js
```diff
@@ -182,6 +182,20 @@ define( [ "core/eventmanager", "core/media" ],
       var oldTarget, targets, targetData,
           mediaData, media, m, i, l;
 
+      // convert text with HTML entities into plain text.
+      function reconstituteHTML( htmlString ) {
+        var unpackDiv = document.implementation.createHTMLDocument("").createElement("div");
+        unpackDiv.innerHTML = htmlString;
+        // strip any non-text DOM nodes that this introduces.
+        var children = unpackDiv.childNodes;
+        for( var i = children.length - 1; i >= 0; i-- ) {
```
jbuck: for isn't a function

src/util/sanitizer.js
```diff
@@ -0,0 +1,31 @@
...
+      unpackDiv = offscreenDocument.createElement("div");
+
+  var Sanitizer = {
+      reconstituteHTML: function reconstituteHTML( htmlString ) {
+        // assign as HTML, resolving any HTML entities present.
+        unpackDiv.innerHTML = htmlString;
+        // strip any non-text DOM nodes that this introduces.
+        var children = unpackDiv.childNodes, i;
+        for ( i = children.length - 1; i >= 0; i-- ) {
+          if ( children[ i ].nodeType !== 3 ) {
```
secretrobotron: To avoid some potential confusion, perhaps `Node.TEXT_NODE` is a bit better than `3`.

jbuck:
Landed as https://github.com/mozilla/butter/commit/5b870d9495e6df28b089c978b8e41587b7f03531

PR #1531, opened 2013-03-04
[2646] We need an internal facing index of created projects

humphd:


routes/index.js
```diff
@@ -23,16 +24,49 @@ module.exports = function routesCtor( app, User, filter, sanitizer, stores, util
...
+    function( req, res ) {
+      Project.findRecentlyUpdated( { limit: req.params.limit }, function( err, projects ) {
+        if ( err ) {
+          res.json( { error: err }, 500 );
+        }
+        res.json( { error: 'okay', results: projects } );
+      });
+  });
+
+  app.get( '/api/projects/remixes/:id?',
```
jbuck: /ap/project/remixes, since you're only looking at a single users remixes.

humphd:
Still waiting on Travis for this...

routes/index.js
jbuck: Actually, better yet, `/api/project/:id/remixes` right? You'll probably need to move this definition above `/api/project/:id` though

server.js
```diff
@@ -119,12 +119,17 @@ app.post( '/api/publish/:id',
   var email = req.session.email,
       id = parseInt( req.params.id, 10 );
 
+  if ( !email ) {
```
jbuck: You don't need to add this code; the `filter.isLoggedIn` middleware for this route will prevent the execution of this code.

routes/index.js
```diff
@@ -23,16 +24,65 @@ module.exports = function routesCtor( app, User, filter, sanitizer, stores, util
...
+        author: result.author,
+        remixedFrom: result.remixedFrom,
+        createdAt: result.createdAt,
+        updatedAt: result.updatedAt
+      });
+    });
+    return pruned;
+  }
+
+  app.get( '/api/projects/recentlyCreated/:limit?',
```
jbuck, line 48: `/api/project/:limit/recentlyCreated`

jbuck, line 48: Hmm, actually....

PR #1535, opened 2013-03-05
[t3275] Use Media Icon's for media gallery clip information

mjschranz:


src/util/mediatypes.js
```diff
@@ -18,23 +18,30 @@ define( [ "util/xhr", "util/uri" ],
       for ( var type in REGEX_MAP ) {
         if ( REGEX_MAP.hasOwnProperty( type ) ) {
           if ( REGEX_MAP[ type ].test( url ) ) {
-            return type;
+            return {
+              pretty: type,
+              normal: type.toLowerCase()
```
k88hudson: While this is not a problem now, I can see this being an issue if we ever have a media type that contains characters that are not valid in CSS, such as a space or period ( e.g. Archive.org ). Also, if you are automating this conversion, is there any advantage to storing it? Why not just run `type.toLowerCase()` whenever you need it?

PR #1537, opened 2013-03-05
[t3253] Don't trigger media loading on empty url input

mjschranz:


src/editor/media-gallery-editor.js
```diff
@@ -168,7 +168,13 @@ define( [ "util/lang", "util/xhr", "util/keys", "util/mediatypes", "editor/edito
...
-    var data = {};
+    var data = {},
+        val = _urlInput.value;
+
+    // Don't trigger with empty inputs
+    if ( !val ) {
+      return;
+    }
 
     data.source = _urlInput.value;
```
cadecairos: If you're putting `_urlInput.value` into `val`, you might as well reference it here too, or not store it at all.

PR #1541, opened 2013-03-05
[t3287] Fixed wonkiness on updateEndAfterExpand

k88hudson:


src/editor/sequencer-editor.js
```diff
@@ -225,8 +225,9 @@ define( [ "util/mediatypes", "editor/editor", "util/time",
           track.addTrackEvent( _trackEvent );
           _this.updateTrackEventSafe( _trackEvent, updateOptions );
           _butter.unlisten( "mediaready", updateEndAfterExpand );
+          _butter.currentTime = _popcornOptions.start;
+          updateUI( updateOptions );
           updateOptions = {};
-          updateUI();
```
cadecairos, line 229: previously, update options was reset before updateUI was called, and upateoptions wasn't passed as a param. Is there a reason it's being reset/passed in?

k88hudson, line 229: I pass in updateOptions so that there will never be a case where the trackEvent hasn't updated yet with updateUI is run

PR #1548, opened 2013-03-05
[t3270] Have slider container respond to clicks. Add text to report curr...

mjschranz:
...ent value

### DIFF
--- a/src/editor/sequencer-editor.js
+++ b/src/editor/sequencer-editor.js
@@ -99,7 +99,12 @@ define( [ "util/mediatypes", "editor/editor", "util/time",
 
       function Slider( el, property ) {
         var scrubber = el.querySelector( ".butter-slider-scrubber" ),
-            MAX_VAL = el.getAttribute( "data-max" ) || 100;
+            MAX_VAL = 100,
+            SCRUBBER_OFFSET = 4,
+            SCRUBBER_TOOLTIP_OFFSET = 3,
+            firstX, startLeft,
+            sliderToolTip = _rootElement.querySelector( ".butter-slider-tooltip" ),
+            rects = el.getBoundingClientRect();
 
         function toReal( n ) {
           var maxUI = el.offsetWidth;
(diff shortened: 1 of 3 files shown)


src/editor/sequencer-editor.js
```diff
@@ -146,13 +148,25 @@ define( [ "util/mediatypes", "editor/editor", "util/time",
...
         function onSlideStart( e ) {
           e.preventDefault();
+          el.removeEventListener( "click", onClick, false );
           firstX = e.clientX;
           startLeft = scrubber.offsetLeft;
           document.addEventListener( "mousemove", onSliding, false );
           document.addEventListener( "mouseup", onSlideStop, false );
         }
 
+        function onClick( e ) {console.log("d")
```
k88hudson: console.log

src/editor/sequencer-editor.js
```diff
@@ -146,13 +148,25 @@ define( [ "util/mediatypes", "editor/editor", "util/time",
...
+          e.preventDefault();
+          var rects = el.getBoundingClientRect(),
+              percentage = ( e.clientX - rects.left ) / rects.width;
```
k88hudson: Do you think this is better than mapping using a function? Is that over-engineered? If so we should probably change the other stuff that uses `.toReal` and `.toUI` to do the same thing as this

src/editor/sequencer-editor.js
```diff
@@ -121,18 +122,19 @@ define( [ "util/mediatypes", "editor/editor", "util/time",
         }
 
         function updateUI() {
+          var sliderVal = _rootElement.querySelector( ".slider-value" );
```
k88hudson: You'll want to make this a tooltip

src/editor/sequencer-editor.js
mjschranz: I most definitely think this is the case. It's only being used in one spot so why add function calls to the stack?

PR #1550, opened 2013-03-06
Bug #3305 - Have embed.js load Popcorn via require

jbuck:
This pull request does a bunch of cool things. The primary goal was to have RequireJS load Popcorn, and as a side-effect we get:

1) Nuke the whole concept of external assets loaded by the server being different from the JS, eliminating a whole class of "which file do I edit to get my new plugin loaded"
2) IE9 shims will _always_ load before app code
3) We only load a single JS file now, so faster loading times
4) The r.js optimizer will concat all of the Popcorn files into `embed.js` so that we'll eventually be able to remove all of our build-related code for that

:yum: 

This is only the first part; I'll do a similar thing for butter.js in another patch.

### DIFF
--- a/src/embed.js
+++ b/src/embed.js
@@ -1,4 +1,8 @@
-function init( window, document ) {
+/*! This Source Code Form is subject to the terms of the MIT license
+ *  If a copy of the MIT license was not distributed with this file, you can
+ *  obtain one at https://raw.github.com/mozilla/butter/master/LICENSE */
+
+function init() {
 
   var stateClasses = [
     "embed-playing",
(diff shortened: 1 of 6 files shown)


cornfield/views/embed.jade
```diff
@@ -2,6 +2,7 @@
 html(lang='en')
   head
     base(href='#{baseHref}')
+    script(src='/src/embed.js')
```
humphd: Why move this?

src/embed.js
```diff
@@ -242,21 +246,105 @@ function init( window, document ) {
...
+      "popcorn.loopPlugin": "../templates/assets/plugins/loopPlugin/popcorn.loopPlugin",
+      "popcorn.pausePlugin": "../templates/assets/plugins/pausePlugin/popcorn.pausePlugin",
+      "popcorn.popup": "../templates/assets/plugins/popup/popcorn.popup",
+      "popcorn.sequencer": "../templates/assets/plugins/sequencer/popcorn.sequencer",
+      "popcorn.skip": "../templates/assets/plugins/skip/popcorn.skip",
+      "popcorn.text": "../templates/assets/plugins/text/popcorn.text",
+      "popcorn.twitter": "../templates/assets/plugins/twitter/popcorn.twitter",
+      "popcorn.wikipedia": "../templates/assets/plugins/wikipedia/popcorn.wikipedia",
+
+      // Wrappers
```
humphd: I'd move these after the ie8 shim above, since they are needed before players can ever be used.  It just helps to understand the deps a bit better--I realize the shim section below codifies it, but for reading purposes.

src/embed.js
```diff
@@ -242,21 +246,105 @@ function init( window, document ) {
...
+      "popcorn.popup": [ "popcorn"],
+      "popcorn.sequencer": [ "popcorn", "popcorn.player" ],
+      "popcorn.skip": [ "popcorn" ],
+      "popcorn.text": [ "popcorn" ],
+      "popcorn.twitter": [ "popcorn" ],
+      "popcorn.wikipedia": [ "popcorn" ],
+
+      // Wrappers
+      "popcorn._MediaElementProto": [ "popcorn" ],
+      "popcorn.HTMLMediaElement": [ "popcorn" ],
```
humphd: I think you should add `popcorn._MediaElementProto` here too, in case we do things with it down the road.  It won't hurt.

src/embed.js
jbuck: In the paths/shim config block, order doesn't matter; I just kept it the same as the shim config block. I'm happy to change the ordering of both, but I'd like both to match each other exactly because it'll be easier to add/remove files that way

cornfield/views/embed.jade
jbuck: Hm, I think I moved it originally as part of a different version of the patch, but it can probably go back to it's original location without harming anyone

src/embed.js
```diff
@@ -408,15 +496,22 @@ function init( window, document ) {
...
+    rscript.onload = function() {
+      init();
+    };
+    rscript.src = "/external/require/require.js";
+    document.head.appendChild(rscript);
   } else {
-    init( window, window.document );
+    init();
   }
-}, false );
```
humphd: Undo this change.

src/embed.js
```diff
@@ -408,15 +496,22 @@ function init( window, document ) {
...
+  });
 }
 
 document.addEventListener( "DOMContentLoaded", function() {
   // Source tree case vs. require-built case.
   if ( typeof require === "undefined" ) {
-    Popcorn.getScript( "../../external/require/require.js", function() {
-      init( window, window.document );
-    });
+    var rscript = document.createElement("script");
```
humphd: space around "script"

src/embed.js
```diff
@@ -408,15 +496,22 @@ function init( window, document ) {
...
+    rscript.onload = function() {
+      init();
+    };
+    rscript.src = "/external/require/require.js";
+    document.head.appendChild(rscript);
```
humphd: space around `rscript`

src/embed.js
jbuck: HTMLMediaElement uses the built-in audio/video elements, and doesn't depend on our MediaElementProto at all; if this file doesn't depend on it, why should it be listed as a dependency?

tools/build-embed.js
```diff
@@ -7,22 +7,19 @@
...
-  },
-
   // Target the AMD loader shim as the main module to optimize,
   // so it shows up first in the built file,
   // since the embed modules use the define/require APIs that the almond
   // provides. Path is relative to baseUrl.
   name: '../tools/almond',
 
-  // Strip extra licenses
-  preserveLicenseComments: false,
```
humphd: We've ended up with a mess of license block comments in the final `embed.js`.  Why remove this?

src/embed.js
humphd: You're right.

cornfield/views/embed.jade
humphd: Yeah, please do.

src/embed.js
humphd: Yeah, let's do that.

Pomax: why is this a hardcoded list, rather than an fs-grep for all the plugins, or a load-from-conf? The location is uniform, it's all in ../templates/assets/plugins

cornfield/views/embed.jade
```diff
@@ -2,6 +2,7 @@
...
     !{templateScripts}
     !{externalAssets}
```
Pomax: if we're doing requires for templates and assets in embed.js now, do we still need these two lines?

src/embed.js
```diff
@@ -1,4 +1,8 @@
-function init( window, document ) {
+/*! This Source Code Form is subject to the terms of the MIT license
+ *  If a copy of the MIT license was not distributed with this file, you can
```
Pomax, line 2: I know what "Source Code" means, but what's "Source Code Form"? Is just the indentation licensed and the actual code text isn't? O_o (can we change this to something that doesn't make people wonder what we're saying?)

cornfield/views/embed.jade
jbuck: Sadly yes; ripping this out will be another bug because it means that we can't have multiple-template publishing. Which I'd actually have no problem with removing! Just that it has large repercussions.

src/embed.js
jbuck, line 2: Feel free to file a bug on it! This is just a license copied from all of our other files

src/embed.js
jbuck: This code only gets executed in the browser, so no file system access. The locations are not uniform, we've got stuff in `templates/assets/` and various directories in `external/popcorn-js/`

PR #1552, opened 2013-03-06
[t3235] Prevent zero duration events from being created inside start/end...

mjschranz:
... fields in editor

### DIFF
--- a/public/src/core/views/trackevent-view.js
+++ b/public/src/core/views/trackevent-view.js
@@ -42,7 +42,17 @@ define( [ "core/logger", "core/eventmanager", "util/dragndrop",
       }
       _element.style.left = _start  / _trackEvent.track._media.duration * 100 + "%";
       _element.style.width = ( _end - _start ) / _trackEvent.track._media.duration * 100 + "%";
+
+      _this.setResizeArrows();
     }
+
+    this.setResizeArrows = function() {
+      if ( _element.offsetWidth < TRACKEVENT_MIN_WIDTH ) {
+        _element.classList.add( "trackevent-small" );
+      } else {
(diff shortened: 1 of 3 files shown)


src/core/trackevent.js
```diff
@@ -148,6 +148,15 @@ define( [ "./logger", "./eventmanager", "./observer",
         media = _track._media;
         duration = media.duration;
 
+        // Prevent no duration events causing unclickable/selectable trackevents
```
Pomax: better to call these "zero-duration" events, since "no" also implies there is no event at all

public/src/core/trackevent.js
```diff
@@ -222,8 +222,8 @@ define( [ "./logger", "./eventmanager", "./observer",
         }
       }
 
-      _popcornOptions.start = newStart;
-      _popcornOptions.end = newEnd;
+      _popcornOptions.start = updateOptions.start = newStart;
```
k88hudson: Remove

public/src/core/views/trackevent-view.js
```diff
@@ -42,8 +42,18 @@ define( [ "core/logger", "core/eventmanager", "util/dragndrop",
       }
       _element.style.left = _start  / _trackEvent.track._media.duration * 100 + "%";
       _element.style.width = ( _end - _start ) / _trackEvent.track._media.duration * 100 + "%";
+
+      _this.setResizeArrows();
     }
 
+    this.setResizeArrows = function() {
```
ScottDowne, line 49: Near the bottom of this file, we have a block of code that looks almost like your new function.

Should be able to use your new function with it.

Line 269 of this file, is where it starts.

PR #1556, opened 2013-03-07
Requesting a pull to mozilla:master from k88hudson:t3323

k88hudson:


css/editors.less
```diff
@@ -726,6 +726,7 @@ textarea {
   &.small {
     background: none;
     border: none;
+    box-shadow: none;
```
k88hudson, line 729: This was causing the lefthand side to look flat instead of rounded

src/editor/sequencer-editor.js
```diff
@@ -202,16 +202,10 @@ define( [ "util/mediatypes", "editor/editor", "util/time",
...
-            clipSection.style.left = timeToPosition( from ) + "px";
-            inInput.value = Time.toTimecode( from );
-            outInput.value = Time.toTimecode( from + end - start );
-          }
-
-          if ( options.end || options.start ) {
-            clipSection.style.width = timeToPosition( end - start ) + "px";
-            outInput.value = Time.toTimecode( from + end - start );
-          }
+          clipSection.style.left = timeToPosition( from ) + "px";
```
k88hudson, line 205: The problem was not allowing 0, but honestly it's not that bad to just always update everything

PR #1560, opened 2013-03-07
Initial crack at rewriting when mediaready is firing

mjschranz:


src/editor/sequencer-editor.js
```diff
@@ -221,10 +221,10 @@ define( [ "util/mediatypes", "editor/editor", "util/time",
 
         function updateEndAfterExpand() {
           var track = _trackEvent.track;
-          track.removeTrackEvent( _trackEvent );
+          track.removeTrackEvent( _trackEvent, true );
```
mjschranz, line 224: This was me fixing an oversight from initial review of the editor.

src/core/track.js
```diff
@@ -236,7 +236,7 @@ define( [ "./eventmanager", "./trackevent", "./views/track-view", "util/sanitize
       }
 
       // Sanitize the track even data prior to building the UI.
-      this.sanitizeTrackEventData(trackEvent);
+      this.sanitizeTrackEventData( trackEvent );
```
mjschranz, line 239: This entire file was whitespace fixes that weren't caught in review of this patch.

src/core/media.js
```diff
@@ -106,6 +115,43 @@
       this.popcornScripts = null;
       this.maxPluginZIndex = 0;
 
+      function onSequencerReady( e ) {
+        var type = e.data.type;
+
+        if ( type === "sequencer" ) {
+          if ( _projectImport ) {
+            _mediaClipsLoaded += 1;
```
ScottDowne: _mediaClipsLoaded++;

src/core/media.js
```diff
@@ -106,6 +115,43 @@
...
+            _projectImport = false;
+          }
+
+          if ( !_projectImport ) {
+            _this.dispatch( "mediaready" );
+          }
+        }
+      }
+
+      function onSequencerAdded( e ) {console.log("a");
```
ScottDowne: Trying to kill the name sequencer. Use mediaClip

src/core/media.js
```diff
@@ -106,6 +115,43 @@
...
+        }
+      }
+
+      function onSequencerRemoved( e ) {
+        if ( e.data.type === "sequencer" ) {
+          _mediaClips.splice( _mediaClips.indexOf( e.data ), 1 );
+        }
+      }
+
+      _this.listen( "trackeventadded", onSequencerAdded );
```
ScottDowne, line 146: Would be nice if you didn't have to do this for all track events.

I am sure we already have events for these, I also think we already have a spot where we do a check for type as it is being added to make choices like this.

src/core/media.js
```diff
@@ -37,6 +37,9 @@
           _mediaUpdateInterval,
           _clipData = {},
           _this = this,
+          _mediaClipsLoaded = 0,
+          _mediaClips = [],
```
ScottDowne: Would be nice if we stored an array of media clips like we do for selected, added, sorted track events. That anyone could look at. They could be sorted. Also consider something for loadedMediaTrackEvents, and even an event for when a media track event is loaded.

PR #1561, opened 2013-03-07
[#3327] Start refactor of trackEvent.update by doing default applying in...

ScottDowne:
... another function.


src/core/trackevent.js
```diff
@@ -62,7 +62,7 @@ define( [ "./logger", "./eventmanager", "./observer",
     _this.popcornTrackEvent = null;
 
     function defaultValue( item ) {
-      if ( item.default ) {
+      if ( item.hasOwnProperty( "default" ) ) {
```
cadecairos, line 65: This isn't really necessary. Yes, `default` is a reserved word but it's safe to reference as a property of an object.

PR #1566, opened 2013-03-08
Have butter.js load Popcorn via require

jbuck:


tools/build-butter.js
```diff
@@ -31,12 +21,15 @@
   // have explicit dependencies for them, but uses them on demand. Also,
   // butter.js references butter-src in a document.write string, so it will
   // not be found by the AST analysis done in the optimizer.
-  include: [
-            '../external/ua-parser/ua-parser',
-            '../external/jsSHA/sha1',
-            'util/shims',
-            'main'
-           ],
+  include: [ 'butter', 'butter-main' ],
```
secretrobotron: I don't think you need `butter-main` here. Looks redundant, and builds/runs successfully without it on my system. Double check?

jbuck: Indeed! Because I enabled `findNestedDependencies`, I do not need to add `butter-main`.

src/main.js
```diff
@@ -25,7 +46,16 @@
     "RIM Tablet OS"
   ];
 
-  define( [
+  var require = requirejs.config({
+    baseUrl: "/src",
+    paths: {
+      // RequireJS
+      "text": "../external/require/text"
```
secretrobotron: The `text` plguin is defined a couple places: here and https://github.com/mozilla/butter/pull/1566/files#L5R35.

Can we cut one out? `text` isn't a `popcorn` dependency, so it'd be nice to pull it out of `popcorn.js`. However, it'd be nice to drop this extra `require.config` call altogether if possible.

src/main.js
```diff
@@ -1019,4 +1049,11 @@
     return Butter;
   });
 
+  // butter depends on popcorn, so don't change this unless you know what you're doing
+  require([ "util/shims", "../external/jsSHA/sha1" ], function() {
```
secretrobotron: Just a thought about this nesting: `util/shims` only resembles an AMD module inasmuch as it calls `define`, but returns nothing. If it's plain JS like Popcorn files are, can it be moved into `popcorn.js` in the `shim` declaration?

src/main.js
jbuck: So, I can't remove the `text` dependency from `src/popcorn.js` because the r.js optimizer uses `src/popcorn.js` as the config file. It does look safe to remove the path config for `text` from  `src/butter.js`, but `src/embed.js` depends on `ui/widget/controls` which depends on `text!layouts/controls.html` so I don't think we can remove it from that. Unless we also make `src/embed.js` a 3 level require() call as well.

src/main.js
```diff
@@ -2,7 +2,28 @@
...
+    }
+    window.Butter.__waiting = arguments;
+  }
+};
+
+(function init() {
+
+  // If we need to load requirejs before loading butter, make it so
+  if ( typeof define === "undefined" ) {
+    var rscript = document.createElement("script");
```
secretrobotron: Might want to fix your spacing inside this if block.

src/main.js
jbuck: If we make it non-AMD code, yes. But since both the editor and embed can use it as an AMD module (and indeed, require it to work in IE9) I think we should leave it as an AMD module.

src/main.js
jbuck: Also, the requirejs.config blocks need to stay to set `baseUrl` at the very minimum.

PR #1572, opened 2013-03-12
[t3364] Prevent drops of trackevents when ghosting occured appearing on ...

mjschranz:
...wrong tracks


public/src/core/views/trackevent-view.js
```diff
@@ -40,6 +40,7 @@ define( [ "core/logger", "core/eventmanager", "util/dragndrop",
       if ( _trackEvent.track.view.element !== _element.parentNode ) {
         _trackEvent.track.view.element.appendChild( _element );
       }
+
```
cadecairos: remove this line change.

PR #1573, opened 2013-03-12
Bug #3346 - Delete duplicate Popcorn loading code

jbuck:


src/butter.js
```diff
@@ -709,53 +703,51 @@ window.Butter = {
...
-        _page.prepare(function() {
-          if ( !!_config.value( "scrapePage" ) ) {
-            var i, j, il, jl, url, oldTarget, oldMedia, mediaPopcornOptions, mediaObj;
-            for( i = 0, il = targets.length; i < il; ++i ) {
-              // Only add targets that don't already exist.
-              oldTarget = _this.getTargetByType( "elementID", targets[ i ].element );
-              if( !oldTarget ){
-                _this.addTarget({ element: targets[ i ].id });
-              }
+        if ( !!_config.value( "scrapePage" ) ) {
```
mjschranz: I honestly wonder if we want to bother doing any of this page scraping anymore. I guess the question to this is what direction we plan on going with custom template authors. This code also isn't really doing much harm so I think we are fine leaving it in.

mjschranz: I mean specifically from a Media perspective.

jbuck: It causes harm by existing! Bobby wasn't around to ask though, so I just left it in. It'd be nice to delete it all though!

src/plugin/module.js
```diff
@@ -92,16 +69,14 @@ define( [ "core/logger", "./plugin-list", "./plugin" ],
...
         for ( i = 0, l = newPlugins.length; i < l; i++ ) {
           plugin = newPlugins[ i ];
           _plugins.push( plugin );
           butter.dispatch( "pluginadded", plugin );
         }
         onReadyCallback();
-      }, function() {
-        console.warn( "Failed to load all plugins. Please check logs and file paths." );
-      });
+      }, 4);
```
mjschranz: Space.

src/core/page.js
```diff
@@ -20,84 +16,5 @@ define( [ "core/logger", "core/eventmanager" ], function( Logger, EventManager )
...
-      loader.load({
-        type: "js",
-        url: PLAYER_TYPE_URL.replace( /\{type\}/g, type ),
-        check: function(){
-          return !!Popcorn[ type ];
-        }
-      }, callback );
-    };
-
   }; // page
```
Pomax: What's left of this code looks like it's a few levels of indirection too many. Can we move this into where it ends up actually being used? (Does it still need to be an independent object-function-object-function-function?)

src/plugin/module.js
```diff
@@ -92,16 +69,14 @@ define( [ "core/logger", "./plugin-list", "./plugin" ],
...
         for ( i = 0, l = newPlugins.length; i < l; i++ ) {
           plugin = newPlugins[ i ];
           _plugins.push( plugin );
           butter.dispatch( "pluginadded", plugin );
         }
         onReadyCallback();
-      }, function() {
-        console.warn( "Failed to load all plugins. Please check logs and file paths." );
-      });
+      }, 4 );
```
Pomax, line 79: ... what? "number determined by rolling a fair die"? which is this on a timeout? (and why is that timeout 4ms?)

jbuck, line 79: It's on a timeout because the code path here expects the onReadyCallback to be fired sometime after it returns newPlugins.

The timeout is 4ms because that's the lower bound for scheduling a setTimeout call across browsers.

src/core/page.js
jbuck:
```
Jons-MacBook-Pro:butter jon$ grep -R '\.scrape' *
src/core/page.js:    this.scrape = function() {
src/butter.js:        var scrapedObject = _page.scrape(),
test/page/page-scrape.html:          scrapedPage = _page.scrape();
```

It's defined in `core/page.js`, used in `src/butter.js`, and tested in `test/page/page-scrape.html`. Maybe just make it a static class, rather than one that can be instantiated? Or should it just be moved in `src/butter.js`?

Pomax: if it needs its own test, making it a static util function makes the most sense to me. If it doesn't, anonymising it into butter.js is :thumbsup:

PR #1581, opened 2013-03-15
T3087

mjschranz:


public/css/super-scrollbar.less
```diff
@@ -15,7 +16,7 @@
       position: absolute;
       height: 16px;
       width: 6px;
-      left: 50%;
+      right: 50%;
```
k88hudson: You should rebase on master, it will have to be `100%` now.

PR #1584, opened 2013-03-15
Bug #3319 - Switch to nconf for configuration

jbuck:


lib/config.js
```diff
@@ -0,0 +1,17 @@
+var nconf = require( "nconf" );
+
+// 1) Load default settings from our config.
```
humphd, line 3: I _love_ seeing docs like this on code.  +1

server.js
```diff
@@ -3,29 +3,21 @@ if ( process.env.NEW_RELIC_HOME ) {
...
-}
-
 var express = require('express'),
     fs = require('fs'),
     path = require('path'),
     jade = require('jade'),
     app = express(),
     lessMiddleware = require('less-middleware'),
-    CONFIG = require('config'),
+    CONFIG = require( './lib/config' ),
```
humphd: This doesn't need to be all caps anymore, lets use `config`.

PR #1585, opened 2013-03-15
[t3209] Refactor generateSafeTrackEvent to accept a popcorn options obje...

mjschranz:
...ct vs only start/end

### DIFF
--- a/public/src/butter.js
+++ b/public/src/butter.js
@@ -180,10 +180,11 @@ window.Butter = {
         return _currentMedia.getManifest( name );
       }; //getManifest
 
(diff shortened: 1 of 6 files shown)


public/src/butter.js
```diff
@@ -244,12 +245,13 @@ window.Butter = {
 
       function targetTrackEventRequested( e ) {
         var trackEvent,
-            popcornOptions,
+            popcornOptions = {},
             start = _currentMedia.currentTime,
             end;
```
cadecairos: This variable is never used ( lint fails )

PR #1594, opened 2013-03-16
Bug #3392 - Add Heroku support

jbuck:


lib/default-config.js
```diff
@@ -1,8 +1,15 @@
-{
-  "server" : {
-    "bindIP" : "localhost",
-    "bindPort" : "8888"
-  },
+// This is a JS file so that we can add inline comments. Yep.
```
humphd: Remove this.  Perhaps add a single block comment at the top indicating that this is the main config file, etc.

PR #1598, opened 2013-03-18
Added GA setDomain method to track users across subdomains.

kylerush:
This change allows us to see aggregate analytics (navigation summaries, etc.) for users that first land on webmaker.org and then navigate to popcorn.webmaker.org.

More info here: https://developers.google.com/analytics/devguides/collection/gajs/gaTrackingSite


jbuck:
Merged https://github.com/mozilla/butter/commit/5d391caa8f2ed636b56365294c3bb7351df36760

PR #1599, opened 2013-03-19
Bug #3406 - use mediatypes.js to get media info if saved JSON has an inv...

cadecairos:
...alid duration

https://webmademovies.lighthouseapp.com/projects/65733/tickets/3406

### DIFF
--- a/public/src/core/media.js
+++ b/public/src/core/media.js
@@ -582,7 +582,7 @@
                 fallbacks = [],
                 sources = [];
 
-            function importTracks() {
+            function doImportTracks() {
               if ( importData.tracks ) {
                 var importTracks = importData.tracks;
                 if( Array.isArray( importTracks ) ) {
@@ -641,12 +641,12 @@
             if ( importData.duration >= 0 ) {
               _duration = importData.duration;
               _this.url = "#t=," + _duration;
-              importTracks();
+              doImportTracks();
             } else {
               MediaTypes.getMetaData( importData.url[ 0 ], function success( data ) {
                 _duration = data.duration;
                 _this.url = "#t=," + _duration;
-                importTracks();
+                doImportTracks();
               });
             }
             if ( importData.clipData ) {
--- a/public/src/core/media.js
+++ b/public/src/core/media.js
@@ -8,9 +8,10 @@
             "core/eventmanager",
             "core/track",
             "core/popcorn-wrapper",
-            "util/uri"
+            "util/uri",
+            "util/mediatypes"
           ],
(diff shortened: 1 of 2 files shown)


public/src/core/media.js
```diff
@@ -589,6 +641,13 @@
             if ( importData.duration >= 0 ) {
               _duration = importData.duration;
               _this.url = "#t=," + _duration;
+              doImportTracks();
+            } else {
+              MediaTypes.getMetaData( importData.url[ 0 ], function success( data ) {
```
mjschranz: What is importData.url isn't an array at this point? We have code at that checks if it is or not. We might need to move this down here to happen before.

ScottDowne: I think Matt is right here.

I think we can do better to move this check lower down.

public/src/core/media.js
```diff
@@ -580,15 +581,76 @@
...
+                    _this.addTrack( newTrack );
+                    newTrack.updateTrackEvents();
+                  }
+                  // Backwards comp for old base media.
+                  // Insert previous base media as a sequence event as the last track.
+                  if ( importData.url && _duration >= 0 ) {
+                    var firstSource;
+
+                    // If sources is a single array and of type null player, don't bother making a sequence.
+                    if ( url.length > 1 || !( /#t=\d*,?\d+?/ ).test( url[ 0 ] ) ) {
```
ScottDowne, line 601: Make this use mediaTypes.checkUrl

public/src/util/mediatypes.js
```diff
@@ -10,7 +10,10 @@ define( [ "util/xhr", "util/uri" ],
   var REGEX_MAP = {
         youtube: /(?:https?:\/\/www\.|https?:\/\/|www\.|\.|^)youtu/,
         vimeo: /https?:\/\/(www\.)?vimeo.com\/(\d+)($|\/)/,
-        soundcloud: /(?:https?:\/\/www\.|https?:\/\/|www\.|\.|^)(soundcloud)/
+        soundcloud: /(?:https?:\/\/www\.|https?:\/\/|www\.|\.|^)(soundcloud)/,
+        // supports #t=<start>,<duration>
+        // where start or duration can be: X, X.X or XX:XX
+        "null": /^\s*#t=(?:\d*(?:(?:\.|\:)?\d+)?),?(\d+(?:(?:\.|\:)\d+)?)\s*$/
```
ScottDowne, line 16: This is an unfortunate regex. Not sure if there is a similar way, but if there is, we should update it.

PR #1600, opened 2013-03-19
[#3405] Make sequencer work in a super simple example, plus example.

ScottDowne:


public/templates/assets/plugins/sequencer/popcorn.sequencer.html
```diff
@@ -0,0 +1,83 @@
...
+      });
+
+      var require = requirejs.config({
+        baseUrl: "../../../../src",
+        paths: {
+          "text": "../external/require/text"
+        }
+      });
+
+      define( "sequencer-example", [ "ui/widget/controls" ], function( Controls ) {
```
jbuck, line 66: You can change this to `require([ "ui/widget/controls" ], function( Controls ) {` and remove the require statement below

PR #1601, opened 2013-03-19
[3409] Expose proxied statsd metrics collection end-point to Butter

humphd:


routes/index.js
```diff
@@ -261,4 +261,44 @@ module.exports = function routesCtor( app, User, filter, sanitizer, stores, util
...
+   *   method: a statsd metric API call to use. Possible values are
+   *           timing, increment, decrement, gauge, set, and unique.
+   *
+   *   name: the name of the stat (e.g., project.save)
+   *
+   *   value: the value of the stat (e.g., 1)
+   *
+   * NOTE: we don't bother with sampleRate, but that could be added.
+   */
+  app.post( '/metric', function( req, res ) {
```
jbuck: Point it at /api/metric

jbuck:
Looks good, but I want a whitelist on the server side so that a malicious user can't send counters for anything.

lib/default-config.js
```diff
@@ -73,5 +73,11 @@ module.exports = {
         "collate": "utf8_general_ci"
       }
     }
+  },
+  "metrics": {
```
jbuck: Need to nuke this block too before it lands

public/src/core/metrics.js
```diff
@@ -0,0 +1,74 @@
...
+
+    methods.forEach( function( method ) {
+      o[ method ] = function( name, value ) {
+        metrics[ method ]( name, value );
+      };
+    });
+  }
+
+  return function( butter ) {
+    var _metrics = butter.cornfield.metrics,
```
jbuck: It'd be nice to just `require("core/metrics")` instead of using the butter global, but if it's not possible because the cornfield module needs more restructuring that's fine.

lib/default-config.js
mjschranz: Is this comment still relevant? It looks like it is needed (although it might be missing a bunch of things that can/should be added to the whitelist block).

public/src/core/popcorn-wrapper.js
```diff
@@ -151,11 +153,13 @@ define( [ "core/logger", "core/eventmanager", "util/uri" ], function( Logger, Ev
       // called when timeout occurs preparing popcorn
       function popcornTimeoutWrapper( e ) {
         _interruptLoad = true;
+        metrics.increment( "timeout.popcorn" );
         _onTimeout( e );
       }
 
       // called when timeout occurs preparing media
       function mediaTimeoutWrapper( e ) {
+        metrics.increment( "timeout.media" );
```
mjschranz: This seems like the wrong place to increase this metric now because of the sequencer. Since we are always constructing Popcorn instances with null video I'm fairly certain in saying this will never get hit.

https://github.com/mozilla/butter/blob/master/public/src/editor/media-gallery-editor.js#L217-L221 seems like one possible candidate for where it might go. This will track whenever our MediaTypes util isn't able to parse the URL properly.

Another might be inside https://github.com/mozilla/butter/blob/master/public/src/util/mediatypes.js itself to keep track of whenever error callbacks are hit.

Unfortunately the code in Popcorn Wrapper is sort of a lie. It will only really keep track if it times out trying to load the players we use otherwise the timeouts aren't hit.

PR #1602, opened 2013-03-20
[t3408] Don't allow negative or 0 to be entered for project duration in ...

mjschranz:
...media editor

### DIFF
--- a/public/src/editor/media-gallery-editor.js
+++ b/public/src/editor/media-gallery-editor.js
@@ -49,25 +49,27 @@ define( [ "util/lang", "util/xhr", "util/keys", "util/mediatypes", "editor/edito
   }
 
   function setBaseDuration( duration ) {
-    if ( duration === "" ) {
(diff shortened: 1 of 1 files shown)


public/src/editor/media-gallery-editor.js
```diff
@@ -49,25 +49,34 @@ define( [ "util/lang", "util/xhr", "util/keys", "util/mediatypes", "editor/edito
...
-      _durationInput.value = Time.toTimecode( _media.duration );
+      _durationInput.value = previousDurationTimeCode;
       return;
     }
-    duration = Time.toTimecode( duration );
-    if ( _durationInput.value !== duration ) {
-      _durationInput.value = Time.toTimecode( duration );
+
+    // Don't allow zero or lower values to be passed in.
+    if ( durationSeconds <= 0 ) {
```
ScottDowne: You can put this if in the duration === "" a if with an ||

This way, you don't need to cache the previousDurationTimeCode, as you'll only be using it once.

PR #1607, opened 2013-03-21
Bug #3313 - Add Description field to share-editor

cadecairos:
https://webmademovies.lighthouseapp.com/projects/65733-popcorn-maker/tickets/3313


public/css/project-editor.less
```diff
@@ -0,0 +1,84 @@
...
+      right: 35px;
+      background-color: @green;
+      >span {
+        background-position: -336px -96px;
+        background-image: url( "../../../../resources/icons/glyphiconshalflings-white.png" );
+      }
+    }
+
+    .butter-editor-body {
+      top: 135px !important;
```
rossbruniges: is there a need for the !important? Would recommend not using this unless there is a very good reason...

public/src/layouts/project-editor.html
```diff
@@ -0,0 +1,66 @@
+<div class="butter-editor project-editor allow-scrollbar">
+  <div class="editor-title-card">Project
+    <input class="butter-project-url" type="text" readonly="readonly">
+    <a class="butter-btn btn-light butter-preview-link" target="_blank"><span class="icon"></span></a>
```
rossbruniges: There should be some anchor text in here saying 'preview' so AT users know the function of this button - it can be hidden using text-offset: -99999em or similar to keep the visuals as intended.

public/src/layouts/project-editor.html
```diff
@@ -0,0 +1,66 @@
...
+  </div>
+  <div class="error-message-container">
+    <div class="error-message"></div>
+  </div>
+  <ul class="butter-editor-tabs">
+    <li><a class="butter-btn project-tab butter-active" data-tab-name="settings">Settings</a></li>
+    <li><a class="butter-btn project-tab" data-tab-name="share">Share</a></li>
+    <li><a class="butter-btn project-tab" data-tab-name="embed">Embed</a></li>
+    <li><a class="butter-btn butter-view-source-link" target="_blank" data-tab-name="view-source">View Source</a></li>
+  </ul>
```
rossbruniges, line 14: To ensure that the buttons (and also the preview button on line 4) are in the keyboard tab order they should contain an href value. Equally you could instead use the <button> element.

If an anchor is used ensure to return false to stop the browser history navigating to #

public/src/layouts/project-editor.html
```diff
@@ -0,0 +1,66 @@
...
+  <div class="butter-editor-body butter-tabs-spacing scrollbar-container">
+    <div class="scrollbar-outer scrollbar-inner">
+
+      <!-- Settings Container -->
+      <div class="settings-container editor-container">
+        <!-- Needs an actual class for this style -->
+        <h4 style="margin-top: -5px;">Author</h4>
```
rossbruniges: Could we remove this and also remove 5px of top padding on the editor-container element instead?

public/src/layouts/project-editor.html
```diff
@@ -0,0 +1,66 @@
...
+          <input class="butter-project-thumbnail" type="text">
+        </div>
+        <div class="upload-container">
+          <div class="image-droparea">
+            <div class="droparea-text">Drag an image here from your desktop.</div>
+          </div>
+        </div>
+        <br>
+        <ul class="butter-editor-tabs">
+          <li><a class="butter-btn settings-tab" data-tab-name="url">URL</a></li>
```
rossbruniges: Same comment as on line 14 applies to these two

public/css/project-editor.less
cadecairos: I can try looking for another way. another top value is stepping on this one in butter-ui.css

public/src/layouts/project-editor.html
```diff
@@ -0,0 +1,61 @@
...
+  </ul>
+  <div class="butter-editor-body butter-tabs-spacing scrollbar-container">
+    <div class="scrollbar-outer scrollbar-inner">
+      <div class="error-message-container">
+        <div class="error-message"></div>
+      </div>
+
+      <!-- Settings Container -->
+      <div class="settings-container editor-container">
+        <!-- Needs an actual class for this style -->
```
mjschranz: This comment needs to go.

public/src/editor/module.js
```diff
@@ -209,10 +209,9 @@ define( [ "core/eventmanager", "core/trackevent", "./editor",
 
       if ( butter.config.value( "ui" ).enabled !== false ) {
 
-        // Set up views for share editor
+        // Set up views for project editor
         butter.listen( "ready", setupHeader );
         butter.listen( "autologinsucceeded", setupHeader );
-        butter.listen( "authenticated", setupHeader );
```
mjschranz, line 215: This needs to remain in the event a project is saved and "in the system" and someone logs out and then logs back in. In that case the project editor tab can't be clicked.

public/src/editor/project-editor.js
```diff
@@ -0,0 +1,189 @@
...
+        _shareTwitter = _rootElement.querySelector( ".butter-share-twitter" ),
+        _shareGoogle = _rootElement.querySelector( ".butter-share-google" ),
+        _embedDimensions = _embedSize.value.split( "x" ),
+        _embedWidth = _embedDimensions[ 0 ],
+        _embedHeight = _embedDimensions[ 1 ],
+        _dropArea = _rootElement.querySelector( ".image-droparea" ),
+        _thumbnailInput = _rootElement.querySelector( ".butter-project-thumbnail" ),
+        _projectTabs = _rootElement.querySelectorAll( ".project-tab" ),
+        _this = this,
+        _hasBeenSavedOnce = false,
```
mjschranz: All of this `_hasBeenSavedOnce` code is pointless in the editors current state since we are keeping in inaccessible when the project is dirty.

public/templates/assets/editors/editorhelper.js
```diff
@@ -351,7 +351,11 @@
               return;
             }
 
-            trackEvent.update( { src: imgURI } );
+            if ( trackEvent ) {
+              trackEvent.update( { src: imgURI } );
```
mjschranz, line 355: An oversight on my part, but this should use the callback here as well if provided or do what it currently is otherwise.

public/templates/basic/index.html
```diff
@@ -41,11 +41,11 @@
       var s = document.getElementsByTagName('script')[0]; s.parentNode.insertBefore(ga, s);
     })();
   </script>
-  <meta property="og:title" content="Popcorn Maker">
-  <meta property="og:description" content="Video beyond the box. Popcorn Maker makes it easy to enhance, remix and share web video. ">
-  <meta property="og:url" content="https://popcorn.webmaker.org">
+  <meta property="og:title" content="Popcorn Maker" data-butter-exclude>
+  <meta property="og:description" content="Video beyond the box. Popcorn Maker makes it easy to enhance, remix and share web video. " data-butter-exclude>
+  <meta property="og:type" content="website" data-butter-exclude>
+  <meta property="og:url" content="https://popcorn.webmaker.org" data-butter-exclude>
```
mjschranz, line 47: There isn't a need for any changes here, nor adding `data-butter-exclude`. These `meta` tags are only used if someone links to `popcorn.webmaker.org/templates/basic` otherwise what's in the embed-shell is what's seen.

cadecairos, line 47: These changes are necessary because the head tag is copied and dumped into embed.jade.

lib/models/project.js
```diff
@@ -53,6 +57,9 @@ module.exports = function(sequelize, DataTypes) {
     },
     remixedFrom: {
       type: DataTypes.INTEGER
+    },
+    thumbnail: {
+      type: DataTypes.STRING
```
jbuck: Any new columns must allow nulls

PR #1614, opened 2013-03-26
Bug #3430 - Secure default cookie config

jbuck:


lib/default-config.js
```diff
@@ -17,7 +17,11 @@ module.exports = {
   },
   "session" : {
     "secret": "thisisareallyreallylongsecrettoencryptcookies",
-    "duration": 2419200000
+    "cookie": {
+      "maxAge": 2419200000,
+      "httpOnly": true,
```
mjschranz, line 22: According to the API this defaults to true, so we could technically remove it but that's a nit. Might be better left in for better understanding of people reading this later.

PR #1619, opened 2013-03-27
Bug #3401 - Only use LESS middleware

jbuck:


make.js
```diff
@@ -29,102 +27,118 @@ function gitDescribe( repoRoot ) {
   return version;
 }
 
-// To supress CSS warnings/errors for a particular line, end the line
```
humphd: I think removing support for `/* csslint-ignore */` is wrong.  Let's keep it so we have access to it for when we need to override CSSLint and not have our builds break.

make.js
```diff
@@ -29,102 +27,118 @@ function gitDescribe( repoRoot ) {
...
   find( "public/" ).filter(function( filename ) {
     return filename.match( /\.css$/ ) && !filename.match( /^public\/test/ ) &&
            !filename.match( /^public\/external/ );
   }).forEach(function( filename ) {
-    checkCSSFile( filename, warnings, errors );
+    checkCSSFile( cat( filename ), filename );
   });
 
+  [
+    "public/css/butter.ui.less",
```
humphd, line 113: If we're going to statically define all these in here, can we do it at the top in a variable that's easy to see?

make.js
```diff
@@ -29,102 +27,118 @@ function gitDescribe( repoRoot ) {
...
+    "public/templates/assets/css/jquery-ui/jquery.ui.butter.less",
+    "public/templates/assets/plugins/twitter/popcorn.twitter.less",
+    "public/templates/assets/plugins/wikipedia/popcorn.wikipedia.less",
+    "public/templates/basic/style.less"
+  ].forEach(function( lessFile ) {
+    lessToCSS( lessFile, function( err, cssFile ) {
+      if ( err ) {
+        passed = false;
+        console.error( "%sError: %s in %s:%d:%d",
+                     err.type, err.message, err.filename, err.line, err.column );
```
humphd: Move this over a few spaces to line `err.type` up with `"%sError`

make.js
```diff
@@ -265,57 +278,16 @@ function checkHTML() {
...
-}
-
 target.check = function() {
   checkJS();
-  buildCSS();
   checkCSS();
-  checkHTML();
 
-  exit( passed ? 0 : 1 );
+  // checkCSS is async, so we need to do this ugly hack. sorry
```
humphd: Obviously you've thought of this, but I don't get why you're not passing a callback so you don't have to do this.  You'd be better off using some kind of event and signalling when you're done doing the check.  Why not use an EventEmitter and register a listener?  This seems like it's going to cause issues to me.

server.js
```diff
@@ -56,8 +52,21 @@ function setupStore( storeConfig ) {
 }
 
 app.configure( function() {
+  var optimize = config.NODE_ENV !== "development",
+      tmpDir = require( "os" ).tmpdir() + "butter/";
```
humphd:
```
dave@hospitality:~/repos/butter(t3401)$ node --version
v0.8.14
dave@hospitality:~/repos/butter(t3401)$ node
> require('os').tmpdir()
TypeError: Object #<Object> has no method 'tmpdir'
    at repl:1:16
    at REPLServer.self.eval (repl.js:109:21)
    at Interface.<anonymous> (repl.js:248:12)
    at Interface.EventEmitter.emit (events.js:96:17)
    at Interface._onLine (readline.js:200:10)
    at Interface._line (readline.js:518:8)
    at Interface._ttyWrite (readline.js:736:14)
    at ReadStream.onkeypress (readline.js:97:10)
    at ReadStream.EventEmitter.emit (events.js:126:20)
    at emitKey (readline.js:1058:12)
```

make.js
jbuck: Re-added on line 99

make.js
```diff
@@ -29,102 +28,127 @@ function gitDescribe( repoRoot ) {
...
+                   filename, msg.line, msg.col, msg.message );
+  });
+}
 
+function checkCSS( callback ) {
   echo( "" );
-  echo( "# Linting CSS files" );
+  echo( "# Linting CSS and LESS files" );
+
+  var files = [
```
humphd, line 112: I still don't like this being buried deep in the makefile.  Adding a new less file requires a change here, but that's not going to be obvious.  Should we build the list of `.less` files dynamically instead?

make.js
```diff
@@ -29,102 +28,127 @@ function gitDescribe( repoRoot ) {
...
-  find( "public/" ).filter(function( filename ) {
+  files = files.concat( find( "public/" ).filter( function( filename ) {
     return filename.match( /\.css$/ ) && !filename.match( /^public\/test/ ) &&
            !filename.match( /^public\/external/ );
-  }).forEach(function( filename ) {
-    checkCSSFile( filename, warnings, errors );
-  });
+  })).sort();
+
+  var q = async.queue( function( lessFile, cb ) {
```
humphd, line 128: `cb` is fine here, I missed the nesting.  Ignore my other comment.

make.js
jbuck, line 112: I attempted to lint all of the lint files, but because of variable declaration in globals.less, it doesn't work.

PR #1620, opened 2013-03-28
Bug #3441 - Rewrite xhr module

jbuck:


public/src/butter.js
```diff
@@ -60,7 +60,7 @@ window.Butter = {
             "core/target", "core/media",
             "./modules", "./dependencies", "./dialogs",
             "dialog/dialog", "editor/editor", "ui/ui",
-            "util/xhr", "util/lang", "util/tutorial",
+            "util/xhr2", "util/lang", "util/tutorial",
```
humphd: Cute, but let's just go all in and rename it to xhr.js

public/src/butter.js
```diff
@@ -70,7 +70,7 @@ window.Butter = {
             Target, Media,
             Modules, Dependencies, Dialogs,
             Dialog, Editor, UI,
-            XHR, Lang, Tutorial,
+            xhr, Lang, Tutorial,
```
humphd, line 73: I really like this style (lower case for things that aren't ctors).  Let's do this everywhere.

public/src/butter.js
```diff
@@ -961,31 +944,10 @@ window.Butter = {
...
-
-        if( xhr.status === 200 || xhr.status === 0 ){
-          try{
-            userConfig = Config.parse( xhr.responseText );
-          }
-          catch( e ){
-            throw new Error( "Butter config file not formatted properly." );
-          }
+        xhr.get( butterOptions.config, function( response ) {
+          var userConfig = Config.reincarnate( response );
```
humphd, line 948: This can't throw like `parse` could?

public/src/core/project.js
```diff
@@ -317,14 +317,14 @@ define( [ "core/eventmanager", "core/media", "util/sanitizer" ],
         return;
       }
 
-      var projectJSON = JSON.stringify({
+      var projectJSON = {
```
humphd: Let's rename this so it's clear this is actually a fully inflated Object.  `project` or `projectData`.

public/src/butter.js
```diff
@@ -961,31 +944,10 @@ window.Butter = {
...
           readConfig( userConfig );
-        }
-        else{
-          _this.dispatch( "configerror", _this );
```
humphd, line 987: If we do dump this error case, I notice that there are no consumers of `configerror` so it should be fine for it to silently die.

public/src/cornfield/module.js
```diff
@@ -2,99 +2,53 @@
...
-                  callback({ error: "an unknown error occured" });
-                }
-              }
-            });
+            // If there was an error of some sort, callback on that
+            callback( response );
+          });
         } else {
-          callback(undefined);
+          callback( undefined );
```
humphd: `callback()`

public/src/butter.js
jbuck, line 948: The xhr module should have converted it to an object already. It'd be possible to have that conversion fail if the json wasn't sent with the mime type `application/json`, but think that's a use-case that we should ignore

public/src/butter.js
jbuck, line 987: Yeah, that's why I nuked it here

public/src/util/xhr.js
```diff
@@ -3,133 +3,79 @@
...
+        if ( !options.success ) {
+          return;
         }
 
-        if ( callback ) {
-          callback.apply( this, arguments );
-        }
+        var response = this.responseText;
 
+        if ( this.getResponseHeader( "Content-Type" ).match( "application/json" ) ) {
```
humphd, line 30: Will we ever get burned by a mime type of `text/json` here?

public/src/util/xhr.js
```diff
@@ -3,133 +3,79 @@
...
+          try {
+            response = JSON.parse( this.responseText );
+          } catch ( ex ) {}
```
humphd: So if this fails somehow, we'll be passing through `response` to the callback below, and we know it's somehow capable of causing a `JSON.parse` to throw.  Should we null it first?  Is there a reason we're passing it through if we can't cope with it?

public/src/util/xhr.js
```diff
@@ -3,133 +3,79 @@
...
-
-  var __csrfToken = "",
-      __onCSRFTokenAcquired;
-
-  define( [], function() {
-
-    function generateCSRFOnReadyStateHandler( callback ) {
-      return function() {
+  var xhrModule = {
+    ajax: function( options ) {
```
humphd, line 15: Why expose `ajax`?

routes/index.js
```diff
@@ -231,21 +231,13 @@ module.exports = function routesCtor( app, User, filter, sanitizer, stores, util
...
-    req.addListener( 'end', function() {
-      var name = generateUniqueName([
-        { name: 'dt', value: formatDate() },
-        { name: 'deployment', value: deploymentType }
-      ]);
-      store.write( name, s, function() {
-        res.writeHead( 200, { 'content-type': 'text/plain' } );
-        res.end();
-      });
+    store.write( name, JSON.stringify(req.body), function() {
```
humphd: Spaces around `req.body`

public/src/util/xhr.js
jbuck, line 15: Instead of having people add new functions like `getUntilComplete()` in the previous iteration of this code, I'd rather they use `ajax()` directly. Then they can customize the request as they see fit.

public/src/util/xhr.js
jbuck, line 30: `application/json` is the assigned mime type for JSON: http://www.ietf.org/rfc/rfc4627.txt. I know within our repo, there's nothing sending or receiving `text/json` (anymore). If an external service is using text/json, they're doing it wrong and we should file a bug against them.

public/src/util/xhr.js
jbuck: I was thinking along the lines of jquery's `dataType` parameter: http://api.jquery.com/jQuery.ajax/. It defaults to an intelligent guess, and I see that it'll actually throw an error if the JSON fails to parse. Maybe we should return null here...

public/src/cornfield/module.js
```diff
@@ -2,99 +2,53 @@
...
-        args = xhrPostQueue.shift();
-        XHR.post.apply( this, args );
-      }
-
-      XHR.setCSRFTokenAcquiredCallback( null );
+    this.login = function( callback ) {
+      navigator.id.get( function( assertion ) {
+        if ( assertion ) {
+          xhr.post( "/persona/verify", { assertion: assertion }, function( response ) {
+            if (response.status === "okay") {
```
mjschranz: Space nit.

public/src/cornfield/module.js
```diff
@@ -2,99 +2,53 @@
...
 
-      sendXHRPost = XHR.post;
-    });
+              // Get email, name, and username after logging in successfully
+              whoami( function( data ) {
+                callback( data );
```
mjschranz: Couldn't you just pass the same callback along here? Nit.

public/src/util/xhr.js
```diff
@@ -3,133 +3,91 @@
...
-       * @param {Object} extraRequestHeaders: Optional. Header key/value pairs to add to the request before it is sent to the server.
-       */
-      "getUntilComplete": function( url, callback, mimeTypeOverride, extraRequestHeaders ) {
-        XHR.get( url, function( e ) {
-          if ( this.readyState === XHR.DONE ) {
-            callback.call( this, e );
-          }
-        }, mimeTypeOverride, extraRequestHeaders );
-      },
+      if (!options.error) {
```
mjschranz: Space nit.

public/src/util/xhr.js
```diff
@@ -3,133 +3,91 @@
...
         }
 
-        if ( callback ) {
-          callback.apply( this, arguments );
-        }
+        var response = this.responseText;
 
+        if ( this.getResponseHeader( "Content-Type" ).match( "application/json" ) ) {
+          try {
+            response = JSON.parse( this.responseText );
```
mjschranz, line 32: We previously did

``` javascript
JSON.parse( this.reponse || this.responseText )
```

Granted all browsers we support do support the responseText property but for sanities sake I think we should keep that as it was no?

jbuck, line 32: No, because the `response` property can change type depending on what you pass to `responseType`: https://developer.mozilla.org/en-US/docs/DOM/XMLHttpRequest#Properties . `responseText` is always guaranteed to be a String.

PR #1634, opened 2013-04-02
Bug #3440 - Switch to RequireJS middleware

jbuck:


server.js
```diff
@@ -66,6 +67,47 @@ app.configure( function() {
...
+          baseUrl: WWW_ROOT + "/src/",
+          preserveLicenseComments: false,
+          include: [ "webmakernav" ],
+          mainConfigFile: WWW_ROOT + "/src/webmakernav.js",
+          wrap: {
+            startFile: __dirname + "/tools/wrap.start",
+            endFile: __dirname + "/tools/wrap.end"
+          }
+        }
+      }
```
humphd, line 99: This options block feels pretty huge, and it's somewhat distracting from what you want to focus on in this code, namely, the middleware we're using, and its order.  Can you do this somewhere else, and pass it in?  Or move more of this to config?

jbuck, line 99: I added a `defaults` block that sets all of the common RequireJS options. Does that work, or do you still think that the options block should move?

I would rather not move the options from server.js because:

1) I don't expect people to override RequireJS settings in local.json
2) I can't take advantage of the `server.js`-relative paths (`WWW_ROOT`, `__dirname`)
3) To the greatest extend possible, I prefer middleware configuration to be located where the middleware is initialized. `server.js` does have a ton of code inside of it that should be refactored into other files but that's mostly because of publish, not middleware.

server.js
```diff
@@ -53,7 +54,8 @@ function setupStore( storeConfig ) {
 
 app.configure( function() {
   var optimize = config.NODE_ENV !== "development",
-      tmpDir = path.normalize( require( "os" ).tmpDir() + "/butter/" );
+      tmpDir = path.normalize( require( "os" ).tmpDir() + "/butter-" +
+        (Math.random() * 0x100000000 + 1).toString(36) + "/" );
```
Pomax: wait, what does this do? generate a unique dir? Because good Math.random() behaviour is to possibly generate the same number many times in a row, and I don't see a check to test whether the dir already exists with a while(exists) { tryagain } pattern.

PR #1635, opened 2013-04-03
[bug-857278] Using the new API's to pull out data into a page.

rossbruniges:
This is super super Alpha but let's see where this lands....

### DIFF
--- a/public/src/firehose.js
+++ b/public/src/firehose.js
@@ -1,0 +1,32 @@
+function init( window, document ) {
+  var require = requirejs.config({
+    baseUrl: "/src",
+    paths: {
+      text: "../external/require/text"
+    }
+  });
+
+  require( [ "ui/firehose", "ui/webmakernav/webmakernav" ], function( Firehose, WebmakerNav ) {
+    Firehose.init();
+    WebmakerNav.call( {}, {
+      hideLogin: true,
+      hideFeedback: true,
+      container: document.getElementById( "webmaker-nav" ),
+      feedbackCallback: function() {}
+    });
+  });
+}
+
+document.addEventListener( "DOMContentLoaded", function() {
+  // Source tree case vs. require-built case.
+  if ( typeof require === "undefined" ) {
+    var requireScript = document.createElement( "script" );
+    requireScript.src = "../../external/require/require.js";
+    requireScript.onload = function() {
+      init( window, window.document );
+    };
+    document.head.appendChild( requireScript );
+  } else {
+    init( window, window.document );
+  }
+}, false );
--- a/public/src/ui/firehose.js
+++ b/public/src/ui/firehose.js
@@ -1,0 +1,56 @@
+define( [ "util/xhr" ], function( xhr ) {
+
+  function insertAfter( referenceNode, newNode ) {
+    referenceNode.parentNode.insertBefore( newNode, referenceNode.nextSibling );
+  }
+
+  function renderResponse( data, sibling, parent ) {
+    var frag = document.createDocumentFragment();
+    if ( !data.length ) {
+      var li = document.createElement( "li" );
+      li.innerHTML = "No remixes";
+      parent.appendChild( li );
+    } else {
+      data.forEach( function( d ) {
+        var li = document.createElement( "li" ),
+            viewUrl = "<a href='" + d.publishUrl + "'>" + d.name + "</a>",
+            remixUrl = "<a data-remixId='" + d.id + "' class='remix' href='remixes/" + d.id + "'>show me the remixes</a>";
+        li.innerHTML = viewUrl + " - " + remixUrl;
+        frag.appendChild( li );
+      } );
+      parent.appendChild( frag );
+    }
+    insertAfter( sibling, parent );
+  }
+
+  function ignition( e ) {
+    var target = e.target;
+    if ( target.className === "remix" ) {
+      var remixId = target.getAttribute( "data-remixId");
+      xhr.get( "/api/project/" + remixId + "/remixes", function ( response ) {
+        renderResponse ( response.results, target, document.createElement( "ol" ) );
+        target.parentNode.removeChild( target );
+      } );
+      e.preventDefault();
+    }
+  }
+
+  function init() {
+    var create = document.getElementById( "new" ),
+        apiCreates = document.createElement( "ol" ),
+        updates = document.getElementById( "updated" ),
+        apiUpdates = document.createElement( "ol" );
+    xhr.get( "/api/projects/recentlyCreated", function( response ) {
+      renderResponse( response.results, create, apiCreates );
+    } );
+    xhr.get( "/api/projects/recentlyUpdated", function( response ) {
+      renderResponse( response.results, updates, apiUpdates  );
+    } );
+    apiCreates.addEventListener( "click", ignition );
+    apiUpdates.addEventListener( "click", ignition );
+  }
+
+  return {
+    "init": init
+  };
+});
--- a/views/firehose.jade
+++ b/views/firehose.jade
@@ -1,0 +1,28 @@
+!!! 5
+html(lang='en')
+  head
+    title Projects
+    link(rel='stylesheet', href='/media/css/sandstone.css')
+    link(rel='stylesheet', href='/css/butter.ui.css')
+    link(rel='stylesheet', href='//www.mozilla.org/tabzilla/media/css/tabzilla.css')
+    style
+      html {
+        background: #F9F9F9;
+        min-height: 101%;
+      }
+      body {
+        overflow: visible;
+      }
+      h1 {
+        padding-top: 1em;
+      }
+  body
+    nav#webmaker-nav
+    div#wrapper
+      div.container
+        h1 Project APIs
+        p We now have API's, and by gad are we going to use them!
+        #fooer
+        h2#new Recent makes
+        h2#updated Recently updated
+        script(src='/src/firehose.js')
--- a/routes/index.js
+++ b/routes/index.js
@@ -298,6 +298,10 @@ module.exports = function routesCtor( app, Project, filter, sanitizer,
     });
   }
 
+  app.get( '/internal/firehose', filter.isStorageAvailable, function( req, res ) {
+    res.render( 'firehose.jade' );
+  });
+
   // Store crash reports
   app.post( '/crash', function( req, res ) {
     storeData( req, res, stores.crash );


server.js
```diff
@@ -309,7 +313,7 @@ app.get( '/dashboard', filter.isStorageAvailable, function( req, res ) {
 
   Project.findAll( { email: email }, function( err, docs ) {
     var userProjects = [];
-
+
```
mjschranz: Whitespace needs to be removed.

public/src/ui/projectS.js
```diff
@@ -0,0 +1,38 @@
+define( [ "util/xhr"], function( xhr ) {
+
+    var init, render_response, insertAfter;
```
mjschranz: Personally I think this can all be done server side fairly easily. Going to take a quick stab at it.

rossbruniges: An nice - that was something I was going to ask after it landed, I would much rather it be done server side. All this express stuff is so new to me it's not even funny :)

public/src/projectS.js
```diff
@@ -0,0 +1,23 @@
...
+    projectS.init();
+  });
+}
+
+document.addEventListener( "DOMContentLoaded", function() {
+  // Source tree case vs. require-built case.
+  if ( typeof require === "undefined" ) {
+    var requireScript = document.createElement( "script" );
+    requireScript.src = "../../external/require/require.js";
+    requireScript.onload = function() {
```
andrewhayward: Might not be an issue here, but script.onload doesn't work in [<IE9].

public/src/ui/projectS.js
rossbruniges: So speaking to @jbuck as the point of the ticket was to create a page using the APIs we're going to leave as is. I imagine that this is just the start of something new and shiny!!

public/src/ui/projectS.js
```diff
@@ -0,0 +1,40 @@
...
+    referenceNode.parentNode.insertBefore( newNode, referenceNode.nextSibling );
+  };
+
+  render_response = function( data, container, to ) {
+    var frag = document.createDocumentFragment(), list;
+    data.forEach(function( d ) {
+      var li = document.createElement( "li" ),
+          id = d.id,
+          published_id = id.toString( 36 ),
+          url = "http://popcorn.webmadecontent.org/" + published_id;
```
rossbruniges: These 2 lines are a bit hacky and will only work in production - is there a better way to achieve this?

public/src/projectS.js
rossbruniges: Copy and paste code so hopefully this is a know issue...

public/src/firehose.js
```diff
@@ -0,0 +1,32 @@
...
+  var require = requirejs.config({
+    baseUrl: "/src",
+    paths: {
+      text: "../external/require/text"
+    }
+  });
+
+  require( [ "ui/firehose", "ui/webmakernav/webmakernav" ], function( firehose, WebmakerNav ) {
+    firehose.init();
+    WebmakerNav.call({}, {
```
humphd: Extra space between `( {}`

public/src/firehose.js
```diff
@@ -0,0 +1,32 @@
...
+      hideLogin: true,
+      hideFeedback: true,
+      container: document.getElementById( 'webmaker-nav' ),
```
humphd: Don't mix quote types.  Stick with double-quotes throughout.

public/src/firehose.js
```diff
@@ -0,0 +1,32 @@
+function init( window, document ) {
```
humphd, line 1: I find it a bit confusing for us to have `firehose.js` and `ui/firehose.js`.  Since nothing else includes `ui/firehose.js`, does it need to be its own file?  Why not define it in here?

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
+define( [ "util/xhr" ], function( xhr ) {
+
+  var init, render_response, insertAfter, ignition;
+
+  insertAfter = function( referenceNode, newNode ) {
```
humphd: Just use `function insertAfter` instead of creating a variable.

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+    referenceNode.parentNode.insertBefore( newNode, referenceNode.nextSibling );
+  };
+
+  render_response = function( data, container, to ) {
```
humphd: Same here.

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+            remix_url = "<a class='remix' href='remixes/" + d.id + "'>show me the remixes</a>";
+        li.innerHTML = view_url + ' - ' + remix_url;
+        frag.appendChild( li );
+      });
+      to.appendChild(frag);
+    }
+    insertAfter(container, to);
+  };
+
+  ignition = function( e ) {
```
humphd: Same here.

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+          remix_id = remix_href[remix_href.length - 1];
+      xhr.get( "/api/project/" + remix_id + "/remixes", function (response) {
+        render_response ( response.results, target, document.createElement( "ol" ) );
+        target.parentNode.removeChild(target);
+      });
+      e.preventDefault();
+    }
+  };
+
+  init = function() {
```
humphd: Same here

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+
+  var init, render_response, insertAfter, ignition;
+
+  insertAfter = function( referenceNode, newNode ) {
+    referenceNode.parentNode.insertBefore( newNode, referenceNode.nextSibling );
+  };
+
+  render_response = function( data, container, to ) {
+    var frag = document.createDocumentFragment(), list;
+    if (!data.length) {
```
humphd: Spaces around args, should be `( !data.length )`

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+      var li = document.createElement( "li" );
+      li.innerHTML = "No remixes yet :(";
```
humphd: Just use "No remixes."

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+      to.appendChild(li);
```
humphd: Spaces around args.

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+    } else {
+      data.forEach(function( d ) {
```
humphd: Spaces around args.

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+        var li = document.createElement( "li" ),
+            view_url = "<a href='" + d.publishUrl + "'>" + d.name + "</a>",
+            remix_url = "<a class='remix' href='remixes/" + d.id + "'>show me the remixes</a>";
```
humphd: Might be useful to add to the data we return from the API with that link built already, since we might change how we do it later.

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+        li.innerHTML = view_url + ' - ' + remix_url;
```
humphd: Double quotes

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+        frag.appendChild( li );
+      });
+      to.appendChild(frag);
```
humphd: Spaces around args.

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+    insertAfter(container, to);
```
humphd: Spaces around args.

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  ignition = function( e ) {
+    var target = e.target;
+    if (target.className === 'remix') {
```
humphd: Spaces around args, double quotes.

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+      var remix_href = target.href.split('/'),
```
humphd: Yeah, this points to a hole in our API and the data we return.  Let's just return the URL you need instead.

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+          remix_id = remix_href[remix_href.length - 1];
+      xhr.get( "/api/project/" + remix_id + "/remixes", function (response) {
+        render_response ( response.results, target, document.createElement( "ol" ) );
+        target.parentNode.removeChild(target);
```
humphd: Spaces around args.

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
+define( [ "util/xhr" ], function( xhr ) {
+
+  var init, render_response, insertAfter, ignition;
```
humphd: JavaScript variables are done like this:

``` javascript
var foo, // lowercase on first letter, unless a ctor
  fooBar, // lowercase on first letter, upper on subsequent
  fooBarBaz;
```

Don't use python style naming (i.e., `foo_bar`)

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+        updates = document.getElementById( "updated" ),
+        api_updates = document.createElement( "ol" ),
+        lists = document.getElementsByTagName('ol');
+    xhr.get( "/api/projects/recentlyCreated", function( response ) {
+      render_response( response.results, create, api_creates );
+    } );
+    xhr.get( "/api/projects/recentlyUpdated", function( response ) {
+      render_response( response.results, updates, api_updates  );
+    } );
+    api_creates.addEventListener('click', ignition );
```
humphd: Spaces around args, double quotes.

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+    api_updates.addEventListener('click', ignition );
```
humphd: Same.

public/src/ui/firehose.js
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  return {
+    'init': init
```
humphd: Double quotes.

views/firehose.jade
```diff
@@ -0,0 +1,15 @@
...
+    link(rel='stylesheet', href='/css/butter.ui.css')
+    link(rel='stylesheet', href='//www.mozilla.org/tabzilla/media/css/tabzilla.css')
+  body
+    nav#webmaker-nav
+    div.container
+      h1 Project APIs
+      p We now have API's, and by gad are we going to use them!
+      h2#new Recent makes
+      h2#updated Recently updated
+      script(src='/src/firehose.js')
```
humphd: Add newline at end of file.

public/src/ui/firehose.js
rossbruniges: A poorly named attribute - have redone this bit to have it make more sense

public/src/firehose.js
rossbruniges, line 1: So at this point I was copying that style of webmakernav - which may not be ideal. 

This way does feel potentially a bit cleaner though and doesn't overload the file. Happy to let the people decide...

public/src/ui/firehose.js
rossbruniges: Personally I think this is the right place for the strong building. Here could be multiple uses of the API all requiring different HTML patterns so I would rather err on the side of flexibility with API usage before including HTML snippets.

jbuck:
Closing due to mozilla/webmaker-firehose being setup.

rossbruniges:
And here is webmaker-firehose: https://github.com/mozilla/webmaker-firehose

:beer:

PR #1637, opened 2013-04-03
[t3252] Adding media clips without a protocol shouldn't hang

bl4de:
New pull-request.

### DIFF
--- a/public/src/editor/media-gallery-editor.js
+++ b/public/src/editor/media-gallery-editor.js
@@ -217,6 +217,14 @@ define( [ "util/lang", "util/uri", "util/keys", "util/mediatypes", "editor/edito
       return;
     }
 
+
+    var checkUrl = URI.parse( url );
+    if ( checkUrl.protocol !== "" ) {
+      url = ( checkUrl.protocol === "https" ) ? "https://" + checkUrl.source : "http://" + checkUrl.source;
+    } else {
+      url = "http://" + checkUrl.source;
+    }
+
     data.source = url;
     data.type = "sequencer";
     _mediaLoadTimeout = setTimeout( function() {


rossbruniges:
Awesome - that's looking much nicer :)

public/src/editor/media-gallery-editor.js
```diff
@@ -217,6 +217,14 @@ define( [ "util/lang", "util/uri", "util/keys", "util/mediatypes", "editor/edito
       return;
     }
 
+
+    var checkUrl = URI.parse( url );
+    if ( checkUrl.protocol !== "" ) {
+      url = ( checkUrl.protocol === "https" ) ? "https://" + checkUrl.source : "http://" + checkUrl.source;
+    } else {
+      url = "http://" + checkUrl.source;
+    }
```
mjschranz, line 226: You could simplify this to be something like...

``` javascript
url = checkUrl.protocol ? checkUrl.toString() : "http://" + checkUrl.toString();
```

An empty string equates to a false value afterall.

rossbruniges, line 226: That's pretty terse syntax though - are we using ternaries across the project much?

bl4de, line 226: @mjschranz I did it as normal 'if' statement to keep code more clear.
If I should use ternary, I'll fix it, of course.

mjschranz, line 226: Truthfully we don't have much of a standard for when we use the ternary operator. It usually comes down to preference of whoever is reviewing. I personally would use it in this case because it's a fairly simple if check and the line itself doesn't stretch out very far. Plus the code itself is fairly self explanatory.

Let whoever is reviewing this decide.

ScottDowne, line 226: I prefer simple if statements myself.

What about something like this?

``` javascript
url = "http://" + checkUrl.source;

if ( checkUrl.protocol === "https" ) {
  url = "https://" + checkUrl.source;
}
```

PR #1641, opened 2013-04-03
Bug 857724 - Prevent same source videos triggering reload because of difference in butteruid

mjschranz:


public/src/editor/sequencer-editor.js
```diff
@@ -384,20 +384,28 @@ define( [ "util/mediatypes", "editor/editor", "util/time",
...
+        if ( newSource === oldSource ) {
+          return;
+        }
+
         MediaUtils.getMetaData( updateOptions.source[ 0 ], function( data ) {
           _mediaType = data.type;
           updateOptions.duration = data.duration;
           updateOptions.denied = data.denied;
           updateOptions.title = data.title;
-          if ( _mediaType === "html5" ) {
```
ScottDowne, line 392: mediatype's check already returns HTML5 as a default. We don't need this change, reads better as === "html5"

public/src/editor/sequencer-editor.js
```diff
@@ -384,20 +384,28 @@ define( [ "util/mediatypes", "editor/editor", "util/time",
         }
         updateOptions.source[ 0 ] = updateOptions.source[ 0 ] || _popcornOptions.source[ 0 ];
 
+        var newSource = URI.stripUnique( updateOptions.source[ 0 ] ).toString(),
+            oldSource = URI.stripUnique( _popcornOptions.source[ 0 ] ).toString();
```
ScottDowne: You are only using the above variables once.

Might not be worth caching them, and just checking against the stripUnique calls in the if?

Not sure, though, what do you think?

PR #1643, opened 2013-04-03
Bug #857762 - 1. Add support for description field

cadecairos:
1. Change share-editor to project-editor

https://bugzilla.mozilla.org/show_bug.cgi?id=857762


public/css/editors.less
```diff
@@ -406,7 +416,7 @@
     > input, textarea, select {
       float: left;
       margin-left: 5px;
-      width: @_editorInnerWidth - 85;
+      width: 210px;
```
jbuck: Probably a good idea to stick with a variable for this

PR #1649, opened 2013-04-04
[#3200] Deselect track events on click, not mouseodwn to allow blur even...

ScottDowne:
...ts to happen first.

### DIFF
--- a/public/src/editor/base-editor.js
+++ b/public/src/editor/base-editor.js
@@ -25,7 +25,6 @@ define( [ "core/eventmanager", "util/scrollbars", "ui/widget/tooltip", "ui/widge
     extendObject.rootElement.style.display = "none";
     // Attach the editor's root element to the given parentElement.
     parentElement.appendChild( rootElement );
-console.log( parentElement, rootElement );
     // Used when applyExtraHeadTags is called -- see below
     var _extraStyleTags = [],
         _extraLinkTags = [];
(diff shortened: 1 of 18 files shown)


public/src/editor/base-editor.js
```diff
@@ -15,14 +15,16 @@ define( [ "core/eventmanager", "util/scrollbars", "ui/widget/tooltip", "ui/widge
    * @param {DOMElement} rootElement: The root element to which the editor's content will be attached
    * @param {Object} events: Events such as 'open' and 'close' can be defined on this object to be called at the appropriate times
    */
-  function BaseEditor( extendObject, butter, rootElement, events ) {
+  function BaseEditor( extendObject, butter, rootElement, parentElement, events ) {
```
mjschranz, line 18: Correct me if I'm wrong, but I'm pretty sure `parentElement` is always:

``` html
<div class="butter-editor-content" >
...
</div>
```

If I'm right, I'm wondering why it needs to be something passed along the entire way through. Can't it just be a hard coded value? Or is this somehow too risky/rigid of a design?

public/src/editor/base-editor.js
```diff
@@ -73,8 +70,7 @@ define( [ "core/eventmanager", "util/scrollbars", "ui/widget/tooltip", "ui/widge
      * Closes the editor
      */
     extendObject.close = function() {
-      // Remove the editor's root element from the element to which it was attached
-      extendObject.rootElement.parentNode.removeChild( extendObject.rootElement );
```
mjschranz, line 77: I'm fairly certain the STR I described in ticket is because of the following line being removed.

PR #1654, opened 2013-04-06
Bug 858950 - Create a Recently Remixed end point api

humphd:


routes/index.js
```diff
@@ -43,33 +42,30 @@ module.exports = function routesCtor( app, Project, filter, sanitizer,
...
+    'recentlyRemixed'
+  ].forEach( function( name ) {
+    var endPoint = '/api/projects/' + name + '/:limit?',
+        projectMethod = 'find' + name[ 0 ].toUpperCase() + name.slice( 1 );
+
+    app.get( endPoint,
+      filter.isStorageAvailable,
+      filter.crossOriginAccessible,
+      function( req, res ) {
+        Project[ projectMethod ]( { limit: req.params.limit }, function( err, projects ) {
```
jbuck, line 61: Add a default limit of `20`, or make the `:limit` parameter required by removing the ?

Right now you can hit `/api/projects/recentlyCreated/ and have your browser blow up when it tries to load 20k items.

PR #1656, opened 2013-04-08
Requesting a pull to mozilla:master from k88hudson:t3368

k88hudson:


public/css/media-editor.less
```diff
@@ -3,127 +3,56 @@
...
     float: right;
   }
 
-  h1.add-new-media {
-    height: @titleBarHeight;
-    line-height: @titleBarHeight;
-    text-transform: none;
-    font-weight: normal;
-    color: #555;
+  .media-gallery-heading {
```
mjschranz, line 23: It seems like we could actually just make this a normal `<h1>` element now rather than having a div with specific styling because of base styles we have for editors. That is, unless you explicitly want this title smaller than the ones used for say the plugin list or the project editor. If that's the case, why?

public/src/ui/ui-widgets.js
```diff
@@ -0,0 +1,52 @@
...
+        this.classList.add( "tab-on" );
+
+        if ( !whichTab ) {
+          return;
+        }
+
+        for ( j = 0; j < panels.length; j++ ) {
+          panels[ j ].classList.remove( "tab-on" );
+        }
+        element.querySelector( "[data-tab=\"" + whichTab + "\"]" ).classList.add( "tab-on" );
```
mjschranz, line 39: Technically element has the chance of being undefined here. I would wrap this in a safe check of sorts to know whether or not it's actually there.

public/src/editor/media-gallery-editor.js
```diff
@@ -2,12 +2,11 @@
  * If a copy of the MIT license was not distributed with this file, you can
  * obtain one at https://raw.github.com/mozilla/butter/master/LICENSE */
 
-define( [ "util/lang", "util/uri", "util/keys", "util/mediatypes", "editor/editor",
+define( [ "util/lang", "util/uri", "util/keys","util/mediatypes", "UI/ui-widgets", "editor/editor",
```
mjschranz, line 5: Change to `ui/ui-widgets`

Also add a space to before the load of media types.

`,"util/mediatypes"` -> `, "util/mediatypes"`

public/src/layouts/media-editor.html
```diff
@@ -1,40 +1,42 @@
...
       </div>
-    </fieldset>
-
-    <div class="media-gallery">
-      <div class="media-gallery-heading">
-        Media Gallery
-      </div>
-      <div class="media-gallery-scroll-container scrollbar-container">
-        <ul class="media-gallery-list scrollbar-outer scrollbar-inner"></ul>
+    </div><!-- add-media-panel -->
```
mjschranz, line 30: Why the comment here? If anything it should be above the div.

PR #1668, opened 2013-04-17
[#862811] Delay loading of popcorn video and track event data until user...

ScottDowne:
... clicks on embed.

### DIFF
--- a/public/src/ui/widget/controls.js
+++ b/public/src/ui/widget/controls.js
@@ -5,13 +5,13 @@
 define( [ "util/lang", "util/time", "text!layouts/controls.html" ],
   function( LangUtils, Time, CONTROLS_LAYOUT ) {
 
-  function Controls( container, p, options ) {
+  function Controls( container, options ) {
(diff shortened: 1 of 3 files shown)


public/src/ui/widget/controls.js
```diff
@@ -30,11 +30,36 @@ define( [ "util/lang", "util/time", "text!layouts/controls.html" ],
...
         onRemixClick = options.onRemixClick || nop,
         onFullscreenClick = options.onFullscreenClick || nop,
-        onLogoClick = options.onLogoClick || nop;
+        onLogoClick = options.onLogoClick || nop,
+        init = options.init || nop;
+
+    function onInit() {
+
+      document.removeEventListener( "click", onInit, false );
+      function setPopcorn( popcorn ) {
```
mjschranz, line 39: Does this callback really need to be declared here? Seems to me like it doesn't, unless this is something specifically done for lint. Just a nit really.

PR #1674, opened 2013-05-02
Tutorial

ScottDowne:


public/src/ui/header.js
```diff
@@ -3,6 +3,10 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
 
   return function( butter, options ){
 
+    var make = Make({
+      apiURL: "http://localhost:6001"
```
ScottDowne: I need to fix this.

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "public/external/popcorn-js"]
 	path = public/external/popcorn-js
 	url = git://github.com/mozilla/popcorn-js.git
+[submodule "public/external/MakeAPI"]
```
humphd, line 4: Here's exactly why I don't want to land this stuff in butter, and would rather fork and do it in a popcorn.webmaker.org repo.

public/src/ui/header.js
humphd: Elsewhere in this patch you do things like login.mofostaging.net.  Why not use mofostaging here too?

PR #1676, opened 2013-05-07
T866923

ScottDowne:


public/src/ui/header.js
```diff
@@ -20,10 +24,8 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
         _projectMenu = _rootElement.querySelector( ".butter-project-menu" ),
         _projectMenuControl = _rootElement.querySelector( ".butter-project-menu-control" ),
         _projectMenuList = _projectMenu.querySelector( ".butter-btn-menu" ),
-        _tabzilla = _rootElement.querySelector( "#tabzilla" ),
```
jbuck, line 23: I don't think we can just delete Tabzilla

public/src/ui/header.js
```diff
@@ -3,6 +3,10 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
 
   return function( butter, options ){
 
+    var make = Make({
+      apiURL: "http://makeapi.mofostaging.net"
```
jbuck: Ah, now I see why you wanted a way to pass an environment config variable into require

public/templates/basic/template.js
```diff
@@ -10,7 +10,7 @@
         var script;
         EditorHelper.init( butter );
         script = document.createElement( "script" );
-        script.src = "//www.mozilla.org/tabzilla/media/js/tabzilla.js";
+        script.src = "//login.mofostaging.net/js/sso.js";
```
jbuck: Are you actually using sso yet? if not, nuke this

lib/utils.js
```diff
@@ -21,10 +21,16 @@ utils = {
     };
   },
   generatePublishUrl: function( id ) {
+    if ( !id ) {
```
jbuck: You don't need this change or the one below, especially because none of our consumers handle errors if you returned undefined

lib/utils.js
```diff
@@ -25,6 +25,9 @@ utils = {
       '/' + stores.publish.expand( utils.generateIdString( id ) ) );
   },
   generateIframeUrl: function( id ) {
+    if ( !id ) {
```
humphd: Why would we not have an `id`?  This looks like you're compensating for bugs in callers.  I get nervous when I see functions silently returning on error cases.

public/src/core/project.js
```diff
@@ -259,6 +273,10 @@ define( [ "core/eventmanager", "core/media", "util/sanitizer" ],
         _remixedFrom = json.remixedFrom;
       }
 
+      if ( json.remixUrl ) {
```
humphd: You can simplify this by just doing:

``` javascript
_remixUrl = json.remixUrl;
```

since it will either stay undefined (and therefore falsy) or get a truthy value.

public/src/ui/header.js
```diff
@@ -1,8 +1,12 @@
-define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data", "ui/webmakernav/webmakernav", "ui/widget/textbox", "ui/widget/tooltip" ],
-  function( Dialog, Lang, HEADER_TEMPLATE, UserData, WebmakerBar, TextBoxWrapper, ToolTip ) {
+define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "text!layouts/tutorial-list.html", "ui/user-data", "ui/webmakernav/webmakernav", "ui/widget/textbox", "ui/widget/tooltip" ],
+  function( Dialog, Lang, HEADER_TEMPLATE, TUTORIAL_LIST_TEMPLATE, UserData, WebmakerBar, TextBoxWrapper, ToolTip ) {
 
   return function( butter, options ){
 
+    var make = Make({
```
humphd: Where are you declaring Make?

public/src/ui/header.js
```diff
@@ -331,6 +313,86 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
...
+      var tutorialUrl;
+      if ( butter.project.id >= 0 || butter.project.remixedFrom >= 0) {
+        if ( butter.project.id >= 0 ) {
+          tutorialUrl = butter.project.publishUrl;
+        } else if ( butter.project.remixedFrom >= 0 ) {
+          tutorialUrl = butter.project.remixUrl;
+        }
+
+        make.tags( "tutorial:" + tutorialUrl ).then( function( err, results ) {
+          var tutorialView = document.createElement( "div" ),
```
humphd: Why have all this HTML being constructed in code vs requiring in a template text fragment?

public/src/ui/header.js
```diff
@@ -331,6 +313,86 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
...
+              iframeCover = document.createElement( "div" ),
+              iframe = document.createElement( "iframe" ),
+              closeButton = document.createElement( "div" ),
+              viewTitle = document.createElement( "div" ),
+              tutorialTemplate = Lang.domFragment( TUTORIAL_LIST_TEMPLATE, ".tutorial-template" ),
+              tutorialList = tutorialTemplate.querySelector( ".tutorial-list" );
+
+          if ( err ) {
+            return;
```
humphd: Silent failure?

public/src/ui/header.js
```diff
@@ -331,6 +313,86 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
...
+              iframeCover.style.display = "block";
+              tutorialView.removeEventListener( "mousedown", onCoverMouseDown, false );
+              document.addEventListener( "mouseup", onCoverMouseUp, false );
+            };
+
+            var createTutorialItem = function( item ) {
+              var tutorialElement = document.createElement( "div" );
+              tutorialElement.classList.add( "tutorial-list-item" );
+              tutorialElement.addEventListener( "click", function() {
+                iframe.src = item.url;
```
humphd: One thing I noticed when you demoed this was that we're showing the "Remix This" banner in these tutorials, which I think is wrong.  You should probably file a bug to special case the loading, maybe adding `?showRemix=0` or something to the URL before you load it into the iframe.

public/src/ui/header.js
```diff
@@ -331,6 +313,86 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
...
+            tutorialView.appendChild( iframe );
+            tutorialView.appendChild( iframeCover );
+            tutorialView.appendChild( closeButton );
+            document.body.appendChild( tutorialView );
+
+            closeButton.addEventListener( "click", function() {
+              tutorialView.style.opacity = 0;
+            }, false );
+
+            $( tutorialView ).draggable({
```
humphd: Our mixing of jQuery and non-jQuery is odd.  Why suddenly start using it here after doing so much `document.*` above?

public/templates/basic/index.html
```diff
@@ -16,9 +16,10 @@
   <link rel="stylesheet" href="../assets/plugins/text/popcorn.text.css">
   <link rel="stylesheet" href="../assets/plugins/wikipedia/popcorn.wikipedia.css">
 
-  <link href="https://mozorg.cdn.mozilla.net/media/css/tabzilla-min.css" rel="stylesheet" data-butter-exclude>
+  <link href="http://login.mofostaging.net/css/nav.css" rel="stylesheet" data-butter-exclude>
```
humphd: Assuming `mofostaging` is going to mean other deployments break.

public/templates/basic/index.html
```diff
@@ -16,9 +16,10 @@
...
 
   <script src="../../src/butter.js" data-butter-exclude></script>
+  <script src="http://makeapi.mofostaging.net/js/make-api.js" data-butter-exclude></script>
```
humphd: Same issue here (and elsewhere, for that matter).

routes/index.js
```diff
@@ -107,6 +107,7 @@ module.exports = function routesCtor( app, Project, filter, sanitizer,
       projectJSON.publishUrl = utils.generatePublishUrl( doc.id );
       projectJSON.iframeUrl = utils.generateIframeUrl( doc.id );
       projectJSON.remixedFrom = doc.remixedFrom;
+      projectJSON.remixUrl = utils.generateIframeUrl( doc.remixedFrom );
```
humphd: I'd do your check here to see if  we have `remixedFrom` on `doc` instead of returning null from `generateIframeUrl`

routes/index.js
```diff
@@ -132,6 +133,7 @@ module.exports = function routesCtor( app, Project, filter, sanitizer,
       projectJSON.name = "Remix of " + project.name;
       projectJSON.template = project.template;
       projectJSON.remixedFrom = project.id;
+      projectJSON.remixUrl = utils.generatePublishUrl( project.id );
```
humphd: Same.

public/src/ui/header.js
ScottDowne: I'm only using jQuery here for jQuery UI draggable and resizable.

humphd: You used it to get the `tutorialView` element out of the DOM too.

public/src/ui/header.js
ScottDowne: K, going to file it.

ScottDowne: https://bugzilla.mozilla.org/show_bug.cgi?id=869946

public/src/ui/header.js
ScottDowne, line 23: It is broken anyway. We should fix it when we add it again once the new nav lands, probably.

public/src/ui/header.js
ScottDowne: I'm not sure I understand.

I am using Make to create an instance of the make api, that later we can do a search for tutorials.

public/templates/basic/index.html
ScottDowne: I'm not sure how we're doing this yet on client side.

I can build a env file here, and pass that in with a template of some sort.

public/src/ui/header.js
ScottDowne: No reason.

It was just quicker at the time, but I can go update it now.

public/src/ui/header.js
ScottDowne: Still thinking on the best way to handle this.

If there is an error here, we probably don't want popcorn maker to stop. We can just not display the tutorial button to the user. If so, what is the best way to inform any devs that this has failed? Should we inform the user there was a failure getting the tutorial? A red UI element? Not... sure.

public/src/ui/header.js
ScottDowne: tutorialView is currently being build by createElement, soon to be a dom fragment.

I'm only seeing the two uses of $. I am missing where the other is.

public/src/ui/header.js
ScottDowne: Github seems to of outdated this.

public/src/ui/header.js
```diff
@@ -331,6 +313,74 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
...
+        make.tags( "tutorial:" + tutorialUrl ).then( function( err, results ) {
+          var tutorialView = Lang.domFragment( TUTORIAL_VIEW_TEMPLATE, ".tutorial-view" ),
+              tutorialTemplate = Lang.domFragment( TUTORIAL_LIST_TEMPLATE, ".tutorial-template" ),
+              iframeCover = tutorialView.querySelector( ".tutorial-iframe-cover" ),
+              iframe = tutorialView.querySelector( ".tutorial-iframe" ),
+              closeButton = tutorialView.querySelector( ".tutorial-close-button" ),
+              viewTitle = tutorialView.querySelector( ".tutorial-view-title" ),
+              tutorialList = tutorialTemplate.querySelector( ".tutorial-list" );
+
+          if ( err ) {
```
ScottDowne: Still thinking on the best way to handle this.

If there is an error here, we probably don't want popcorn maker to stop. We can just not display the tutorial button to the user. If so, what is the best way to inform any devs that this has failed? Should we inform the user there was a failure getting the tutorial? A red UI element? Not... sure.

public/src/ui/header.js
```diff
@@ -1,8 +1,12 @@
-define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data", "ui/webmakernav/webmakernav", "ui/widget/textbox", "ui/widget/tooltip" ],
-  function( Dialog, Lang, HEADER_TEMPLATE, UserData, WebmakerBar, TextBoxWrapper, ToolTip ) {
+define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "text!layouts/tutorial-list.html","text!layouts/tutorial-view.html", "ui/user-data", "ui/webmakernav/webmakernav", "ui/widget/textbox", "ui/widget/tooltip" ],
+  function( Dialog, Lang, HEADER_TEMPLATE, TUTORIAL_LIST_TEMPLATE, TUTORIAL_VIEW_TEMPLATE, UserData, WebmakerBar, TextBoxWrapper, ToolTip ) {
 
   return function( butter, options ){
 
+    var make = Make({
```
humphd: Can you get this into requireJS?  You can do something like we did with the UAParser, where it's an external file we include, but it comes in through require vs. being a global.

public/src/ui/header.js
humphd: Yeah, just return here I think.

public/templates/basic/index.html
humphd: Well, we do have a config, why not use it?

public/src/ui/header.js
humphd: You had a comment here that got deleted saying you didn't quite understand.  The issue is that you're assuming a global thing called `Make`, which we don't do in Butter.  Instead, you need to pass it in via the `define` call above.  We have other external JS libs we do this with (UAParser is one).  Follow that same pattern.

ScottDowne: Yeah, I got it now.

public/templates/basic/index.html
ScottDowne: We do, but without nunjucks or something like that, it is hard to get it where I need it.

This is going to be solved when the nav is landed. Can we sort it out there?

I talked to Jon a bit about this, in general, Popcorn Maker needs work here. It is not engineered to handle external dependencies client side.

Also, this is likely to only change once. I don't see a big deal hard coding it. Once it is on deployed, do you still feel the need to have it in a config?

public/templates/basic/config.json
```diff
@@ -3,6 +3,10 @@
   "template": "index.html",
   "baseDir": "../../",
   "savedDataUrl": "saved-data.json",
+  "endpoints": {
+    "make": "http://makeapi.mofostaging.net",
+    "login": "http://login.mofostaging.net"
+  },
```
ScottDowne: Not sure how I feel about this yet, but trying.

public/src/ui/header.js
jbuck, line 23: Yeah, that sounds good

public/src/ui/header.js
```diff
@@ -331,6 +310,81 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
     });
 
     butter.listen( "ready", function() {
+      var tutorialUrl;
+      if ( butter.project.id >= 0 || butter.project.remixedFrom >= 0) {
```
jbuck: Can you change this to an early return?

public/src/ui/header.js
```diff
@@ -331,6 +310,81 @@ define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data
...
+        if ( butter.project.id >= 0 ) {
+          tutorialUrl = butter.project.publishUrl;
+        } else if ( butter.project.remixedFrom >= 0 ) {
+          tutorialUrl = butter.project.remixUrl;
+        }
+
+        require( [ _makeEndpoint + "/js/make-api.js" ], function( Make ) {
```
jbuck: I think this might break with the rjs optimizer

public/templates/basic/index.html
jbuck: I think it'd be best to just include the css from this file inside the local webmakernav.css file

routes/index.js
jbuck: Maybe `remixedFromUrl`?

public/templates/basic/template.js
```diff
@@ -7,11 +7,6 @@
     Butter.init({
       config: "config.json",
       ready: function( butter ) {
-        var script;
-        EditorHelper.init( butter );
```
jbuck, line 11: I think this needs to come back

public/src/ui/header.js
jbuck: Can you move this to the top-level require? Then we can do something like `npm install makeapi` and then add some code to the server like:

``` javascript
app.get( "/external/makeapi.js", function( req, res ) {
  res.sendfile("node_modules/makeapi/public/js/make-api.js");
});
```

And then inside the top-level require:

``` javascript
require("external/makeapi", function(MakeAPI) { ... });
```

routes/index.js
```diff
@@ -106,7 +106,10 @@ module.exports = function routesCtor( app, Project, filter, sanitizer,
       projectJSON.template = doc.template;
       projectJSON.publishUrl = utils.generatePublishUrl( doc.id );
       projectJSON.iframeUrl = utils.generateIframeUrl( doc.id );
-      projectJSON.remixedFrom = doc.remixedFrom;
+      if ( doc.remixedFrom || doc.remixedFrom === 0 ) {
```
jbuck, line 109: !== 0?

package.json
```diff
@@ -96,7 +96,9 @@
     "node-uuid": "1.4.0",
     "requirejs-middleware": "0.0.5",
     "sequelize": "1.5.0",
-    "shelljs": "0.1.2"
+    "shelljs": "0.1.2",
+    "habitat": "0.4.2",
```
jbuck: No need to add habitat, just use the existing nconf

package.json
```diff
@@ -96,7 +96,9 @@
...
+    "makeapi": "0.0.11"
```
jbuck: alphabetical order please

env.sample
```diff
@@ -0,0 +1,2 @@
+# make API url
+export MAKE_ENDPOINT="http://makeapi.mofostaging.net"
```
jbuck: You can move this into lib/default-config

public/src/ui/header.js
```diff
@@ -1,5 +1,5 @@
-define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "ui/user-data", "ui/webmakernav/webmakernav", "ui/widget/textbox", "ui/widget/tooltip" ],
-  function( Dialog, Lang, HEADER_TEMPLATE, UserData, WebmakerBar, TextBoxWrapper, ToolTip ) {
+define([ "dialog/dialog", "util/lang", "text!layouts/header.html", "text!layouts/tutorial-list.html","text!layouts/tutorial-view.html", "ui/user-data", "ui/webmakernav/webmakernav", "ui/widget/textbox", "ui/widget/tooltip", "/external/make-api.js", "json!/api/butterconfig" ],
```
jbuck, line 1: I'm happy this was so easy to implement!

server.js
```diff
@@ -386,6 +390,16 @@ app.get( '/dashboard', filter.isStorageAvailable, function( req, res ) {
 
 app.get( '/healthcheck', routes.api.healthcheck );
 
+app.get( '/external/make-api.js', function( req, res ) {
+  res.sendfile( "node_modules/makeapi/public/js/make-api.js" );
+});
+
+app.get( '/api/butterconfig', function( req, res ) {
+  res.json({
+    "makeEndpoint": env.get('MAKE_ENDPOINT')
```
jbuck: using our current nconf-based config system this would get changed to `config.makeEndpoint` I think? Or maybe `config.MAKE_ENDPOINT`.

routes/index.js
ScottDowne, line 109: === 0 is intended. well, from a number point of view.

If we do not allow projects as /0, which, we probably don't we can probably assume we don't need to even check if 0 is true.

Anyway, def not !== 0. In this case I am ensuring it is a number.

server.js
```diff
@@ -22,7 +23,10 @@ var express = require('express'),
     WWW_ROOT = path.resolve( __dirname, config.dirs.wwwRoot ),
     VALID_TEMPLATES = config.templates;
 
-var templateConfigs = {};
+habitat.load();
```
jbuck, line 26: Remove these bits and you're golden

PR #1677, opened 2013-05-08
T858739b

mjschranz:
https://bugzilla.mozilla.org/show_bug.cgi?id=858739


public/src/core/project.js
```diff
@@ -32,7 +32,7 @@ define( [ "core/eventmanager", "core/media", "util/sanitizer" ],
         // Interval for backups, starts first time user clicks Save.
         _backupInterval = -1,
 
-        _thumbnail = "/resources/icons/fb-logo.png";
+        _thumbnail = window.location.origin + "/resources/icons/fb-logo.png";
```
humphd: Why this change?

public/templates/assets/editors/editorhelper.js
```diff
@@ -268,7 +260,8 @@
       }, false );
 
       dropContainer.addEventListener( "drop", function( e ) {
-        var file, imgSrc, imgURI, image, div;
+        var file,
+            fd = new FormData();
```
humphd: Are you going to shim this? http://caniuse.com/#search=formdata says IE9 won't have it.

public/templates/assets/editors/editorhelper.js
```diff
@@ -281,107 +274,24 @@
...
-                resizeRatio = Math.min( wRatio, hRatio );
-                scaledHeight = height * resizeRatio;
-                scaledWidth = width * resizeRatio;
-              }
+              butter.dispatch( "droppable-upload-failed" );
             }
+          });
+        });
 
-            canvas.width = scaledWidth;
```
humphd, line 345: Ripping out all this scaling code is something we should ponder.  We did it originally so that we wouldn't bother with image sizes that were way bigger than the frame.  However, that was mostly something we were doing to try and limit datauri sizes exploding in projects.  Since we're now going to just go straight to uploading, and ditch datauris, maybe this isn't necessary anymore.  On the other hand, it means that if you dnd a bunch of photos from your camera that are 5M each, we'll store that, and then loading will be slow.  We could post process these on the server before we push them up to S3, and shrink them down a bit if they exceed a max size.  Might be worth dumping it here, and filing a bug to figure it out later.

server.js
```diff
@@ -155,6 +155,32 @@ function writeEmbed( embedPath, url, data, callback ) {
   stores.publish.write( embedPath, sanitized, callback );
 }
 
+app.put( "/api/upload", function( req, res ) {
```
humphd: This name should more accurately reflect what we're doing here.  I'd suggest `/api/image` since you're PUTing an image.

public/templates/assets/editors/editorhelper.js
mjschranz, line 345: This is basically how I felt when making the changes. We don't want to really loose the functionality that we had here but for the sake of getting this moving along I think it's good to get file this for next week.

server.js
```diff
@@ -155,6 +155,32 @@ function writeEmbed( embedPath, url, data, callback ) {
...
+  var image = req.files.image,
+      urlPair = utils.generateDataURIPair();
```
humphd: Probably we should update the naming here, since this is just giving back a unique filename and URL, nothing to do with data-uri.

public/templates/assets/editors/editorhelper.js
mjschranz: IE9 doesn't actually even support the File API enough to use the drop/uploading for thumbnails. We bailed in the past because of this and that same code still exists to prevent that from happening.

However, I am initializing my FormData object in the wrong place.

public/src/core/project.js
mjschranz: Your right, this might be potentially the wrong results. Right now we populate the OG meta data link with a baseHref determined https://github.com/mozilla/butter/blob/master/server.js#L307 and setup https://github.com/mozilla/butter/blob/master/views/embed-shell.jade#L14

To me this seemed like the safest way to handle our default thumbnail. Is there a better alternative I'm missing?

lib/file-store.js
```diff
@@ -70,11 +70,18 @@ function LocalFileStore( options ) {
...
 
-  fs.writeFile( path, data, function( err ) {
+  var type = contentType;
+
+  if ( typeof contentType === "function" ) {
+    type = this.contentType;
+    callback = contentType;
+  }
+
+  fs.writeFile( path, data, { encoding: type }, function( err ) {
```
jbuck: I don't think you use `encoding` like this; it's meant to be like `utf8`, `ascii`, `binary`, etc

server.js
```diff
@@ -155,6 +155,32 @@ function writeEmbed( embedPath, url, data, callback ) {
...
+      res.json( { error: "Failed to upload image. Reading file failed." }, 500 );
+      metrics.incremenet( 'error.save.store-image' );
+      return;
+    }
+
+    stores.images.write( urlPair.filename, data, image.type, function( error ) {
+
+      if ( error ) {
+        res.json( { error: "Failed to upload image. Uploading file failed." }, 500 );
+        metrics.incremenet( 'error.save.store-image' );
```
jbuck: `increment`

lib/file-store.js
mjschranz: The NodeJS docs seems to say it takes an options object http://nodejs.org/api/fs.html#fs_fs_readfile_filename_options_callback

server.js
```diff
@@ -155,6 +155,32 @@ function writeEmbed( embedPath, url, data, callback ) {
   stores.publish.write( embedPath, sanitized, callback );
 }
 
+app.put( "/api/image", function( req, res ) {
```
jbuck: Can you move this code to `routes/api/image.js` or something along those lines? It shouldn't be in server.js, if it's possible to move it out there.

public/src/core/project.js
jbuck: `location.origin` is undefined in FF20. Maybe `location.protocol + "//" + location.host + "/resources/icons/fb-logo.png";` ?

lib/file-store.js
jbuck: Right, but it doesn't accept a mime type like "image/png", it accepts a file encoding like "utf8" or "binary". I suspect you want "binary" here, but testing is needed.

jbuck: Yep!

```
fs.js:113
    throw new Error('Unknown encoding: ' + encoding);
          ^
Error: Unknown encoding: image/jpeg
    at assertEncoding (fs.js:113:11)
    at Object.fs.writeFile (fs.js:939:3)
    at Object.LocalFileStore.write (/Users/jon/Dropbox/Sites/butter/lib/file-store.js:84:6)
    at /Users/jon/Dropbox/Sites/butter/server.js:170:19
    at fs.js:266:14
    at Object.oncomplete (fs.js:107:15)
```

routes/api/image.js
```diff
@@ -0,0 +1,29 @@
...
+
+    fs.readFile( image.path, { encoding: "binary" }, function( err, data ) {
+
+      if ( err ) {
+        res.json( { error: "Failed to upload image. Reading file failed." }, 500 );
+        metrics.increment( 'error.save.store-image' );
+        return;
+      }
+
+      utils.stores().images.write( urlPair.filename, data, function( error ) {
```
humphd: I think you should pass `stores` in like `metrics` above, I don't think it belongs on utils.

routes/api/image.js
```diff
@@ -0,0 +1,29 @@
...
+  var fs = require( "fs" );
+
+  return function( req, res ) {
+    var image = req.files.image,
+        urlPair = utils.generateFileName();
+
+    fs.readFile( image.path, { encoding: "binary" }, function( err, data ) {
+
+      if ( err ) {
+        res.json( { error: "Failed to upload image. Reading file failed." }, 500 );
```
humphd: This is backward: http://expressjs.com/api.html#res.json

routes/api/image.js
```diff
@@ -0,0 +1,29 @@
...
+        metrics.increment( 'error.save.store-image' );
+        return;
+      }
+
+      utils.stores().images.write( urlPair.filename, data, function( error ) {
+
+        if ( error ) {
+          res.json( { error: "Failed to upload image. Uploading file failed." }, 500 );
```
humphd: Backwards as well.  Can you file a bug to fix this to fix the rest of cornfield too?

routes/api/image.js
```diff
@@ -0,0 +1,29 @@
...
+          metrics.increment( 'error.save.store-image' );
+          return;
+        }
+
+        res.json( { url: urlPair.url }, 200 );
```
humphd: Backwards, but also note, you don't need the `200` here--only when it's an error case.  Just include the object arg.

routes/index.js
```diff
@@ -164,68 +166,30 @@ module.exports = function routesCtor( app, Project, filter, sanitizer,
...
     var projectData = req.body;
 
     if ( req.body.id ) {
-      files = datauri.filterProjectDataURIs( projectData.data, utils.generateDataURIPair );
 
       Project.update( { email: req.session.email, id: req.body.id, data: projectData },
-                      function( err, doc, imagesToDestroy ) {
+                      function( err, doc ) {
         if ( err ) {
           res.json( { error: err }, 500 );
```
humphd, line 185: /me cries

routes/index.js
```diff
@@ -234,22 +198,8 @@ module.exports = function routesCtor( app, Project, filter, sanitizer,
...
-            metrics.increment( 'project.images-upload', files.length );
-          });
-        }
-        else {
-          // Send back the newly added row's ID
-          res.json( { error: 'okay', projectId: doc.id }, 200 );
-        }
-
+        // Send back the newly added row's ID
+        res.json( { error: 'okay', projectId: doc.id }, 200 );
```
humphd: Drop the 200 arg.

routes/api/image.js
mjschranz: Will do.

public/templates/assets/editors/editorhelper.js
```diff
@@ -281,107 +273,25 @@
...
-                scaledWidth = width * resizeRatio;
-              }
-            } else {
-              if ( height <= MAX_IMAGE_WIDTH && width <= MAX_IMAGE_HEIGHT ) {
-                scaledWidth = width;
-                scaledHeight = height;
+        fd = new FormData();
+        fd.append( "image", file );
+
+        require( [ "util/xhr" ], function( XHR ) {
```
humphd: Why require this deep in here, and multiple times (i.e., per call), vs once at the top?

public/templates/assets/editors/editorhelper.js
```diff
@@ -281,107 +273,25 @@
...
-                resizeRatio = Math.min( wRatio, hRatio );
-                scaledHeight = height * resizeRatio;
-                scaledWidth = width * resizeRatio;
-              }
-            } else {
-              if ( height <= MAX_IMAGE_WIDTH && width <= MAX_IMAGE_HEIGHT ) {
-                scaledWidth = width;
-                scaledHeight = height;
+        fd = new FormData();
+        fd.append( "image", file );
```
humphd: If you know the `contentType` throw that on `fd` as well, and send it to to the server with the file.

public/templates/assets/editors/editorhelper.js
```diff
@@ -281,107 +273,25 @@
...
+            if ( !data.error ) {
+              if ( trackEvent ) {
+                trackEvent.update( { src: data.url } );
               } else {
-                wRatio = MAX_IMAGE_HEIGHT / width;
-                hRatio = MAX_IMAGE_WIDTH / height;
-                resizeRatio = Math.min( wRatio, hRatio );
-                scaledHeight = height * resizeRatio;
-                scaledWidth = width * resizeRatio;
+                callback( data.url );
```
humphd, line 285: Question: we really only ever call the callback if this works?  Wouldn't you typically also trigger it for error cases, and flow the error through?  Maybe this is OK, just seems unexpected to me.

mjschranz, line 285: So, I was thinking of modifying how this all worked so that both the image editor and project editor actually expected the callback only and handled themselves how they wanted to use the information. This is probably a better, or at least less confusing way to read the code and always have a callback that is used.

lib/file-store.js
```diff
@@ -70,11 +70,18 @@ function LocalFileStore( options ) {
...
+LocalFileStore.prototype.write = function( path, data, contentType, callback ) {
   path = Path.join( this.root, this.expand( path ) );
   ensurePathExistsSync( Path.dirname( path ) );
 
-  fs.writeFile( path, data, function( err ) {
+  if ( typeof contentType === "function" ) {
+    callback = contentType;
+  }
+
+  // Hard code content type to binary for local.
```
humphd: Don't even bother with this.  Just rip the callback off contentType like you do above, then hard code `{ encoding: "binary" }` in line 84 below.  It's confusing to refer to that as a contentType, since it isn't.

routes/api/image.js
```diff
@@ -0,0 +1,29 @@
...
+        urlPair = utils.generateFileName();
+
+    fs.readFile( image.path, { encoding: "binary" }, function( err, data ) {
+
+      if ( err ) {
+        res.json( 500, { error: "Failed to upload image. Reading file failed." } );
+        metrics.increment( 'error.save.store-image' );
+        return;
+      }
+
```
humphd, line 15: Let's add another check (maybe middleware exists for this already) to filter out contentTypes that aren't image/\* (image/png, image/gif, ...).  We don't want to store arbitrary files for people.

routes/api/image.js
```diff
@@ -0,0 +1,29 @@
...
+      stores.images.write( urlPair.filename, data, image.type, function( error ) {
+
+        if ( error ) {
+          res.json( 500, { error: "Failed to upload image. Uploading file failed." } );
+          metrics.increment( 'error.save.store-image' );
+          return;
+        }
+
+        res.json( { url: urlPair.url } );
+        metrics.increment( 'project.images-upload', 1 );
```
humphd: You don't need the 1 here, it defaults to that.

lib/file-store.js
```diff
@@ -70,11 +70,15 @@ function LocalFileStore( options ) {
...
+LocalFileStore.prototype.write = function( path, data, contentType, callback ) {
   path = Path.join( this.root, this.expand( path ) );
   ensurePathExistsSync( Path.dirname( path ) );
 
-  fs.writeFile( path, data, function( err ) {
+  if ( typeof contentType === "function" ) {
+    callback = contentType;
+  }
+
+  fs.writeFile( path, data, { encoding: "binary" }, function( err ) {
```
jbuck: Remove `{ encoding: "binary" }`, upload to S3 doesn't work if that's set

routes/api/image.js
```diff
@@ -0,0 +1,29 @@
+module.exports = function( metrics, utils, stores ) {
+  var fs = require( "fs" );
+
+  return function( req, res ) {
+    var image = req.files.image,
+        urlPair = utils.generateFileName();
+
+    fs.readFile( image.path, { encoding: "binary" }, function( err, data ) {
```
jbuck: Remove `{ encoding: "binary" }`, upload to S3 doesn't work if that's set

lib/file-store.js
mjschranz: This is only the local store anyway. Not sure this is what you intended.

PR #1678, opened 2013-05-09
870403 - Fix all server responses to pass args in correct order

alicoding:
https://bugzilla.mozilla.org/show_bug.cgi?id=870403


alicoding:
Closed already fixed in this bug https://github.com/mozilla/popcorn.webmaker.org/commit/dd1bc234f21d34c17ec9e473f822928ab6b5b7d0

PR #1681, opened 2013-05-10
T863873

mjschranz:
https://bugzilla.mozilla.org/show_bug.cgi?id=863873

### DIFF
--- a/lib/default-config.js
+++ b/lib/default-config.js
@@ -36,6 +36,10 @@ module.exports = {
     "wwwRoot": "public",
     "templates": "public/templates"
   },
+  "makeapi": {
+    "apiURL": "http://makeapi.mofostaging.net",
+    "auth": "testuser:password"
+  },
   "publishStore": {
     "type": "local",
     "options": {
--- a/lib/project.js
+++ b/lib/project.js
@@ -98,7 +98,7 @@ module.exports = function( config, makeapi, utils, dbReadyFn ) {
           title: doc.name,
           author: doc.author,
           email: email,
-          contentType: "application/x-popcorn-maker",
+          contentType: "application/x-popcorn",
           url: utils.generatePublishUrl( doc.id ),
           thumbnail: doc.thumbnail,
           description: doc.description,
--- a/lib/project.js
+++ b/lib/project.js
@@ -7,7 +7,7 @@ function defaultDBReadyFunction( err ) {
   }
 }
 
-module.exports = function( config, dbReadyFn ) {
+module.exports = function( config, makeapi, utils, dbReadyFn ) {
   config = config || {};
 
   dbReadyFn = dbReadyFn || defaultDBReadyFunction;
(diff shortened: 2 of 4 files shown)


lib/default-config.js
```diff
@@ -36,6 +36,10 @@ module.exports = {
     "wwwRoot": "public",
     "templates": "public/templates"
   },
+  "makeapi": {
+    "apiURL": "http://makeapi.mofostaging.net",
+    "auth": "testuser:password"
```
humphd, line 41: We won't be able to put this in the file like this, so we'll need another solution that has it in `env` somehow.

lib/project.js
```diff
@@ -88,7 +89,33 @@ module.exports = function( config, dbReadyFn ) {
...
+          title: doc.name,
+          author: doc.author,
+          email: email,
+          contentType: "application/x-popcorn",
+          url: utils.generatePublishUrl( doc.id ),
+          thumbnail: doc.thumbnail,
+          description: doc.description,
+          remixedFrom: data.remixedFrom,
+          tags: [
+            "makeType:popcorn",
```
humphd: This is wrong.  I know they want it, but we aren't going to do it this way.  If you just want it for a working demo, cool, but it needs to get a bug filed to fix it later.

lib/project.js
```diff
@@ -88,7 +89,33 @@ module.exports = function( config, dbReadyFn ) {
...
+            "demo:popcorn"
```
humphd: Same here.  We're going to do app tags like `popcorn.webmaker.org:foo`.

mjschranz: This was mostly for the demo purposes. I actually don't have any intent of landing this right now.

lib/project.js
```diff
@@ -214,7 +253,24 @@ module.exports = function( config, dbReadyFn ) {
           callback( err );
         })
         .success( function( projectUpdateResult ) {
-          callback( null, projectUpdateResult );
+          makeClient.url( utils.generatePublishUrl( projectUpdateResult.id ) ).then(function( err, result ) {
+            if ( err ) {
+              return callback( err, projectUpdateResult );
+            }
+
+            makeClient.update( result.hits[ 0 ]._id, {
```
humphd, line 260: One thing to note for later, I'm changing this so that you use `results` directly vs. `results.hits`.

lib/project.js
```diff
@@ -108,7 +135,19 @@ module.exports = function( config, dbReadyFn ) {
 
           project.destroy()
           .success(function() {
-            callback( null, project );
+            makeClient.url( utils.generatePublishUrl( project.id ) ).then(function( err, result ) {
+              if ( err ) {
+                return callback( err );
+              }
+
+              makeClient.remove( result.hits[ 0 ]._id, function( error ) {
```
humphd, line 142: Same thing here, this will be changing to drop `hits`.

PR #1685, opened 2013-11-25
Bug904139 - Add grab and grabbing icons for the timeline layers

maronin:
Tested in Chrome, Firefox, and IE10. Works in all three. 

Note: IE does not support the grab cursor like Chrome and Firefox, so I used custom icons I made myself when the user is on IE.


mjschranz:
Hey Mark! Thanks for the pull request but the repository we use for development with popcorn.webmaker.org is at mozilla/popcorn.webmaker.org. This repository is closed/deprecated by and large.

Could you please reopen this there? Thanks!

maronin:
Alright will do! :)

public/css/tray-timeline.less
```diff
@@ -281,7 +281,9 @@
 .track-handle {
   border: none;
   border-bottom: 1px solid darken( @baseLight, 25% );
-  cursor: pointer;
+  cursor: url(/media/img/hand.cur), default; /*EI Support*/
```
ScottDowne, line 284: I don't think I care enough about cursors in IE to add a custom cursor file.

maronin, line 284: Ok, how about using google's grab cursor? using something like:

`cursor: url(https://mail.google.com/mail/images/2/openhand.cur)`
and
`cursor: url(https://mail.google.com/mail/images/2/closedhand.cur)`

I don't know about the licensing and if you can use those icons, so I'm not sure. 

Maybe IE can just have a `cursor: pointer` and a `cursor: move` instead?

public/css/tray-timeline.less
```diff
@@ -281,7 +281,9 @@
...
+  cursor: -webkit-grab;                      /*Chrome Support*/
+  cursor: -moz-grab;                         /*Mozzila Support*/
```
ScottDowne, line 286: Come FF 27, grab and grabbing are no longer going to need -moz-, so should list that one too.

So like this:

cursor: -moz-grab;
cursor: -webkit-grab;
cursor: grab;

Same for grabbing.

For grabbing, you can make a class for grabbing, then in the js you can just add and remove the grabbing class.

